PR #11, opened 2018-02-27
[WIP] feat(manifest): author data from npmrc

### LINKED ISSUE #2: author data?
we can grab author data from Cargo as well, but we MIGHT consider checking if there is an `.npmrc` and preferring that if it exists

ashleygwilliams:
✨👾 WIP; do no merge 👾 ✨ 

fixes #2

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,4 @@
-
+.npmrc
 /target
 **/*.rs.bk
 /pkg
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -186,6 +186,17 @@ dependencies = [
 ]
 
 [[package]]
+name = "npmrc"
+version = "0.1.2"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+dependencies = [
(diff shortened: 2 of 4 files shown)


sendilkumarn:
Does it includes fallback to cargo author when there is no .npmrc file?

ashleygwilliams:
not yet! it falls back to `""` but i will add that now :D. 

also: just filed https://github.com/yoshuawuyts/npmrc/issues/2

ashleygwilliams:
hrm @sendilkumarn - this is actually a *lil* tricky, as Crates support an authors vector... should i just pick the first one? or do you think it's better to leave it as an empty string? i'm torn.

sendilkumarn:
the first one makes sense. I was stuck at the same point. or intelligent identification based on the computer name. well let it be empty whn the length is more than one.

ashleygwilliams:
yeah ok cool. thanks for workshopping that with me ☺️

ashleygwilliams:
closing in favor of a new PR as this is from a very old version of the project and not worth the merge

Issue #16, opened 2018-03-02
reorg to use example project


ashleygwilliams:
currently i'm "testing" this tool by running it recrusively on itself (i.e. using the wasm-pack cargo.toml and wasm-pack rust files to compile) 

this stops working now that i'm trying to build to wasm, as much of the code in the rust files for this project won't compile to wasm for several reasons.

rewrite this to be targeting an example so it actually compiles

ashleygwilliams:
fixed by #17 blocked by #20

PR #17, opened 2018-03-02
feature: cargo/wasm-bindgen build step

### LINKED ISSUE #10: shell out to cargo build and wasm-bindgen
after talking with @alexcrichton about this, it seems like shelling out vs using cargo/wasm-bindgen as libraries is not a clear call. shelling out is easier so i'mma do that for now! we may move to the library usage later tho.

ashleygwilliams:
fixes #10 
~blocked by https://github.com/ashleygwilliams/wasm-pack/issues/16 at the moment~

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,4 @@
 
 **/target
 **/*.rs.bk
-/pkg
+**/pkg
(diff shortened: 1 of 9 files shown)


ashleygwilliams:
blocked by #16

ashleygwilliams:
i'm gonna test this in a separate PR, i'd really like to get this out and published since it works, but just very very barely lol

PR #31, opened 2018-03-05
fix : file extension to have bg

sendilkumarn:


ashleygwilliams:
thanks so much, @sendilkumarn ! i'll merge once we go green :D

Issue #24, opened 2018-03-04
move stuff outta lib.rs into manifest.rs


ashleygwilliams:

ashleygwilliams:
closed by #32

PR #32, opened 2018-03-05
feat(readme): copy from crate

### LINKED ISSUE #30: copy over readme from crate (~for now~)
for 0.1.0 let's just copy over the README.md from the crate

there's great convo in #7 and i'd like to move to something more like that for 0.2.0 or greater.

ashleygwilliams:
fixes #30 and #24

### DIFF
--- a/src/readme.rs
+++ b/src/readme.rs
@@ -1,0 +1,10 @@
+use std::fs;
+use std::fs::File;
+use failure::Error;
+
+pub fn copy_from_crate(path: &str) -> Result<(), Error> {
+    let crate_readme_path = format!("{}/README.md", path);
+    let new_readme_path = format!("{}/pkg/README.md", path);
+    fs::copy(&crate_readme_path, &new_readme_path)?;
+    Ok(())
+}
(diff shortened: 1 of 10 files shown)


src/readme.rs
```diff
@@ -0,0 +1,11 @@
+use std::fs;
+use std::fs::File;
+use failure::Error;
+
+pub fn copy_from_crate(path: &str) -> Result<(), Error> {
+    let path_to_crate_readme = format!("{}/README.md", path);
```
sendilkumarn: nit: can rename to `crate_readme_path`

ashleygwilliams: lol yes yes this is a good idea

ashleygwilliams: i'm a long variable names monster 🐉

ashleygwilliams:
whew this is a big one but also it makes things a LOT nicer.

@yoshuawuyts this backs out the dir_diff stuff- i think i'd like to have a larger discussion on your vision for how that works? i'm not entirely sure it was testing what we wanted, so for now i'm doing this really naive `fs::metadata` testing instead. wanna start an issue and maybe spell it out a bit more? (i'm very open to whatever strategy but it wasn't working for my brain at the moment and i really want to get this merged)

Issue #46, opened 2018-03-09
Installing wasm-bindgen tool is non obvious


mgattozzi:
When first running wasm-pack there's no indication that wasm-bindgen is being installed at first. I had a huge spike in CPU usage and was confused why it happened the first time I ran the command. It wasn't until looking in the code for something that I saw why.

If this is the first time then we should print out that it will be compiled and installed. If it's not and it exists and the version is not outdated we should print out that it's already installed. If it's installed and outdated we should print out that we're updating it.

ashleygwilliams:
now that i've merged #49 i'm curious if you think this is ok now? i've made it clear that one step is installing wasm-bindgen and i've filed #51 to eventually account for not taking that step if we don't need to

mgattozzi:
@ashleygwilliams I'll need to run this tomorrow at some point. Will let you know!

mgattozzi:
This looks good to me! I think we can definitely follow up with #51 now.

Issue #44, opened 2018-03-09
No stdout when compiling


mgattozzi:
When using `wasm-pack init` the following output is produced:

```
✅ rustup_add_wasm_target succeeded and stdout was:
✅  cargo_build_wasm succeeded and stdout was:
🏎️ 💨  compiled to wasm!
```

This makes it seem like there was supposed to be output. If there's not it should just cut off the rest of the sentance after and.

ashleygwilliams:
yup, working on this now in combination with #39

ashleygwilliams:
now that i've merged #49 i'm curious if you think this is ok now? i've turned off all the std out shenanigans

mgattozzi:
This looks good to me and is a bit more accurate. I think we can close this in favor of #50

PR #49, opened 2018-03-12
fix(style): improve output messaging

### LINKED ISSUE #39: cleanup tool messaging output
it looks gross right now

ashleygwilliams:
fixes #39, #44, #46

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -1,11 +1,28 @@
+extern crate console;
 extern crate failure;
+extern crate indicatif;
 #[macro_use]
 extern crate quicli;
 extern crate wasm_pack;
 
+mod progressbar;
+
+use std::time::Instant;
+
+use console::{style, Emoji};
+use indicatif::HumanDuration;
+use quicli::prelude::*;
 use wasm_pack::{bindgen, build, manifest, readme};
 
-use quicli::prelude::*;
+static TARGET: Emoji = Emoji("🎯  ", "");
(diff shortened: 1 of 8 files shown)


src/main.rs
```diff
@@ -32,18 +46,57 @@ enum Command {
...
+        );
         manifest::write_package_json(&crate_path)?;
+        println!(
+            "{} {}Copying over your README...",
+            style("[5/7]").bold().dim(),
+            DANCERS
+        );
         readme::copy_from_crate(&crate_path)?;
+        println!(
+            "{} {}Installing WASM-bindgen...",
```
mgattozzi: Maybe we can make this an optional thing at some point? like if it's installed we just do 6 things rather than 7. But that's like a future improvement and not something that needs to be done now and block 0.1

mgattozzi:
This is all very nice and much easier to see where things are happening :D

src/main.rs
ashleygwilliams: so currently, if wasm-bindgen is already installed this step doesn't actually do anything (it actually fails and yells in stderr). we should def check if it's installed and just skip if it is... both this and adding the target should just not happen if they are already present.

mgattozzi: In which case the numbering will also need to be dynamically done rather than it being hardcoded. Maybe a step at the beginning to calculate what can be done without then run the rest.

PR #58, opened 2018-03-14
clean up imports

steveklabnik:


ashleygwilliams:
@steveklabnik looks like this needs a cargo fmt run

steveklabnik:
I was using stable, not nightly, and there's discrepancies. Fixed!

PR #59, opened 2018-03-14
Improve test layout

steveklabnik:
I suggested this layout, but since Rust 1.22, this layout now works!

https://github.com/ashleygwilliams/wasm-pack/pull/57 should be merged first, and then this will need to be rebased off of it.


ashleygwilliams:
yes this is what i wanted and did not know it was possible. 🎉

steveklabnik:
This has been rebased and is g2g

sendilkumarn:
@steveklabnik good one 👍

PR #60, opened 2018-03-14
Add scope flag to init subcommand

### LINKED ISSUE #45: --scope flag on init command
@mgattozzi brought this up while working on the tutorial. we should support namespaces (which are called scopes in npm land)! this way people taking the tutorial won't all attempt to publish a pkg with the same name.

`wasm-pack init --scope MYSCOPE` should write to the package.json:

```json
{
   "name": "@MYSCOPE/cratename"
}
```

mgattozzi:
npm allows scopes to avoid name collisions. In order to support this
optional feature a flag has been added so that the name in the generated
package.json is correct.

Example for a package named wasm-add:

```bash
wasm-pack init
```

package.json

```json
{
  "name": "wasm-add"
}
```

```bash
wasm-pack init --scope mgattozzi
```

package.json

```json
{
  "name": "@mgattozzi/wasm-add"
}
```

I tested this locally and it worked fine but I'm more than happy to add test cases for it somehow if you want them!

Closes #45

### DIFF
--- a/src/manifest.rs
+++ b/src/manifest.rs
@@ -49,12 +49,21 @@ fn read_cargo_toml(path: &str) -> Result<CargoManifest, Error> {
 }
 
 impl CargoManifest {
-    fn into_npm(self) -> NpmPackage {
+    fn into_npm(self, scope: Option<String>) -> NpmPackage {
         let filename = self.package.name.replace("-", "_");
         let js_file = format!("{}.js", filename);
         let wasm_file = format!("{}_bg.wasm", filename);
         NpmPackage {
-            name: self.package.name,
+            name: match scope {
+                Some(s) => {
+                    let mut name = String::from("@");
+                    name.push_str(&s);
+                    name.push('/');
+                    name.push_str(&self.package.name);
+                    name
+                },
+                None => self.package.name,
+            },
             description: self.package.description,
             version: self.package.version,
             license: self.package.license,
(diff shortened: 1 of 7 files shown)


sendilkumarn:
A test will always be nice. 😁

mgattozzi:
Yeah format is fine! I just did the `push_str` out of habit. A separate method might be fine. We're already passing the value of scope down pretty deeply anyways. What's another function call?

mgattozzi:
Looks like I'll need to fix the tests anyways so I'll add a new one!

mgattozzi:
@sendilkumarn since the function consumed the data needed to make a `package.json` I made it mutable and had it overwrite the name if there was scope wanted otherwise keep the name as is. This ended up not needing a new function but kept the code neater and logically the same!

I also fixed up the tests and added a new one!

mgattozzi:
Ah I knew I was missing something. Gonna run rustfmt real quick.

mgattozzi:
Should be good now!

src/manifest.rs
```diff
@@ -49,21 +49,15 @@ fn read_cargo_toml(path: &str) -> Result<CargoManifest, Error> {
 }
 
 impl CargoManifest {
-    fn into_npm(self, scope: Option<String>) -> NpmPackage {
+    fn into_npm(mut self, scope: Option<String>) -> NpmPackage {
```
sendilkumarn: can we do it without this mutation? 
replacing the `if` with a `match | if.else` will help right?

ashleygwilliams:
yayayayay THANK U. i think this looks good, it's late here tho, so gonna reread tomorrow morning and likely merge then. we're so close to a 0.1.0 and i'm very glad we got this in. thanks so much!!

mgattozzi:
Sounds good :)

ashleygwilliams:
closing via rollup of commits + a doc commit here: https://github.com/ashleygwilliams/wasm-pack/pull/62 thanks so much for this!

Issue #51, opened 2018-03-13
if target/wasm-bindgen already exists don't attempt to re-add/install it


ashleygwilliams:

data-pup:
I would love to help out with this one if it wouldn't be too difficult for a first contribution! Would you have any advice for identifying whether or not target/wasm-bindgen exists already?

I'm guessing that this would mean adding a check to the `cargo_install_wasm_bindgen` function, but I'm not quite sure what the correct method is to identify whether wasm-bindgen has already been added.

ashleygwilliams:
hey @data-pup ! so, i wasn't actually entirely sure- this stack overflow answer seems mostly reasonable: https://stackoverflow.com/questions/35045996/check-if-a-command-is-in-path-executable-as-process. lemme know if you need help digesting it!

data-pup:
Hi! Thank you for the link, that definitely helped. I started working on this today, here's the branch with some first steps. I think I will need to add some extra logic for this to work on Windows, which uses `;` to separate items in $PATH iirc.

Should the progress bar still show up if wasm-bindgen is already installed? Right now it just exits the `cargo_install_wasm_bindgen` function if wasm-bindgen is found in the path.

https://github.com/data-pup/wasm-pack/commit/7b4552081f495e1cb94c3a2ae4febf07fc9a1eb3

ashleygwilliams:
hey @data-pup ! sorry for the delay, i was out sick earlier this week- do you think you could open up a PR? it's much easier to give feedback :D looking forward to reviewing it and thanks so much!

data-pup:
Hi @ashleygwilliams! Sure thing, I can open a PR today. Thanks for helping me out with contributing, I really appreciate it and am looking forward to helping out more!

ashleygwilliams:
@data-pup can you run `rustc --version` and `cargo fmt --version` and let me know what the output is?

data-pup:
@ashleygwilliams Here are the versions. These might be slightly out of date, I can update these and see if that fixes the issue.

```
✨  rustc --version
rustc 1.26.0-nightly (a04b88d19 2018-03-19)

✨  cargo fmt --version
0.3.8-nightly (346238f 2018-02-04)
```

fitzgen:
Fixed in #192

PR #61, opened 2018-03-14
Add assert-cli tests for wasm-pack

### LINKED ISSUE #18: test using assert_cli
https://github.com/killercup/assert_cli

mgattozzi:
This works but for some reason I can't test the `[1/7]` `[2/7]` etc. because it's not showing up in the testing framework as output that happened but it does show up in the console. Not sure if this is an identicaf bug or assert-cli

@killercup thoughts?

Closes #18

### DIFF
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -4,6 +4,7 @@ install:
   - set PATH=%PATH%;C:\Users\appveyor\.cargo\bin
   - rustc -V
   - cargo -V
+  - cargo install wasm-bindgen-cli
 
 build: false
 
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -12,6 +12,19 @@ version = "0.10.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 
 [[package]]
+name = "assert_cli"
+version = "0.5.4"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+dependencies = [
+ "colored 1.6.0 (registry+https://github.com/rust-lang/crates.io-index)",
+ "difference 1.0.0 (registry+https://github.com/rust-lang/crates.io-index)",
+ "environment 0.1.1 (registry+https://github.com/rust-lang/crates.io-index)",
+ "error-chain 0.11.0 (registry+https://github.com/rust-lang/crates.io-index)",
+ "serde_json 1.0.9 (registry+https://github.com/rust-lang/crates.io-index)",
+ "skeptic 0.13.2 (registry+https://github.com/rust-lang/crates.io-index)",
+]
+
+[[package]]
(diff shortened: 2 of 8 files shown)


ashleygwilliams:
fyi this looks like it needs at least a rustfmt (nightly) run, also fails on windows? (also yay thank uu)

killercup:
The progress bar might only show if the ouput is a tty, or to write to stderr.

mgattozzi:
Got it. It didn't show up in either so that is unfortunate. Yeah I'll definitely run a rustfmt

mgattozzi:
I ran rustfmt and then changed the commits to fit your prefered format like you had with my other PR @ashleygwilliams. I think this should be good to go.

ashleygwilliams:
fwiw i'm not super strict about the commit message stuff but figured i'd edit cuz i was adding a commit and doing a fixup anyways <3 (i do appreciate you following it tho 😄)

ashleygwilliams:
hrm, this is still failing on appveyor and i'm not familiar enough with assert-cli to know what's causing the issue. @killercup or @mgattozzi any thoughts?

mgattozzi:
Oh I think I know why. It timed out but appveyor probably didn't have wasm-bindgen installed and cached so it took too long.

ashleygwilliams:
@mgattozzi yeah that was my initial thought too, don't know if we can pass a higher timeout to appveyor?

mgattozzi:
I think the simpler solution might be to have appveyor build wasm-bindgen as part of the build script. Part of the timeout seemed to be cargo and not necessarily AppVeyor looking at the logs again.

ashleygwilliams:
true. happy to accept an update to the appveyor config that preloads it. can probably just add it as a step here https://github.com/ashleygwilliams/wasm-pack/blob/master/.appveyor.yml#L1

ashleygwilliams:
alternatively should it be a dep of this crate? i'm.. not sure. open to thoughts

mgattozzi:
I just pushed the CI change for now. I think maybe we should consider adding it as a dep as part of #51 and update accordingly.

mgattozzi:
ah woops it's wasm-bindgen-cli not wasm-bindgen

ashleygwilliams:
lol i ran into that same issue a few days ago lol. when CI passes i'll merge this

mgattozzi:
Yaaaay! :D

mgattozzi:
Okay so I think there's some kind of bug with assert-cli and how it tests windows.

mgattozzi:
@ashleygwilliams I'm gonna close this PR for now. We can figure out this issue a bit later. I'll have to test this on my windows computer at some point and come up with a better solution.

Issue #12, opened 2018-02-27
handle the specific errors specifically


ashleygwilliams:
right now i'm lazily(read: barely) using Failure and haven't defined any error shenanigans. i should fix that!

PR #65, opened 2018-03-16
feat(warn): warn on all missing fields for a pkg

### LINKED ISSUE #43: Warn on all missing fields in Cargo.toml
When using wasm-pack via `wasm-pack init` only one field at a time is output as to what's missing from the Cargo.toml file. It would be nicer if all the fields missing were printed out all at once!

mgattozzi:
Because the toml error only spits out the error for the first thing it
runs into it doesn't parse all of the missing fields. We first serialize
the data to a generic toml type to check this ourselves. If everything
is there we serialize directly to the type we expected. Otherwise we
throw an error with all the fields that are missing.

Closes #43 

I think we should make a custom error type at some point for #12 but that would be a little outside the scope of the PR. I'd like to test with assert-cli but that's still blocked for windows, however I included a test fixture for it to eventually be added in and one that can be used to test locally. Output looks like this:

2018-03-16-114359_1799x317_scrot

### DIFF
--- a/src/manifest.rs
+++ b/src/manifest.rs
@@ -4,6 +4,7 @@ use std::io::prelude::*;
 use console::style;
 use emoji;
 use failure::Error;
+use indicatif::MultiProgress;
 use progressbar;
 use serde_json;
 use toml;
@@ -17,20 +18,20 @@ struct CargoManifest {
 struct CargoPackage {
     name: String,
     authors: Vec<String>,
-    description: String,
+    description: Option<String>,
     version: String,
-    license: String,
-    repository: String,
+    license: Option<String>,
+    repository: Option<String>,
 }
 
 #[derive(Serialize)]
 struct NpmPackage {
     name: String,
     collaborators: Vec<String>,
-    description: String,
+    description: Option<String>,
     version: String,
-    license: String,
-    repository: Repository,
+    license: Option<String>,
+    repository: Option<Repository>,
     files: Vec<String>,
 }
 
@@ -64,10 +65,10 @@ impl CargoManifest {
             description: self.package.description,
             version: self.package.version,
             license: self.package.license,
-            repository: Repository {
+            repository: self.package.repository.map(|repo_url| Repository {
                 ty: "git".to_string(),
-                url: self.package.repository,
-            },
+                url: repo_url,
+            }),
             files: vec![js_file, wasm_file],
         }
     }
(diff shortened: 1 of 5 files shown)


sendilkumarn:
I would really like to see some icons here 😁 otherwise looks good to me 👍

mgattozzi:
Maybe something like: ❌?

sendilkumarn:
There is already a emoji error available. you can reuse that

mgattozzi:
Cool! Cleaned up some of the logic so things look a little less noisy in terms of word density and I added the emoji. Doesn't show up on my terminal despite the font having support for it but I'm using alarcritty so I expect some bugs. Anyways this should be good now.

src/manifest.rs
```diff
@@ -45,7 +45,39 @@ fn read_cargo_toml(path: &str) -> Result<CargoManifest, Error> {
...
+
+    let maybe_missing_fields = vec![
+        contents.get("description").is_none(),
+        contents.get("license").is_none(),
+        contents.get("repository").is_none(),
+    ];
+
+    if maybe_missing_fields.contains(&true) {
+        let mut error_buffer = String::new();
+        if maybe_missing_fields[0] {
```
ashleygwilliams: i'm not the hugest fan of "magic numbers", i.e. these missing fields have magic numbers associated. could we do this with a data structure that has named fields?

src/manifest.rs
```diff
@@ -45,7 +45,39 @@ fn read_cargo_toml(path: &str) -> Result<CargoManifest, Error> {
...
+            error_buffer.push_str(&format!(
+                "  [ERROR] {} Missing field `description` for key `package`\n",
```
ashleygwilliams: these error messages are highly redundant, perhaps we have a function to make the message so we don't have to repeat it?

ashleygwilliams: we also should add that this field is missing in the `Cargo.toml` as it's not entirely clear that that is what this means.

mgattozzi: Yeah let's make it clear then!

mgattozzi:
Oh okay! I made it a hard error because that's how it currently works. If it's optional I can rework the code to do that and put out a warning!

src/manifest.rs
mgattozzi: Like a small struct? Sure, I wasn't exactly happy about them either.

mgattozzi:
Here is what it looks like now. I think we should probably abstract warning and error handling code at some point but this works for now. So main thing is that it now makes the fields optional when dealing with deserialization and serialization. This means they're null when serialized to json!

One question I have is if the optional fields can have null and npm will accept that as is. Let me know what you think about these changes @ashleygwilliams 

2018-03-20-162650_1819x527_scrot

ashleygwilliams:
@mgattozzi oh gross it looks like travis is throwing up at how i pass arguments to cargo fmt? (weird to do this now when i haven't touched the file in a month)

anyways i've been holding off review because i thought travis was broken, which it is, but not because of you...

i'll sort that and get back to you with all the reviews i owe u!

sendilkumarn:
@ashleygwilliams this seems that there are already reports related to this. Can we try running it again? It should be running correctly 

https://github.com/rust-lang/cargo/issues/5208

ashleygwilliams:
ah thanks @sendilkumarn ! i *thought* it was a regression in cargo but i haven't finished my coffee this morning so i was willing to just blame myself about it. will kick off some re-runs now!

mgattozzi:
Yeah let me push a fmt fix and then you can do whatever design stuff you want to do for it :D

mgattozzi:
@ashleygwilliams I did a cargo fmt before sending it in and fixed the file conflict as well. I think this should be good to go once CI passes

PR #67, opened 2018-03-27
init pack and publish

jamiebuilds:
an effort was made

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -6,17 +6,18 @@ extern crate serde_derive;
 extern crate serde_json;
 extern crate toml;
 
-pub mod build;
 pub mod bindgen;
-pub mod readme;
-pub mod manifest;
-pub mod progressbar;
+pub mod build;
 pub mod emoji;
+pub mod manifest;
+pub mod npm;
+pub mod progressbar;
+pub mod readme;
 
 use std::fs;
 
-use failure::Error;
 use console::style;
+use failure::Error;
 
 pub fn create_pkg_dir(path: &str) -> Result<(), Error> {
     let step = format!(
--- a/src/npm.rs
+++ b/src/npm.rs
@@ -1,0 +1,32 @@
+use emoji;
+use std::process::Command;
+
+pub fn npm_pack(path: &str) {
+    let pkg_file_path = format!("{}/pkg", path);
+    let output = Command::new("npm")
+        .current_dir(pkg_file_path)
+        .arg("pack")
+        .output()
+        .unwrap_or_else(|e| panic!("{} failed to execute process: {}", emoji::ERROR, e));
+    if !output.status.success() {
+        let s = String::from_utf8_lossy(&output.stderr);
+        print!("{}  npm_pack failed and stderr was:\n{}", emoji::ERROR, s);
+    }
+}
+
+pub fn npm_publish(path: &str) {
+    let pkg_file_path = format!("{}/pkg", path);
+    let output = Command::new("npm")
+        .current_dir(pkg_file_path)
+        .arg("publish")
+        .output()
+        .unwrap_or_else(|e| panic!("{} failed to execute process: {}", emoji::ERROR, e));
+    if !output.status.success() {
+        let s = String::from_utf8_lossy(&output.stderr);
+        print!(
+            "{}  npm_publish failed and stderr was:\n{}",
+            emoji::ERROR,
+            s
+        );
+    }
+}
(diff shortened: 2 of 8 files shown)


jamiebuilds:
If you want to give me requirements for what you want these commands to do I can make more efforts

src/lib.rs
```diff
@@ -6,17 +6,18 @@ extern crate serde_derive;
...
+pub mod npm;
 pub mod progressbar;
-pub mod emoji;
+pub mod readme;
 
 use std::fs;
 
-use failure::Error;
 use console::style;
+use failure::Error;
```
jamiebuilds: my editor just started reordering these when i installed the rust ide.. 🤷‍♀️

ashleygwilliams:
zomg yay thank u @jamiebuilds - for context the travis PR is failing based on the cargo fmt lint. let me know if you want help setting that up!

src/lib.rs
ashleygwilliams: it's super fine <3

ashleygwilliams:
oh also to address your original concern: my intention was to map those commands directly to the npm ones so you have intuited 100% correctly! in the long run i would like to rewrite these - so if you are interested, that is where i plan to spend the most time next, so let me know if you are interested!

short term i think shelling out is a totally valid strategy, the one thing i'd like to see is some updates to the README to indicate that these are implemented and depend on a system npm.

src/npm.rs
```diff
@@ -0,0 +1,32 @@
+use emoji;
+use std::process::Command;
+
+pub fn npm_pack(path: &str) {
+    let pkg_file_path = format!("{}/pkg", path);
```
sendilkumarn: wow @jamiebuilds didn't expect you here

src/npm.rs
```diff
@@ -0,0 +1,32 @@
+use emoji;
+use std::process::Command;
+
+pub fn npm_pack(path: &str) {
```
sendilkumarn: Can we refactor this into a single method? Do you foresee that it will be changed at any time?

jamiebuilds: Yeah, there are lots of other options you can pass to publish in the future

sendilkumarn: if they are just adding new arguments update, I would suggest to move them to a single function with `struct`

ignore if it is not.

jamiebuilds: dont think its going to be that easy https://docs.npmjs.com/cli/publish

src/lib.rs
mgattozzi: I like having those fields alphabetical personally so this seems great to me :D

src/npm.rs
```diff
@@ -0,0 +1,32 @@
...
+    let pkg_file_path = format!("{}/pkg", path);
+    let output = Command::new("npm")
+        .current_dir(pkg_file_path)
+        .arg("pack")
+        .output()
+        .unwrap_or_else(|e| panic!("{} failed to execute process: {}", emoji::ERROR, e));
```
mgattozzi: I don't think we should be panicking here since main allows propogating up errors. I would refactor this method to do `map_err` for this error message, then use ? to propogate it up and make the method return a Result<(), failure::Error>

src/npm.rs
```diff
@@ -0,0 +1,32 @@
...
+    if !output.status.success() {
+        let s = String::from_utf8_lossy(&output.stderr);
+        print!("{}  npm_pack failed and stderr was:\n{}", emoji::ERROR, s);
```
mgattozzi: Which would allow you to return an Err here else Ok(())!

src/npm.rs
```diff
@@ -0,0 +1,32 @@
...
+    }
+}
+
+pub fn npm_publish(path: &str) {
+    let pkg_file_path = format!("{}/pkg", path);
+    let output = Command::new("npm")
+        .current_dir(pkg_file_path)
+        .arg("publish")
+        .output()
+        .unwrap_or_else(|e| panic!("{} failed to execute process: {}", emoji::ERROR, e));
```
mgattozzi: Same idea as above! :D

src/npm.rs
ashleygwilliams: i would recommend not moving them into a single struct. i much prefer the way it is set up now!

src/npm.rs
jamiebuilds: I just copied the behavior from existing code, this is copy and pasted

mgattozzi: Oh is it? Hmm then this might need to be a separate Pr then to just go through and change it all. :) I'll leave that call up to @ashleygwilliams

ashleygwilliams: yeah let's merge this as is and then do a refactor that hits all the points. PRs welcome :D

jamiebuilds:
🔥

PR #76, opened 2018-04-04
update to wasm bindgen 0.2.0

### LINKED ISSUE #74: update wasm-bindgen dep to 0.2.0

sendilkumarn:
fixes #74

### DIFF
--- a/README.md
+++ b/README.md
@@ -59,12 +59,12 @@ if you want to publish packages, you'll also need an account on [npm] and have
     crate-type = ["cdylib"]
 
     [dependencies]
-    wasm-bindgen = "0.1"
+    wasm-bindgen = "0.2"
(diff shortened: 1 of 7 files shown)


ashleygwilliams:
hey so some context: this project uses wasm-bindgen 2 ways... one as a lib in the examples (which you've handled excellently) and then one as a CLI tool- which still needs to be updated! let me know if you want help- the `bindgen.rs` file is where the update needs to be made :)

sendilkumarn:
But `bindgen.rs` don't have any version related information right. Do you mean add the version there ?

ashleygwilliams:
ah. hrm. yeah i guess adding the version would be a pain. ok! then i think this is good! will merge. thanks!

PR #78, opened 2018-04-06
Refactor main.rs

Andy-Bell:
This pull request is designed to move all the functionality out of the main.rs binary and into the main library itself - for testability and to hopefully provide some help with https://github.com/ashleygwilliams/wasm-pack/issues/25 .

Currently throws warning due to line 11 in command.rs - unused import of quicli - import is used for the StructOpt derive macro, so is needed to be imported, but due to no structs/functions from quicli being usied it still throws this warning.

### DIFF
--- a/src/command.rs
+++ b/src/command.rs
@@ -1,0 +1,100 @@
+use bindgen;
+use build;
+use console::style;
+use emoji;
+use failure::Error;
+use indicatif::HumanDuration;
+use manifest;
+use npm;
+use readme;
+use PBAR;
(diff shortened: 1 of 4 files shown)


ashleygwilliams:
curious about ur thoughts @mgattozzi

mgattozzi:
@ashleygwilliams I'm of the opinion that `main.rs` should really only be a file that sets up and drives the program and was going to suggest we do this at some point anyways. It makes things easier to reason about so consider this an LGTM

PR #79, opened 2018-04-12
Don't install wasm-bindgen if it already exists

data-pup:
I started working on some changes to try and fix the problems mentioned in #51. Per the discussion on that issue, I am opening up a PR so that the changes made can be reviewed and discussed. I don't think this is ready to merge _quite_ yet, but I have some questions about testing, cross-platform issues (specifically Windows), and general style.

This change adds a function called `wasm_bindgen_installed`, which will return a boolean value representing whether or not `wasm-bindgen` exists in the path. The `cargo_install_wasm_bindgen` will only progress to the installation process if `wasm_bindgen_installed` returns false.

### Testing

I am not exactly sure how to design test cases for this, since it depends on the environment. I did test this manually and things seemed to be working correctly, but ideally some test cases would make sure that this behaves correctly.

### Windows

Windows uses `;` as a path separator, in contrast to the `:` separator on Linux. Should I add a separator variable using the `cfg!` macro to set this accordingly? Are there other compatibility issues to consider aside from that?

### Style

Ashley sent me a link to [this] SO answer regarding how to check if a command is in the path, which I used as a starting point for this. The code below uses `map` and `fold` instead of a for loop as shown in the SO answer. Is there a preference regarding that? I do not mind rewriting this to follow an imperative style.

[this]: https://stackoverflow.com/questions/35045996/check-if-a-command-is-in-path-executable-as-process

P.S. Thank you for being welcoming to contributions! I am still a little new to helping out with open source projects, so I really appreciate it. :)

Update: The Travis build did not pass, this seems to have failed: `The command "cargo fmt -- --write-mode diff" exited with 4.` Is that a formatting issue?

### DIFF
--- a/src/bindgen.rs
+++ b/src/bindgen.rs
@@ -3,8 +3,16 @@ use console::style;
 use emoji;
 use std::{env, fs, process::Command};
 
+#[cfg(target_family = "windows")]
+static PATH_SEP: &str = ";";
+
+#[cfg(not(target_family = "windows"))]
+static PATH_SEP: &str = ":";
+
 pub fn cargo_install_wasm_bindgen() {
-    if wasm_bindgen_installed() { return; }
+    if wasm_bindgen_installed() {
+        return;
+    }
     let step = format!(
         "{} {}Installing WASM-bindgen...",
         style("[6/7]").bold().dim(),
@@ -40,13 +48,13 @@ pub fn wasm_bindgen_build(path: &str, name: &str) {
 
 fn wasm_bindgen_installed() -> bool {
     if let Ok(path) = env::var("PATH") {
-        return path.split(":")
+        path.split(PATH_SEP)
             .map(|p: &str| -> bool {
-                let prog_str = format!("{}/{}", p, "wasm-bindgen");
+                let prog_str = format!("{}/wasm-bindgen", p);
                 fs::metadata(prog_str).is_ok()
             })
-            .fold(false, |res, b| res || b);
+            .fold(false, |res, b| res || b)
     } else {
-        return false;
+        false
     }
 }
--- a/src/bindgen.rs
+++ b/src/bindgen.rs
@@ -1,9 +1,10 @@
 use PBAR;
 use console::style;
 use emoji;
-use std::process::Command;
+use std::{env, fs, process::Command};
 
 pub fn cargo_install_wasm_bindgen() {
+    if wasm_bindgen_installed() { return; }
     let step = format!(
         "{} {}Installing WASM-bindgen...",
         style("[6/7]").bold().dim(),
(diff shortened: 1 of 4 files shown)


src/bindgen.rs
```diff
@@ -36,3 +37,16 @@ pub fn wasm_bindgen_build(path: &str, name: &str) {
         .unwrap_or_else(|e| panic!("{} failed to execute process: {}", emoji::ERROR, e));
     pb.finish();
 }
+
+fn wasm_bindgen_installed() -> bool {
+    if let Ok(path) = env::var("PATH") {
+        return path.split(":")
```
mgattozzi: Since PATH is just a String you can test by doing path.contains("wasm-bindgen")

mgattozzi: You also don't need to use return just leave off the semicolon! :D 
Rust will always return a value for the last expression without a semicolon. We use return when we want to exit a function early much like you did in the file above!

src/bindgen.rs
```diff
@@ -36,3 +37,16 @@ pub fn wasm_bindgen_build(path: &str, name: &str) {
...
+            .map(|p: &str| -> bool {
+                let prog_str = format!("{}/{}", p, "wasm-bindgen");
+                fs::metadata(prog_str).is_ok()
+            })
+            .fold(false, |res, b| res || b);
+    } else {
+        return false;
```
mgattozzi: Same thing here you can just put false!

src/bindgen.rs
```diff
@@ -36,3 +37,16 @@ pub fn wasm_bindgen_build(path: &str, name: &str) {
         .unwrap_or_else(|e| panic!("{} failed to execute process: {}", emoji::ERROR, e));
     pb.finish();
 }
+
+fn wasm_bindgen_installed() -> bool {
+    if let Ok(path) = env::var("PATH") {
```
mgattozzi: Normally I'd say we should handle error cases but if your PATH variable isn't set I think you have bigger problems so I think this is fine for now. I plan on going through and cleaning up errors in general later on and I think that is outside the scope of this PR

mgattozzi:
Also your Travis CI build failed because of the need to run rust-fmt. Do you know how to do that or do you have installed? If you don't I can write up some instructions for you

src/bindgen.rs
mgattozzi: Oh you know what you're right. PATH contains directories of binaries that can be run not the binaries themselves! You'll need to traverse them to find it. I'd like so say we can assume it's in .cargo/bin and just check there, which is easier. However, I feel we should probably check each one like you've done already.

Regarding cross platform you can do [conditional](https://doc.rust-lang.org/book/first-edition/conditional-compilation.html) compilation depending on the platform so we can set a constant in the file and use it here to say "on this platform use : and on this one use" whatever it is.

mgattozzi:
awesome! you should be able to add it with rustup and then you'll be able to just do cargo fmt and it'll do it all for you!

mgattozzi:
These changes look great @data-pup! I believe this should work just from reading the code, but it would be good to confirm. I have a windows and linux build I can try these on at some point. Also you'll need to format the changes but I believe this should be good. @ashleygwilliams tagging you in case you haven't seen this yet

data-pup:
I added some edits in a second commit containing some of the changes discussed above. I installed and ran rust-fmt, but it seems like the Travis build is still failing? I am not sure what to do about that :confused:

mgattozzi:
You'll need to run it and commit the changes for it. Again. My guess is it didn't run on everything somehow.

PR #82, opened 2018-04-16
Add PR template to the repo

mgattozzi:
We want to make sure people format their code as well as reference the
issue that's going to be closed as well so we don't have to close the
issue manually! This PR adds a template for PRs that includes these
points as a checklist to ensure future contributors know what they need
to do!

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -1,0 +1,5 @@
+Make sure these boxes are checked!
+
+- [ ] You have the latest version of `rustfmt` installed
+- [ ] You ran `rustfmt` on the code base before submitting
+- [ ] You reference which issue is being closed in the PR text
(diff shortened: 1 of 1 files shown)


.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -0,0 +1,5 @@
+Make sure these boxes are checked!
```
ashleygwilliams: let's add something like "thanks so much!" and like an emoji or two. i would also add the lines:

```
$ rustup override set nightly
```

ashleygwilliams:
this is a great idea! thanks :D

.github/PULL_REQUEST_TEMPLATE.md
mgattozzi: Sounds good. Throwing in emojis and instructions is easy to do!

PR #90, opened 2018-04-18
feat(refactor/early abort): Exit early on failure

### LINKED ISSUE #71: refactor out the panics
as per @mgattozzi's comments on #67 and my own nagging gut, let's refactor out those panics when we are shelling out so we can have nicer error handling

### LINKED ISSUE #86: if compilation fails the program should exit

mgattozzi:
This commit does quite a few things in order to get this to work:

1. We move all of the code dealing with knowing which command to run
   into it's own function. This wrapper command allows us to always
   close out PBAR before dumping error output. This fixes a problem
   where stderr and stdout were borked and not printing out error
   messages correctly.
2. We then refactor the code that has a panic to return early with that
   error message.
3. If the command we ran errored, we print out with PBAR that there was
   an error with the program we ran (not wasm-pack itself) then dump the
   stderr from the command to the actual stderr

This means we can abort early on without continuing any of the other
parts of wasm-pack and let the user know what the error was rather than
just saying "There's an error"

Closes #86 
Closes #71

```bash
$ rustup override set nightly
```

### DIFF
--- a/src/bindgen.rs
+++ b/src/bindgen.rs
@@ -1,24 +1,31 @@
 use console::style;
 use emoji;
+use failure::Error;
 use std::process::Command;
 use PBAR;
 
-pub fn cargo_install_wasm_bindgen() {
+pub fn cargo_install_wasm_bindgen() -> Result<(), Error> {
     let step = format!(
         "{} {}Installing WASM-bindgen...",
         style("[6/7]").bold().dim(),
         emoji::DOWN_ARROW
     );
     let pb = PBAR.message(&step);
-    let _output = Command::new("cargo")
+    let output = Command::new("cargo")
         .arg("install")
         .arg("wasm-bindgen")
-        .output()
-        .unwrap_or_else(|e| panic!("{} failed to execute process: {}", emoji::ERROR, e));
+        .output()?;
     pb.finish();
+    if !output.status.success() {
+        let s = String::from_utf8_lossy(&output.stderr);
+        PBAR.error("npm_publish failed");
+        bail!(format!("stderr was {}", s));
+    } else {
+        Ok(())
+    }
 }
 
-pub fn wasm_bindgen_build(path: &str, name: &str) {
+pub fn wasm_bindgen_build(path: &str, name: &str) -> Result<(), Error> {
     let step = format!(
         "{} {}Running WASM-bindgen...",
         style("[7/7]").bold().dim(),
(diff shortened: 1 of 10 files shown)


src/bindgen.rs
```diff
@@ -1,24 +1,31 @@
...
         .arg("install")
         .arg("wasm-bindgen")
-        .output()
-        .unwrap_or_else(|e| panic!("{} failed to execute process: {}", emoji::ERROR, e));
+        .output()?;
     pb.finish();
+    if !output.status.success() {
+        let s = String::from_utf8_lossy(&output.stderr);
+        PBAR.error("npm_publish failed");
+        bail!(format!("stderr was {}", s));
```
mgattozzi: I should probably make these have a `:\n` before the `{}`

src/bindgen.rs
```diff
@@ -1,24 +1,31 @@
...
+    let output = Command::new("cargo")
         .arg("install")
         .arg("wasm-bindgen")
-        .output()
-        .unwrap_or_else(|e| panic!("{} failed to execute process: {}", emoji::ERROR, e));
+        .output()?;
     pb.finish();
+    if !output.status.success() {
+        let s = String::from_utf8_lossy(&output.stderr);
+        PBAR.error("npm_publish failed");
```
ashleygwilliams: don't think this is `npm_publish`. this is `cargo install wasm bindgen`.

src/bindgen.rs
```diff
@@ -27,12 +34,18 @@ pub fn wasm_bindgen_build(path: &str, name: &str) {
...
         .arg(&wasm_path)
         .arg("--out-dir")
         .arg("./pkg")
-        .output()
-        .unwrap_or_else(|e| panic!("{} failed to execute process: {}", emoji::ERROR, e));
+        .output()?;
     pb.finish();
+    if !output.status.success() {
+        let s = String::from_utf8_lossy(&output.stderr);
+        PBAR.error("npm_publish failed");
```
ashleygwilliams: same here, this is wasm_bindgen_build

src/command.rs
```diff
@@ -46,19 +65,19 @@ pub fn create_pkg_dir(path: &str) -> result::Result<(), Error> {
     Ok(())
 }
 
-pub fn init(path: Option<String>, scope: Option<String>) -> result::Result<(), Error> {
```
ashleygwilliams: yay

mgattozzi:
@ashleygwilliams Yeah I'd like the more human readable stuff so that's doable!

mgattozzi:
Sounds good to me! Oh. You know what it is. My cargo fmt was updated on my desktop not my laptop -_-

Issue #83, opened 2018-04-16
add npm login command


ashleygwilliams:
we have pack and publish wrapped up, we might as well wrap login as well

djfarly:
I'd like to try and implement this – if you mind to bare with me, this my first rust project. 🐣

djfarly:
One question upfront: Since [npm calls this "adduser"](https://docs.npmjs.com/cli/adduser). Should this be the name in code here too?
Of course with alias "login" and "add-user".

mgattozzi:
You could use one or other in the code but I think going with login might be the best, as adduser is a bit counter intuitive as to what it does.

If you want to take a crack at it here's what needs to get done!

1) Below [here](https://github.com/ashleygwilliams/wasm-pack/blob/master/src/command.rs#L31) you'll need to add a value `Login`. It'll need an attribute  `#[structopt(name = "login")]` above it so that the tool knows that this represents the login command. I'm not sure what we want to do for the doc comment but I think @ashleygwilliams should know! :grinning: 
2) You'll need to now match [here](https://github.com/ashleygwilliams/wasm-pack/blob/40a5e66d2a41d8b736ff0f0240b40046b13847ff/src/command.rs#L40) with the new `Login` value. It'll execute a command `npm_login()`
3) You'll need to actually execute the command. Youll want to add another function [here](https://github.com/ashleygwilliams/wasm-pack/blob/40a5e66d2a41d8b736ff0f0240b40046b13847ff/src/npm.rs#L20) that's simmilar to the other functions but it takes no input and you don't want to use the current dir function call.

This should be all you need to do! To test your work just type `cargo run -- login` and it should bring up the prompts to actually login to npm. If this compiles and doesn't happen ping the issue here as I have an idea as to why that might happen.

Hopefully this will be enough for you to get it implemented @djfarly. If you get stuck let us know!

djfarly:
Oops sorry - I haven’t read your comment before eagerly starting to work 🙄
I think I basically did what you’re describing. I can rename the command from adduser to login. But let’s discuss this in the pr maybe? :)

PR #100, opened 2018-04-20
Wrap npm adduser

djfarly:
Implements #83 

This wraps npm adduser. I’ve copied the argument descriptions from npm.
By the way: why are all sub commands marked as NOT IMPLEMENTED?

I’m a little unsure about some of the solutions I used here.

1. `if let` inside `arg()` seems wired.
2. passing 4 args down multiple times feels strange

I took the liberty to add some spacing inside the command enum, since it really confused me for a few minutes.

Does this seem okay otherwise?

Cheers!

```bash
$ rustup override set nightly
````

### DIFF
--- a/src/command.rs
+++ b/src/command.rs
@@ -156,7 +156,7 @@ fn login(
 
     npm::npm_login(&registry, &scope, always_auth, &auth_type)?;
 
-    PBAR.message(&format!("👋  logged you in to {}!", registry));
+    PBAR.message(&format!("👋  logged you in!"));
     Ok(())
 }
 
--- a/src/command.rs
+++ b/src/command.rs
@@ -32,9 +32,9 @@ pub enum Command {
     /// 🎆  pack up your npm package and publish! [NOT IMPLEMENTED]
     Publish { path: Option<String> },
 
-    #[structopt(name = "adduser", alias = "login", alias = "add-user")]
-    /// 👤  Add a registry user account! [NOT IMPLEMENTED]
-    Adduser {
+    #[structopt(name = "login", alias = "adduser", alias = "add-user")]
+    /// 👤  Add a registry user account! (aliases: adduser, add-user) [NOT IMPLEMENTED]
+    Login {
         #[structopt(long = "registry", short = "r")]
         /// Default: 'https://registry.npmjs.org/'.
         /// The base URL of the npm package registry. If scope is also specified, this registry will only be used for packages with that scope. scope defaults to the scope of the project directory you're currently in, if any.
@@ -64,12 +64,12 @@ pub fn run_wasm_pack(command: Command) -> result::Result<(), Error> {
         Command::Init { path, scope } => init(path, scope),
         Command::Pack { path } => pack(path),
         Command::Publish { path } => publish(path),
-        Command::Adduser {
+        Command::Login {
             registry,
             scope,
             always_auth,
             auth_type,
-        } => adduser(registry, scope, always_auth, auth_type),
+        } => login(registry, scope, always_auth, auth_type),
     };
 
     match status {
@@ -146,15 +146,15 @@ fn publish(path: Option<String>) -> result::Result<(), Error> {
     Ok(())
 }
 
-fn adduser(
+fn login(
     registry: Option<String>,
     scope: Option<String>,
     always_auth: bool,
     auth_type: Option<String>,
 ) -> result::Result<(), Error> {
-    npm::npm_adduser(registry, scope, always_auth, auth_type)?;
+    npm::npm_login(registry, scope, always_auth, auth_type)?;
 
-    PBAR.one_off_message("👋  added registry user account!");
+    PBAR.one_off_message("👋  logged you in!");
     Ok(())
 }
 
--- a/src/command.rs
+++ b/src/command.rs
@@ -154,7 +154,7 @@ fn login(
 ) -> result::Result<(), Error> {
     npm::npm_login(registry, scope, always_auth, auth_type)?;
 
-    PBAR.one_off_message("👋  logged you in!");
+    PBAR.message("👋  logged you in!");
     Ok(())
 }
 
--- a/src/command.rs
+++ b/src/command.rs
@@ -23,12 +23,38 @@ pub enum Command {
         #[structopt(long = "scope", short = "s")]
         scope: Option<String>,
     },
+
     #[structopt(name = "pack")]
     /// 🍱  create a tar of your npm package but don't publish! [NOT IMPLEMENTED]
     Pack { path: Option<String> },
+
     #[structopt(name = "publish")]
     /// 🎆  pack up your npm package and publish! [NOT IMPLEMENTED]
     Publish { path: Option<String> },
+
+    #[structopt(name = "adduser", alias = "login", alias = "add-user")]
+    /// 👤  Add a registry user account! [NOT IMPLEMENTED]
+    Adduser {
+        #[structopt(long = "registry", short = "r")]
+        /// Default: 'https://registry.npmjs.org/'.
+        /// The base URL of the npm package registry. If scope is also specified, this registry will only be used for packages with that scope. scope defaults to the scope of the project directory you're currently in, if any.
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

**Comment by djfarly** [comment]
Woooo 😲 #100 💯 🙊

File: src/command.rs [hunk 1]
```diff
@@ -23,12 +23,38 @@ pub enum Command {
...
+
     #[structopt(name = "pack")]
     /// 🍱  create a tar of your npm package but don't publish! [NOT IMPLEMENTED]
     Pack { path: Option<String> },
+
     #[structopt(name = "publish")]
     /// 🎆  pack up your npm package and publish! [NOT IMPLEMENTED]
     Publish { path: Option<String> },
+
+    #[structopt(name = "adduser", alias = "login", alias = "add-user")]
```
mgattozzi: Ah okay you did all them here this is good. I'd swap login and adduser in the arguments

djfarly:
Also I seem to have missed a white space in the formatting.

src/command.rs
```diff
@@ -38,6 +64,12 @@ pub fn run_wasm_pack(command: Command) -> result::Result<(), Error> {
         Command::Init { path, scope } => init(path, scope),
         Command::Pack { path } => pack(path),
         Command::Publish { path } => publish(path),
+        Command::Adduser {
+            registry,
+            scope,
+            always_auth,
+            auth_type,
+        } => adduser(registry, scope, always_auth, auth_type),
```
mgattozzi: I feel like this should be the exception and just have npm::add_user be called directly. Though I also get having consistency. @ashleygwilliams thoughts?

src/npm.rs
```diff
@@ -31,3 +31,40 @@ pub fn npm_publish(path: &str) -> Result<(), Error> {
         Ok(())
     }
 }
+
+pub fn npm_adduser(
+    registry: Option<String>,
+    scope: Option<String>,
+    always_auth: bool,
+    auth_type: Option<String>,
+) -> Result<(), Error> {
```
mgattozzi: You could instead do:

```rust
let args = String::new();
if let Some(registry) = registry {
    args.push_str(&format!(" --registry {}", registry));
}
```
Do that for each one to generate the optional args then

```rust
let status = Command::new("npm")
     .arg("login")
     .arg(args)
     .output()?;
```
You'll want the output instead of success so we can dump to stderr like the above functions incase things go wrong.

Let me know if this doesn't make sense

djfarly: 👍 for combining the args into a string.
I tried using output here but it instantly failed the npm command, that's why I resorted to using status. It felt like output wasn't waiting for npm to do it's stdin and stdout stuff.
I'll try that again though.

src/command.rs
djfarly: sure can do. Shall we keep the function names as adduser?
I feel like changing them to login would make things less confusing.

djfarly:
I've went all in and renamed everything to login to have it consistent in this codebase.

Regarding output, here's what happens once I use it:
bildschirmfoto 2018-04-20 um 20 27 49

With status it works just fine. 😕

mgattozzi:
Hmmm okay so I suspected something like this was going to happen. So `npm login` doesn't let you pass in the parameters needed it always gets input from stdin to do it. I think we need to do something to let us interact with the program in order for it to work. Possibly a lock on stdin temporarily and passing that in to the program. Though I'm not sure off the top of my head. I haven't had enough coffee this morning so I might be missing something.

src/npm.rs
mgattozzi: Awesome glad that helped. As for stdin failing stuff I posted a comment below about it.

mgattozzi:
Wait are you saying you interacted with stdin fine when you used status instead of output?

djfarly:
Yes :) You can give it a try.

mgattozzi:
That's interesting! let me check this out locally and try something out

djfarly:
Interesting indeed - I’ve even read this section of the docs but missed the second half of that sentence 😢

I’ll have a look tomorrow! Thanks for digging it up.

mgattozzi:
It's a quick line that got buried. I tested the changes locally and it works perfectly so if you get the last bits in we should be able to merge it no problem!

ashleygwilliams:
FYI npm intends to separate out these commands in the future, as a result i would really prefer if this were just named login! thanks!

djfarly:
Sorry for the delay 😢 

- This should be rebased to current master now.
- Uses output instead of status + and shows stderr.

I could not test the latter… it seems pretty hard to make npm login actually fail.

src/command.rs
```diff
@@ -107,6 +139,18 @@ fn publish(path: Option<String>) -> result::Result<(), Error> {
...
 
+fn login(
+    registry: Option<String>,
+    scope: Option<String>,
+    always_auth: bool,
+    auth_type: Option<String>,
+) -> result::Result<(), Error> {
+    npm::npm_login(registry, scope, always_auth, auth_type)?;
+
+    PBAR.message("👋  logged you in!");
```
ashleygwilliams: for this, i think it'd be good to print what registry they were logged into :) (whatever was provided, or the default registry.npmjs.com)

src/command.rs
```diff
@@ -23,12 +23,38 @@ pub enum Command {
...
     #[structopt(name = "pack")]
     /// 🍱  create a tar of your npm package but don't publish! [NOT IMPLEMENTED]
     Pack { path: Option<String> },
+
     #[structopt(name = "publish")]
     /// 🎆  pack up your npm package and publish! [NOT IMPLEMENTED]
     Publish { path: Option<String> },
+
+    #[structopt(name = "login", alias = "adduser", alias = "add-user")]
+    /// 👤  Add a registry user account! (aliases: adduser, add-user) [NOT IMPLEMENTED]
```
ashleygwilliams: do we want this to still say not implemented? it looks very implemented here :)

ashleygwilliams: oh i see. there are some errant "NOT IMPLEMENTED" on the pack and publish commands... you can just remove that :)

src/npm.rs
```diff
@@ -31,3 +31,42 @@ pub fn npm_publish(path: &str) -> Result<(), Error> {
...
+    let output = Command::new("npm")
+        .arg("login")
+        .arg(args)
+        .stdin(Stdio::inherit())
+        .stdout(Stdio::inherit())
+        .output()?;
+
+    if !output.status.success() {
+        let s = String::from_utf8_lossy(&output.stderr);
+        PBAR.error("Registry user account login failed");
```
ashleygwilliams: I think it'd be useful to mention the registry they were trying to log into here as well :)

ashleygwilliams:
not for this PR but in the long run we mayyyy consider adding support for npmrc (which would potentially be a fallback for some of these CLI options) @yoshuawuyts wrote this crate to kick off some of that work https://github.com/yoshuawuyts/npmrc

mgattozzi:
@ashleygwilliams we should probably open up a separate issue for that if we haven't or at least a discussion ticket!

src/command.rs
djfarly: Unless we somehow want to filter and or parse the npm output…
For now i'd suggest keeping the wasm-pack output as concise as possible (without registry info). What do you think?

src/command.rs
djfarly: Will do - that already confused me :)

src/npm.rs
djfarly: Here it seems a little more helpful, because this way we don't depend on the stderr output of npm.

src/command.rs
djfarly: *gentlypush* :)

ashleygwilliams: i didn't realize it was redundant! no worries, feel free to not print it here! my main concern was that it was printed at all :)

djfarly:
I introduced a const for the default npm registry. That way we can show the correct one in the error msg, even when the default is used.

ashleygwilliams:
thank you so much for this! i really appreciate it :D

PR #105, opened 2018-04-23
fix(bindgen): don't rename to index

### LINKED ISSUE #102: don't rename the js file
Hello! I was following the tutorial for wasm-pack and got to this step: https://rust-lang-nursery.github.io/rust-wasm/wasm-pack/package-code.html

Compiling took a while and then printed this:

```
wasm-pack -v init --scope thallada
  [1/7] Adding WASM target...
  [2/7] Compiling to WASM...
  [3/7] Creating a pkg directory...
  [4/7] Writing a package.json...
| [5/7] Copying over your README...
| :-) [WARN]: origin crate has no README
  [6/7] Installing WASM-bindgen...
  [7/7] Running WASM-bindgen...
:-) Done in 1 minute
:-) Your WASM pkg is ready to publish at ./pkg
```

But, if I look at the `./pkg` directory, there's only a `package.json` file:

```
$ ls pkg/
package.json
```

Do you know what might have gone wrong?

```
$ npm --version
v9.11.1

$ npm --version
5.6.0

$ rustup show
Default host: x86_64-unknown-linux-gnu

installed toolchains
--------------------

stable-x86_64-unknown-linux-gnu
nightly-x86_64-unknown-linux-gnu (default)

installed targets for active toolchain
--------------------------------------

asmjs-unknown-emscripten
wasm32-unknown-emscripten
wasm32-unknown-unknown
x86_64-unknown-linux-gnu

active toolchain
----------------

nightly-x86_64-unknown-linux-gnu (default)
rustc 1.27.0-nightly (ac3c2288f 2018-04-18)
```

ashleygwilliams:
fixes #102

### DIFF
--- a/src/bindgen.rs
+++ b/src/bindgen.rs
@@ -1,7 +1,6 @@
 use console::style;
 use emoji;
 use failure::Error;
-use std::fs;
 use std::process::Command;
 use PBAR;
 
@@ -51,9 +50,6 @@ pub fn wasm_bindgen_build(path: &str, name: &str) -> Result<(), Error> {
         PBAR.error("wasm-bindgen failed to execute properly");
         bail!(format!("Details:\n{}", s));
     } else {
-        let js_file = format!("{}/pkg/{}.js", path, binary_name);
-        let index_file = format!("{}/pkg/index.js", path);
-        fs::rename(&js_file, &index_file)?;
         Ok(())
     }
 }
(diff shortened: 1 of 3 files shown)


alexcrichton:
FWIW if `index.js` is desired then it's actually relatively easy to rewrite imports of wasm modules (especially with parity-wasm), so that's at least an other option here although certainly not required if this strategy also works!

ashleygwilliams:
@mgattozzi it doesn't we should be all set by declaring it as main. @alexcrichton good to know, i really wanna cut a 0.2.0 today so i'll fix this now but look into rewriting the import and also making sure we have tests so we dont have a regression like this

mgattozzi:
Perfect! Then this definitely LGTM. Should we open a separate issue to address @alexcrichton's stuff later on when we have more time?

Issue #107, opened 2018-04-24
typescript support


ashleygwilliams:
take a `--typescript` flag to `wasm-bindgen` and add a `types` key to the `package.json`
http://www.typescriptlang.org/docs/handbook/declaration-files/publishing.html

robertohuertasm:
How have you thought to approach this? Trying to automatically generate the typings via some tool like [dts-gen](https://github.com/Microsoft/dts-gen) (which by the way, it doesn't seem to be supporting es6 at the moment)? or maybe you have something else in mind? :wink:

robertohuertasm:
oh! That's awesome! Indeed @ashleygwilliams message already talked about it. I didn't read it carefully and  read `wasm-pack` instead of `wasm-bindgen` 😱 

@mgattozzi many thanks for your clarification! 😃

ashleygwilliams:
solved by #109

PR #109, opened 2018-04-25
feat(command): expose --no-typescript option to init command

kwonoj:
- relates to #107 

I got so excited about #107 as typescript user so gave a quick stab, but not sure if this change is intended scope for issue itself. Feel freely close / or suggest as needed. 🙏 

This PR intends to ask bindgen to generate typescript definition as well, also expand `NpmPackage` struct to include `types` property into `package.json`. It currently takes naïve approach by specifying type definition to same entrypoint of `main`.

- [ ] You have the latest version of `rustfmt` installed and have your 
```bash
$ rustup override set nightly
```

### DIFF
--- a/src/bindgen.rs
+++ b/src/bindgen.rs
@@ -29,7 +29,7 @@ pub fn cargo_install_wasm_bindgen() -> Result<(), Error> {
     }
 }
 
-pub fn wasm_bindgen_build(path: &str, name: &str) -> Result<(), Error> {
+pub fn wasm_bindgen_build(path: &str, name: &str, disable_dts: bool) -> Result<(), Error> {
     let step = format!(
         "{} {}Running WASM-bindgen...",
         style("[7/7]").bold().dim(),
@@ -38,11 +38,18 @@ pub fn wasm_bindgen_build(path: &str, name: &str) -> Result<(), Error> {
     let pb = PBAR.message(&step);
     let binary_name = name.replace("-", "_");
     let wasm_path = format!("target/wasm32-unknown-unknown/release/{}.wasm", binary_name);
+    let dts_arg = if disable_dts == false {
+        "--typescript"
+    } else {
+        "--no-typescript"
+    };
+
     let output = Command::new("wasm-bindgen")
         .current_dir(path)
         .arg(&wasm_path)
         .arg("--out-dir")
         .arg("./pkg")
+        .arg(dts_arg)
         .output()?;
     pb.finish();
     if !output.status.success() {
--- a/src/command.rs
+++ b/src/command.rs
@@ -22,6 +22,8 @@ pub enum Command {
         path: Option<String>,
         #[structopt(long = "scope", short = "s")]
         scope: Option<String>,
+        #[structopt(long = "no-typescript")]
+        disable_dts: bool,
     },
     #[structopt(name = "pack")]
     /// 🍱  create a tar of your npm package but don't publish! [NOT IMPLEMENTED]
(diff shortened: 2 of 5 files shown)


mgattozzi:
Hey @kwonoj thanks for the PR! So this is definitely a good start but there are somethings in #107 that weren't exactly mentioned which is this should be optional! The invocation would be along the lines of `wasm-pack init --typescript` meaning that instead of the default behavior it adds the typescript flag to wasm-bindgen. The way you have it now typescript is the default.

Your enthusiasm is great so if you want to tackle 107 completely let's get this PR into shape before merging! Do you need help in understanding what needs to be changed in order to get what I described above? I'm more than happy to type those instructions out for you :D

kwonoj:
@mgattozzi Thanks for guidance.

For matter of 

>instead of the default behavior it adds the typescript flag to wasm-bindgen. The way you have it now typescript is the default.

Is it necessarily have to be opt-in? in case of consumer point of npm packages, any JS user doesn't get interrupted by existence of type definition, while TS users will definitely want to have types most times. In which cases it'd be encouraged to generate npm packages doesn't include type definitions?

kwonoj:
(Clarifying, I can definitely change PR to enable opt in behavior - just wanted to understand usecases before getting into)

mgattozzi:
From my conversations with @ashleygwilliams my understanding is opt-in, but I didn't know about this behavior before. If none of that matters then maybe it can be implicitly put in but that's not my call.

kwonoj:
I see. Maybe it depends on how accurate generated type definition matches to original signature in rust code - but if assumption's it is correct enough, I think it'd better to have types by default. (Or can make opt-out to not to have it). Typescript definition's primary consumer is obviously typescript users, but also lot of tooling like vs code's lsp can understand those so plain js user gets additional benefits such as intellisense. Course actual js build pipeline doesn't affected by existence of type definition.

Still, lett me know if desired behavior should be opt in - I'll update PR accordingly.

Pauan:
It's not my decision to make either, but I would just like to point out that it's common (and recommended) to compile TypeScript projects to `.js` + `.d.ts`, so that way they can be used seamlessly with either ES6 or TypeScript. In particular, it is *not* recommended to publish `.ts` files.

kwonoj:
@Pauan definitely, and discusdions about whthere include d.ts or not. Wasm-bindgen's flag generates 'd.ts' as type definition and doesn't generate typescript files, reason I thought it can be default.

robertohuertasm:
I agree with @kwonoj. I think `.d.ts` files should be created by default. As he states it only adds value and doesn't interfere with users consuming it from a plain js project; on the contrary, it will help their IDE to provide full intellisense (like in VSCode for instance).

ashleygwilliams:
hi! i definitely do not think that typescript should *NOT* be the default. i really like having the option, and am open to promoting it to default at some point if that seems appropriate, but for landing this i would like to see it behind a `--typescript` flag! i'm certainly open to changing my mind - but let's start small, land this behind a flag, and then continue this convo in a new issue. i'll open one :) thanks everyone for your thoughtful convo! i'm excited to chat more about this.

mgattozzi:
Sweet. @kwonoj let me know if you need help at all to do this! :D

kwonoj:
Ok, sounds fair - I'll try to update PR to have flags.

ashleygwilliams:
@kwonoj actually.... i think u might have convinced me .... https://github.com/ashleygwilliams/wasm-pack/issues/114#issuecomment-384856738 lemme think on this, but i think i might be down to have this land sans-flag, you make some very compelling points

kwonoj:
@ashleygwilliams per discussion around #114, PR now changed behavior to introduce `--no-typescript` instead to opt-out from wasm-bindgen type generation.

kwonoj:
Actually, this PR requires bump up `wasm-bindgen` to use `--no-typescript` args.

mgattozzi:
@kwonoj You are correct. I have verified that it does produce the typescript files though in it's current state. I'm going to tag this as blocked until we get a 0.2.7 release for wasm-bindgen.

cc/ @alexcrichton

alexcrichton:
I've now published 0.2.7!

ashleygwilliams:
THANK U!! this is awesome :D

PR #116, opened 2018-04-27
Load dependencies from `wasm-bindgen`

### LINKED ISSUE #101: Coordinating with wasm-bingden about versions and list of NPM dependencies
This is related to https://github.com/rust-lang-nursery/rust-wasm/issues/34 and https://github.com/rust-lang-nursery/rust-wasm/issues/36, but I'd like to propose a strawman idea here.

Right now wasm-bindgen has the ability to say where you're importing functionality from:

```rust
#[wasm_bindgen(module = "foo")]
extern {
    // ...
}
```

If `module = "..."` is omitted it's assumed to come from the ambient environment (aka the browser). Otherwise `foo` is interpreted as an ES module directive and that's where items are imported from in the generated JS that wasm-bindgen emits.

The wasm-bindgen output is pretty "dumb" right now in that it's just bland ES module information, it doesn't currently attempt to give extra information like verison requirements for NPM packages. But that's where wasm-pack comes in! Ideally wasm-bindgen would emit information that wasm-pack later reads, parses, and emits a package.json for.

So I'd propose something like:

* Add support for `version = "..."` next to `module = "..."` in wasm-bindgen
* For all `module = "..."` directives that *don't* start with `.`, require that `version` is listed
* Emit a new section in the wasm executable in the wasm that `wasm-bindgen` emits in the format:
  * The section name is `__wasm_pack_unstable`
  * The section format is a list of elements. Each element is prefixed with 4 bytes of how long it is and then the rest of the data is a JSON payload.
  * The JSON payload for each element is currently `{ "module": "string", "version": "string" }`
* When `wasm-pack` executes it slurps up the `__wasm_pack_unstable` section, if present, and emits these packages into the `package.json` that's generated.

An example of using this would look like:

```rust
#[wasm_bindgen(module = "moment", version = "2.0.0")]
extern {
    type Moment;
    fn moment() -> Moment;
    #[wasm_bindgen(method)]
    fn format(this: &Moment) -> String;
}

#[wasm_bindgen]
pub fn run() -> String {
    moment().format()
}
```

And `wasm-pack` would automatically generate a `package.json` with:

```json
{ 
   "dependencies": {
        "moment": "2.0.0"
    }
}
```

I'm definitely up for changing anything here, this is just an initial strawman! I'd love to hear others' thoughts on this as well.

alexcrichton:
This commit updates `wasm-pack` to read the `__wasm_pack_unstable` section of
wasm binaries that it exeutes over. This section is then parsed and NPM
dependencies are extracted, if any, and inserted into the generated
`package.json`.

Closes #101


src/manifest.rs
```diff
@@ -42,24 +47,101 @@ struct Repository {
...
+        V1 {
+            version: &'a str,
+            modules: Vec<(String, String)>,
+        },
+        Unknown {
+            version: &'a str,
+        },
+    }
+
+    println!("{:?}", wasm);
```
mgattozzi: Was this for debugging purposes?

ashleygwilliams:
lol haha i was gonna implement this, but i guess i don't have to anymore, will review later today.

src/manifest.rs
alexcrichton: Oops, indeed!

ashleygwilliams:
blocked pending RFC on npm deps expression

ashleygwilliams:
i'm going to close this because it is very stale and we've decided that this is not how we want to load npm deps.

PR #118, opened 2018-04-28
pretty-print panics

yoshuawuyts:
Heya!

This PR adds [human-panic](https://github.com/yoshuawuyts/human-panic), a pretty-printer & report generator for panics. I figured since this crate is targeted at Rust folks from all skill levels, it might be useful to make sure that we display friendly messages even if an unexpected error occurs.

I've also opened a sibling issue on `quicli` (https://github.com/killercup/quicli/issues/75) to discuss if it might make sense to include by default.

Hope this PR is useful!

## Refs
- [Human Panic example output](https://github.com/yoshuawuyts/human-panic#human-panic-output)


sendilkumarn:
Does it also have `inbuilt` emoji support 😉

mgattozzi:
@yoshuawuyts few quick questions:
Is this on stable or nightly? (Not that it matters much since we're using wasm32)
It works by grabbing authorship data from Cargo.toml right?

yoshuawuyts:
@mgattozzi it should work on stable!

PR #120, opened 2018-05-01
feat(error_msg): Make custom module for errors

### LINKED ISSUE #12: handle the specific errors specifically

mgattozzi:
This does two big things:

- We add custom error messages and conversions from other error types
  into our own error type. This means we can create or modify more
  fine grained errors rather than using whatever `failure::Error` gives
  us. The first one we have is a `Cli` error, whereby calls to things
  like wasm-bindgen or npm failed.
- We also make it so that `PBAR.error()` is called only on exit. We grab
  a reference to whatever error gets passed up the chain, have it
  generate a message to be printed as an error to PBAR, close up PBAR
  like normal, and then print our actual error message to stderr!

The main benefits of this is that as long as we return an error, PBAR
will print and close itself up and we don't need to call it every time
or forget that we would need to call it, and we put error details
specifically into our error!

Closes #12

```bash
$ rustup override set nightly
```

### DIFF
--- a/src/bindgen.rs
+++ b/src/bindgen.rs
@@ -1,6 +1,6 @@
 use console::style;
 use emoji;
-use failure::Error;
+use error::Error;
 use std::process::Command;
 use PBAR;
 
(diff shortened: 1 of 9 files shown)


mgattozzi:
Example output:

2018-05-01-170147_1965x667_scrot

mgattozzi:
Awesome! Glad you liked that refactor of error handling a bit! :D Another issue down!

PR #122, opened 2018-05-07
Context struct proof of concept

### LINKED ISSUE #25: reading the cargo toml 2x is kinda gross
*This issue is reserved for someone who has not yet contributed to this codebase.*

Currently we read the Cargo.toml twice! this is not preferred:

1. read it to write package.json
2. read it to pass crate name to wasm-bindgen

MENTORING INSTRUCTIONS (from @mgattozzi !):

1) Create a small expression or function to read the `Cargo.toml` file to a `String` and assign it to a variable, maybe something like `CARGO_TOML`
2) Replace instances of it being read with the variable you just created
3) Remove any code associated with reading the file

data-pup:
Per a previous conversation in irc, I looked into what adding a context struct would look like, if we wanted to pursue that idea further. This would fix #25, and only require that the `Cargo.toml` file be read once.

This would offer some benefits, such as not needing lazy_static for the progressbar, or for the manifest. The other nice thing is that the logic for running commands can be separated from the progressbar stuff, but that's a subjective benefit I suppose. This also means commands like `init` would work whether or not the path argument is given before or after the scope argument, I believe.

The downside, and I think it may be considerable, is that I think it would end up being a pretty __drastic__ change to the existing codebase. If we don't mind requiring that the path is the last argument given to the program, `lazy_static` would work quite fine. In that case, I think I can open a PR using the approach @mgattozzi mentioned, as that would be a simpler/smaller change to make.

Feel free to tag this as blocked. This was definitely speculative work to see what that approach would look like, and I figured I would put it here for the sake of helping y'all evaluate whether that would be a desirable road to go down. :)

By the way, I don't mind rebasing all of this if needed, I've left all the dev commits for now in case you'd like to build off of a specific commit etc.

---

```bash
$ rustup override set nightly
```


mgattozzi:
For context of the conversation we had:

lazy_static would read the env variables like so

```rust
use std::env::args;

// skip program name and subcommand
for arg in args().skip(2) {
// init code here
}
```

Where metadata can be used to see if the value is a path, if it is open up the file and read it in. If not try for each arg. If none of them match try `.` and if that doesn't work throw an error.

Doable but a bit hackish. I'm going to look through the Context stuff but this is a common pattern in Rust and used in things like cargo/rustc so I'm not opposed to this idea either.

data-pup:
_"I'm inclined to go with the hack for now then deffering this to a new issue and doing piecemeal updates to switch over to context with multiple PRs that are easier to review and test."_

@mgattozzi This sounds like a good plan to me too, pardon the amount of change here! I'll close this now, since it was mostly intended as a PoC :)

mgattozzi:
@data-pup I'm gonna open up an issue for this so we can discuss this further :D

data-pup:
Sounds good to me! ~[^.^]~

data-pup:
@mgattozzi Just wanted to mention that I can work on a branch that follows the env variable method you showed above, that shouldn't take too long. I'll get that done and open a (much smaller, lol) PR soon :D

mgattozzi:
Awesome I'll keep an eye out for it :)

PR #131, opened 2018-05-28
feat(remove quicli dependency)

### LINKED ISSUE #89: Replace quicli with only crates we need
"I want to use quicli to quickly build CLI apps – And I want to stop there! If you are no longer writing a small CLI tool (I’d say something like 500 lines of code is no longer small), your problems are likely more complicated than what quicli gives you tools for." from [quicli - The ideas behind my small Rust CLI framework](https://deterministic.space/quicli.html)

We're over 500 and I think it gave us a good starting point to get this built up but as we expand features and logging we will need to be more selective of imports and we don't need everything from the prelude, which there was a `not used` warning I could not get rid of unless I imported some crates anyways. I think we should take what crates we need from quicli then fix up the code to work without it.

I have this on a wip bit of code whereby in `command.rs` I removed the `prelude` import to get rid of the warning during compilation. I had to import structopt derive into the crate as well for that to work.

Thoughts @ashleygwilliams?

mgattozzi:
Remove quicli from the code base. It's a fantastic library to get
started, but in order to implement logging as well as just maintaining
the library in general it was easier to remove it and continue work
without it. While we were going to remove it in 0.4 we found it easier
to remove now to implement logging.
Closes #89 

```bash
$ rustup override set nightly
```


src/main.rs
```diff
@@ -1,16 +1,29 @@
...
+use wasm_pack::error::Error;
 use wasm_pack::Cli;
 
-main!(|args: Cli, log_level: verbosity| {
+fn main() {
     setup_panic!();
+    if let Err(e) = run() {
+        eprintln!("{}", e);
+        for cause in e.causes().skip(1) {
+            eprintln!("Caused by: {}", cause);
```
ashleygwilliams: can you give some examples of what this error ends up looking like?

mgattozzi: This doesn't change functionality this is what quicli already does

killercup:
In case you want to keep the env_logger setup, I present to you: https://crates.io/crates/clap-verbosity-flag

PR #149, opened 2018-06-05
Add WASM to Cargo categories

### LINKED ISSUE #148: add category wasm to cargo.toml
this issue is reserved for someone who has not yet contributed to wasm-pack

<hr/>

this is a wasm tool- so we should be discoverable as such. add a category to cargo.toml to declare this (for an example see https://github.com/rustwasm/wasm-bindgen/blob/master/Cargo.toml#L7)

TomasHubelbauer:
Closes #148. Snatching this one while it's hot.


ashleygwilliams:
thanks so much for this! will merge once CI finishes :)

ashleygwilliams:
🔥🔥🔥🔥🔥🔥<3

PR #150, opened 2018-06-05
check that crate type includes cdylib

kedromelon:
Checks that crate-type includes "cdylib" -- fails with an error if it does not.

Closes https://github.com/ashleygwilliams/wasm-pack/issues/140

thank you for holding the rust workshop at jsconf this past weekend @ashleygwilliams! 😄

### DIFF
--- a/src/error.rs
+++ b/src/error.rs
@@ -16,7 +16,7 @@ pub enum Error {
     #[fail(display = "{}. stderr:\n\n{}", message, stderr)]
     Cli { message: String, stderr: String },
     #[fail(display = "{}", message)]
-    Config { message: String },
+    CrateConfig { message: String },
 }
 
 impl Error {
@@ -27,8 +27,8 @@ impl Error {
         })
     }
 
-    pub fn config(message: &str) -> Result<(), Self> {
-        Err(Error::Config {
+    pub fn crate_config(message: &str) -> Result<(), Self> {
+        Err(Error::CrateConfig {
             message: message.to_string(),
         })
     }
@@ -42,7 +42,9 @@ impl Error {
                 message: _,
                 stderr: _,
             } => "There was an error while calling another CLI tool. Details:\n\n",
-            Error::Config { message: _ } => "There was a configuration error. Details:\n\n",
+            Error::CrateConfig { message: _ } => {
+                "There was a crate configuration error. Details:\n\n"
+            }
         }.to_string()
     }
 }
--- a/src/error.rs
+++ b/src/error.rs
@@ -15,6 +15,8 @@ pub enum Error {
     SerdeToml(#[cause] toml::de::Error),
     #[fail(display = "{}. stderr:\n\n{}", message, stderr)]
     Cli { message: String, stderr: String },
+    #[fail(display = "{}", message)]
+    Config { message: String },
 }
 
 impl Error {
@@ -22,6 +24,12 @@ impl Error {
         Err(Error::Cli {
             message: message.to_string(),
             stderr: stderr.to_string(),
+        })
+    }
+
+    pub fn config(message: &str) -> Result<(), Self> {
+        Err(Error::Config {
+            message: message.to_string(),
         })
     }
 
@@ -34,6 +42,7 @@ impl Error {
                 message: _,
                 stderr: _,
             } => "There was an error while calling another CLI tool. Details:\n\n",
+            Error::Config { message: _ } => "There was a configuration error. Details:\n\n",
         }.to_string()
     }
 }
--- a/src/manifest.rs
+++ b/src/manifest.rs
@@ -11,6 +11,7 @@ use PBAR;
 #[derive(Deserialize)]
 struct CargoManifest {
     package: CargoPackage,
+    lib: Option<CargoLib>,
 }
 
 #[derive(Deserialize)]
@@ -21,6 +22,11 @@ struct CargoPackage {
     version: String,
     license: Option<String>,
     repository: Option<String>,
+}
+
+#[derive(Deserialize)]
+struct CargoLib {
+    crate_type: Option<Vec<String>>,
 }
 
 #[derive(Serialize)]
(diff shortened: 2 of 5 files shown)


src/manifest.rs
```diff
@@ -49,7 +55,9 @@ fn read_cargo_toml(path: &str) -> Result<CargoManifest, Error> {
     let mut cargo_contents = String::new();
     cargo_file.read_to_string(&mut cargo_contents)?;
 
-    Ok(toml::from_str(&cargo_contents)?)
+    Ok(toml::from_str(
+        &cargo_contents.replace("crate-type", "crate_type"),
```
kedromelon: 😱 this makes me nervous. any ideas for a more robust way to handle this?

src/manifest.rs
```diff
@@ -136,3 +144,21 @@ pub fn write_package_json(
 pub fn get_crate_name(path: &str) -> Result<String, Error> {
     Ok(read_cargo_toml(path)?.package.name)
 }
+
+fn has_cdylib(path: &str) -> bool {
+    match read_cargo_toml(path).unwrap().lib {
+        Some(lib) => match lib.crate_type {
```
kedromelon: the nested pattern matching here feels unnecessarily verbose -- is there a more succinct way to express this?

src/manifest.rs
```diff
@@ -136,3 +144,21 @@ pub fn write_package_json(
...
+            Some(types) => types.iter().any(|s| s == "cdylib"),
+            _ => false,
+        },
+        _ => false,
+    }
+}
+
+pub fn check_crate_type(path: &str) -> Result<(), Error> {
+    if !has_cdylib(path) {
+        Error::config("crate-type must include cdylib to compile to wasm32-unknown-unknown")
```
kedromelon: the existing errors didn't seem appropriate for this use-case, so i added a new one

src/manifest.rs
mgattozzi: What's the reason you're chaning it from crate-type to crate_type here? It might not be possible if the reason why is valid. If you're deserializing from a file I think we can have serde deserialize the field better with an attribute elsewhere

src/manifest.rs
mgattozzi: You probably want to use [and_then](https://doc.rust-lang.org/std/option/enum.Option.html#method.and_then) in order to do this as well as `map_or`. If you haven't looked at Option or Result's docs definitely do so as you can find all kinds of neat methods to help!

The code will probaby end up looking like this if you want to make it more succinct

```rust
// If nothing goes wrong when it reads then we have the value wrapped up in an Ok.
// If something goes wrong the ? will return the error early
Ok(read_cargo_toml(path)?.lib
    // If the value is some we can then doe something with the value
    .and_then(|lib| {
         // Use the map_or function for Options to map the inner value to a boolean. If the
         // value for lib.crate_type is None then default to false. Return the boolean as the
         // Option is consumed with map_or
         lib.crate_type.map_or(|types| types.iter().any(|s| s == "cdylib), false)
     }))
```
If that's not clear tell me why and I can explain it a bit better! :D

src/manifest.rs
mgattozzi: That's totally fine! This function will also need to put a ? at the end of `has_cdylib` once you change it to a Result type as well

src/manifest.rs
mgattozzi: Oh I see now! You'll want to add this attribute to above the `crate_type` field for `CargoLib`

`#[serde(rename = "crate-type")]`

This tells serde that if it sees a field crate-type in the toml file it should be `crate_type` for Rust

src/manifest.rs
mgattozzi: Actually you'll want to use map_or instead of and_then, with a default value of false now that I'm looking at this again.

ashleygwilliams:
@kedromelon this is a super awesome start and thanks so much for doing this! i think @mgattozzi's comments are on point- but in general your initial instincts seem great! don't hesitate to reach out if you have any other questions!

kedromelon:
thanks for the feedback @mgattozzi and @ashleygwilliams!! very helpful 😊

will add these revisions asap!

src/manifest.rs
kedromelon: oooooh nice! very cool

mgattozzi:
Awesome! Looking forward to the revisions and hopefully we can get this merged soon :D

kedromelon:
@mgattozzi pushed up those changes! lmk if i missed anything. thanks again for the comments, really helpful for getting up to speed with Rust 😄

regarding testing a Cargo.toml that has cdylib, does the test i added using the [`js-hello-world` fixture](https://github.com/ashleygwilliams/wasm-pack/pull/150/files#diff-e2aaec4f8436cc0cad957485d4904136R35) work or were you referring to something else?

ashleygwilliams:
@kedromelon i think that test sounds great! i'll take a look at the other changes in a second :)

src/error.rs
```diff
@@ -34,6 +42,7 @@ impl Error {
                 message: _,
                 stderr: _,
             } => "There was an error while calling another CLI tool. Details:\n\n",
+            Error::Config { message: _ } => "There was a configuration error. Details:\n\n",
```
ashleygwilliams: we might add "crate" here, i.e. "there was a crate configuration error" - there's a lot of moving parts here so it might be useful to explain that the error was with the crate. we might want to call this CrateConfig as a result. open to ya'lls thoughts tho

src/manifest.rs
```diff
@@ -136,3 +143,18 @@ pub fn write_package_json(
...
+fn has_cdylib(path: &str) -> Result<bool, Error> {
+    Ok(read_cargo_toml(path)?.lib.map_or(false, |lib| {
+        lib.crate_type
+            .map_or(false, |types| types.iter().any(|s| s == "cdylib"))
+    }))
+}
+
+pub fn check_crate_type(path: &str) -> Result<(), Error> {
+    if !has_cdylib(path)? {
+        Error::config("crate-type must include cdylib to compile to wasm32-unknown-unknown")
```
ashleygwilliams: it would be awesome if we can give the exact code a user would need to put into their cargo.toml to fix the issue

src/error.rs
kedromelon: that makes sense to me, i had been considering something similar at first as well!

kedromelon:
pushed those in https://github.com/ashleygwilliams/wasm-pack/pull/150/commits/2cb292164cc7adc688dd28db0086ff53d4f39995!

any further thoughts on the wording / formatting of the example code in the error message?

kedromelon:
great!! thanks @ashleygwilliams and @mgattozzi!

mgattozzi:
No, thank you @kedromelon! :D

Issue #144, opened 2018-06-05
--skip-build flag


ashleygwilliams:
while teaching wasm-pack, i discovered that sometimes we want to run wasm-pack init again (updated readme or cargo.toml) but don't want to rebuild the whole pkg

let's add a `--skip-build` flag that will skip:

- adding target
- compiling to wasm
- installing wasm-bindgen
- running wasm-bindgen

it should warn if the wasm file does not exist in target (was not yet compiled) or if the bg.wasm or js module wrapper are not present (hasnt yet run wasm-bingden)

mgattozzi:
One important thing to note is that the order of what happens has a specific number of steps hardcoded in the output. We'll need to add checks to choose the right amount if this flag is added. i.e. instead of say step [3/7] with this flag we might just need to do [2/3]

ashleygwilliams:
@mgattozzi good call. i mean, this is probably a good reason to stop hardcoding that stuff in the first place. the more steps and flags we add the more likely we'll want to have a dynamic way to set up a list of steps we want and then count them from there.

clanehin:
First apologies. Just noticed this issue. I'm aware there's already a pull request. This flag simultaneously does something I really want and also qualities that make it less useful.

Parts of this flag that I really like: "Adding the target" and "installing wasm-bindgen" affect the user's global environment. That always has a lot of potential to cause surprise or breakage. Especially if I give anyone (coworker or stranger on github) a build script that uses wasm-pack as an intermediate step I have to be very careful to educate them about what is going to happen (and some people would interpret an indirectly-commanded software install as malware, eep!). Also these steps are slow. So suppressing these is a big win except when they're wanted.

Parts of this flag that feel regrettable: Having disabled the above, I still want "Compiling to wasm" and "running wasm-bindgen". And these are relatively fast operations. In --debug mode, we get incremental compilation, and even in release mode there seems to at least be some kind of short-circuit behavior. So, AFAICS, if someone is saving time by skipping these steps, then that itself is evidence that their output is stale and probably wrong and not what they intend.

I put some times below; emphasis on the very last result, which is _sub-second despite apparently rebuilding a 1.4M wasm file_. I actually re-run wasm-pack in some cases 10-20 times in an hour if I'm fine-tuning something, so I'm intensely interested in these times.

I hope this is helpful.

(Note to avoid confusing anyone who comes across this ticket: as of this writing, the --debug flag lives in a pull request and is not yet a part of wasm-pack.)

```
[...]

PR #151, opened 2018-06-05
Add --skip-build flag for init command #144

kohensu:
pull request for #144

### DIFF
--- a/src/command.rs
+++ b/src/command.rs
@@ -1,11 +1,11 @@
 use bindgen;
 use build;
-use console::style;
 use emoji;
 use error::Error;
 use indicatif::HumanDuration;
 use manifest;
 use npm;
+use progressbar::Step;
 #[allow(unused)]
 use readme;
 use slog::Logger;
@@ -23,6 +23,10 @@ pub enum Command {
 
         #[structopt(long = "scope", short = "s")]
         scope: Option<String>,
+
+        #[structopt(long = "--skip-build")]
+        /// Do not build, only update metadata
+        skip_build: bool,
 
         #[structopt(long = "no-typescript")]
         /// By default a *.d.ts file is generated for the generated JS file, but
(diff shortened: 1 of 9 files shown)


src/command.rs
```diff
@@ -145,111 +151,202 @@ pub fn run_wasm_pack(command: Command, log: &Logger) -> result::Result<(), Error
...
+    ) -> Init {
+        Init {
+            crate_path: set_crate_path(path),
+            scope,
+            skip_build,
+            disable_dts,
+            target,
+            crate_name: None,
+        }
+    }
```
ashleygwilliams: would like a line here :)

src/command.rs
```diff
@@ -145,111 +151,202 @@ pub fn run_wasm_pack(command: Command, log: &Logger) -> result::Result<(), Error
...
+    pub fn process(&mut self, log: &Logger) -> result::Result<(), Error> {
+        let steps: Vec<fn(&mut Init, &Step, &Logger) -> result::Result<(), Error>> =
+            if self.skip_build {
```
ashleygwilliams: (ugh GH didn't post this comment in my original review)

so- i think we might be exploring even more "build" modes (see #153 ) - so (and i am likely overengineering here) but perhaps an enum and a match statement would work best here. would love to hear your thoughts!

src/command.rs
kohensu: oups

src/command.rs
kohensu: I did the modification this modification in the PR, if it is not like you what you had in mind I will revert it.

ashleygwilliams:
also- just as a heads up we are probably landing 2 PRs today which means this will likely need a rebase, but i'd like to prioritize getting this in after that so that we don't have too much rebasing that needs to happen :)

kohensu:
ok, I will do the rebase, the test and the doc

PR #156, opened 2018-06-06
Bumping human-panic to 1.0.0

spacekookie:
None of the boxes were applicable, this is really a minor change.

Since `human-panic` is now stable, you can just use the stable version to signal that the API won't just change :smile:


mgattozzi:
Gonna wait for the CI to pass but yeah this looks like a very simple change to get in thanks for the update @spacekookie! :D

ashleygwilliams:
thank you so much! so excited that ya'll are 1.0 now 🦄 🎉 ✨ ❤️

PR #164, opened 2018-06-11
Fix spinners

### LINKED ISSUE #145: spinner no longer working
the spinner used to spin- it seems to not now. this isn't a huge problem, but it'd be nice to fix by the next release!

migerh:
This is an attempt to fix #145.

Since I'm new to rust there are a few things I'm not sure about, e.g. the use of RwLock to not have to deal with mutable ProgressOutput instance in the rest of the codebase and how to properly cache the infos/warnings/errors until the spinner is finished. If there are better ways - some pointers in the right direction would be appreciated.

This is not my first attempt to fix the spinners, but it is the first one that does not cause more problems than it solves:

* Using multibar does not work properly while running a child process.
* Instantiating a new spinner and not add it to a multibar breaks as soon an error/warning/info is emitted. The last of these immediate outputs will get absorbed by the spinner as well (see example in #145).
* Keeping a single spinner the whole time and using the indicatif prefix feature to keep the previous spinners output as well as infos/warnings/errors did not work out so well either: The prefix would be output again and again on every tick of a spinner, spamming the terminal.

- [x] You have the latest version of `rustfmt` installed and have your cloned directory set to nightly

Thanks for creating/maintaining wasm-pack :)

### DIFF
--- a/src/bindgen.rs
+++ b/src/bindgen.rs
@@ -10,13 +10,12 @@ pub fn cargo_install_wasm_bindgen() -> Result<(), Error> {
         style("[6/7]").bold().dim(),
         emoji::DOWN_ARROW
     );
-    let pb = PBAR.message(&step);
+    PBAR.message(&step);
     let output = Command::new("cargo")
         .arg("install")
         .arg("wasm-bindgen-cli")
         .arg("--force")
         .output()?;
-    pb.finish();
     if !output.status.success() {
         let s = String::from_utf8_lossy(&output.stderr);
         if s.contains("already exists") {
@@ -40,7 +39,7 @@ pub fn wasm_bindgen_build(
         style("[7/7]").bold().dim(),
         emoji::RUNNER
     );
-    let pb = PBAR.message(&step);
+    PBAR.message(&step);
     let binary_name = name.replace("-", "_");
     let wasm_path = format!("target/wasm32-unknown-unknown/release/{}.wasm", binary_name);
 
@@ -63,7 +62,6 @@ pub fn wasm_bindgen_build(
         .arg(dts_arg)
         .arg(target_arg)
         .output()?;
-    pb.finish();
     if !output.status.success() {
         let s = String::from_utf8_lossy(&output.stderr);
         Error::cli("wasm-bindgen failed to execute properly", s)
--- a/src/build.rs
+++ b/src/build.rs
@@ -10,13 +10,12 @@ pub fn rustup_add_wasm_target() -> Result<(), Error> {
         style("[1/7]").bold().dim(),
         emoji::TARGET
     );
-    let pb = PBAR.message(&step);
+    PBAR.message(&step);
     let output = Command::new("rustup")
         .arg("target")
         .arg("add")
         .arg("wasm32-unknown-unknown")
         .output()?;
-    pb.finish();
     if !output.status.success() {
         let s = String::from_utf8_lossy(&output.stderr);
         Error::cli("Adding the wasm32-unknown-unknown target failed", s)
@@ -31,7 +30,7 @@ pub fn cargo_build_wasm(path: &str) -> Result<(), Error> {
         style("[2/7]").bold().dim(),
         emoji::CYCLONE
     );
-    let pb = PBAR.message(&step);
+    PBAR.message(&step);
     let output = Command::new("cargo")
         .current_dir(path)
         .arg("build")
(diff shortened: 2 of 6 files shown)


mgattozzi:
If it returns an error when it drops on exit of program then I think we have bigger problems. This is probably the one place you should `unwrap()` because there's really no way to handle it. Best to use `expect("Progress bar failed to drop properly")` so that we know where it failed. Human Panic will take care of making a nice error that people can use in the bug report

As for your second point those have to work with how `failure_derive` expands out the code for an error. I'll leave comments for you on how to get that updated to be real nice and it'll be the last thing to do. It'll help make the errors more detailed if something does fail :)

Otherwise this looks fantastic. Great work!

src/error.rs
```diff
@@ -13,6 +14,8 @@ pub enum Error {
     SerdeJson(#[cause] serde_json::Error),
     #[fail(display = "{}", _0)]
     SerdeToml(#[cause] toml::de::Error),
+    #[fail(display = "Acquiring lock failed")]
+    PoisonedLockError,
```
mgattozzi: Okay so for this make it `PoisonedLockError(#[cause] PoisonError)`
It's a wrapper for the inner error and the `#[cause]` tells failure how to get it

src/error.rs
```diff
@@ -13,6 +14,8 @@ pub enum Error {
     SerdeJson(#[cause] serde_json::Error),
     #[fail(display = "{}", _0)]
     SerdeToml(#[cause] toml::de::Error),
+    #[fail(display = "Acquiring lock failed")]
```
mgattozzi: Then just change this to:
`#[fail(display = "{}", _0)]`

Which basically says if this error happens for it's `Display` implementation grab the first value (the `_0`) and use it's display implementation for this string. Since `PoisonError` implements `std::error::Error` we know it has a display implementation which we want to use here. same reason for the `#[cause]` part

src/error.rs
```diff
@@ -38,6 +41,7 @@ impl Error {
             Error::Io(_) => "There was an I/O error. Details:\n\n",
             Error::SerdeJson(_) => "There was an JSON error. Details:\n\n",
             Error::SerdeToml(_) => "There was an TOML error. Details:\n\n",
+            Error::PoisonedLockError => "There was an RwLock error. Details: \n\n",
```
mgattozzi: This will have to then be `Error::PoisonedLockError(_)`

src/error.rs
```diff
@@ -66,3 +70,9 @@ impl From<toml::de::Error> for Error {
         Error::SerdeToml(e)
     }
 }
+
+impl<T> From<PoisonError<T>> for Error {
+    fn from(_: PoisonError<T>) -> Self {
```
mgattozzi: then change this to `fn from(e: PoisonError<T>) -> Self {`

src/error.rs
```diff
@@ -66,3 +70,9 @@ impl From<toml::de::Error> for Error {
...
+        Error::PoisonedLockError
```
mgattozzi: And lastly `Error::PoisonedLockError(e)`

ashleygwilliams:
assuming we can resolve some of these nits, i'd love to get this into a release tomorrow! thanks so much @mgattozzi and @migerh :D

mgattozzi:
It shouldn't take long at all! Just one more commit :D

src/error.rs
migerh: I tried it that way first but this is where I got stuck because `PoisonError` is generic. So if I write exactly that the type parameter is missing. Not sure how to fix that.

migerh:
> It shouldn't take long at all! Just one more commit :D

Yes, if it weren't for that type parameter in PoisonError ;) If it's urgent, feel free to patch it up yourselves, I won't be able to do anything before 6pm CEST (noon ET).

mgattozzi:
Oof @migerh you are absolutely right. Hmmmm... Might patch it up but it's probably okay for now as is then and we can make it nicer later. Since we're not panicking except on drop then having a poisoned lock isn't really going to happen.

migerh:
> Since we're not panicking except on drop then having a poisoned lock

Nope, we don't :) That I managed to put in this morning in 191262c. Only the improved error conversion is missing.

ashleygwilliams:
ok i am gonna attempt the needed rebase on this to get it in- @mgattozzi or @migerh could one of you file an issue about the refactoring we'd like to see for this later? thanks so much for all your work!!

migerh:
The changes in this PR rebased against the current master (commit 460dade1dfe1e94b2) in a single commit: https://github.com/migerh/wasm-pack/commit/9734e472fdfd1d994eb955e2951dc643ec360a29

migerh:
> @mgattozzi or @migerh could one of you file an issue about the refactoring we'd like to see for this later?

Done: #179

Issue #171, opened 2018-06-12
Add nightly toolchain to commands


FreeMasen:
Currently, unless the user has set cargo to use the nightly compiler, this project will fail in a somewhat unfriendly way.

Updating the rustup/cargo commands to account for this would be helpful.

`rustup target add wasm32-unknown-unknown --toolchain nightly`
`cargo +nightly build --target wasm32-unknown-unknown`

It may be helpful to check that the nightly toolchain is actually installed as well
`rustc +nightly -V` or similar.

ashleygwilliams:
@FreeMasen this is an excellent suggestion! if you'd like to add it- we'd welcome a PR- otherwise i'll open this up to contributors!

FreeMasen:
I should be able to knock that out tomorrow.

PR #172, opened 2018-06-13
Add nightly toolchain to commands

FreeMasen:
To address issue #171 

This change will check if `rustc +nightly -V` returns an error, if it does it will run `rustup toolchain add nightly`. It then updated the previous process to run the following.

`rustup target add wasm32-unknown-unknown --toolchain nightly`
`cargo +nightly build --target wasm32-unknown-unknown`

This should make things a little bit easier for new rust users

- ✅ You have the latest version of `rustfmt` installed and have your 
```bash
$ rustup override set nightly
```
- ✅ You ran `rustfmt` on the code base before submitting
- ✅ You reference which issue is being closed in the PR text


src/build.rs
```diff
@@ -11,10 +11,30 @@ pub fn rustup_add_wasm_target() -> Result<(), Error> {
         emoji::TARGET
     );
     let pb = PBAR.message(&step);
+    let nightly_check = Command::new("rustc").arg("+nightly").arg("-V").output()?;
```
mgattozzi: This whole part should be it's own section/fn as there's no sub bar really. It should just be part of the process.

src/build.rs
```diff
@@ -11,10 +11,30 @@ pub fn rustup_add_wasm_target() -> Result<(), Error> {
...
+    if !nightly_check.status.success() {
+        let sub_bar = PBAR.message(&format!(
```
mgattozzi: There's another PR where this is going to change a little so be prepared to rebase!

Specifically here:
https://github.com/ashleygwilliams/wasm-pack/pull/151

You'll need to add this function to the vec that way it knows to call it and how to calculate how many steps there are etc.

src/build.rs
FreeMasen: Just pushed an update with this change

src/build.rs
FreeMasen: Let me know when I need to perform the rebase

ashleygwilliams:
hey! thanks so much for this- i think this is great, and mostly echo @mgattozzi 's comments re: rebasing, particularly from #151 which i think we intend to land first- if you could rebase this off of that i think it'd be great to try and get this into today's release.

if you're not up for that- don't worry, we can also do it for you- but i always like to ask if folks would like to do it themselves :)

ashleygwilliams:
oh also- would you mind fixing up into a single commit message? thanks so much!

FreeMasen:
This should be 1 commit and it was re-based against [the 151 branch](https://github.com/kohensu/wasm-pack/tree/add_--skip-build_flag_%23144)

ashleygwilliams:
thank you so much!!

PR #180, opened 2018-06-14
Let the spinners spin again

### LINKED ISSUE #145: spinner no longer working

ashleygwilliams:
fixes #145 

This will fix the broken spinners that indicate steps which are
currently in progress. To achieve this the behavior of the progressbar
helper had to be adjusted.

The multibar does not work while a child process is run with
`std::process::Command`. Using multiple spinners without a multibar is
problematic as well because emitting warnings, infos and errors while a
single spinner is active will duplicate the spinners message. The
spinner will also absorb the last warning/error/info. Instead of
publishing warnings, errors and infos immediately they will now be
cached and only output when the current spinner finishes.

To make sure we can output all warnings to the user ProgressBars are no
longer exposed to the caller. Instead the active spinner will be
finished implicitly when a new spinner is allocated with the `message()`
function. Instead of relying on the `done()` function to be called the
progressbar now implements the `Drop` trait which will finish the last
spinner automatically.

Unfortunately, this introduces members that have to be mutable. To not
have to deal with mutable ProgressOutput all over the place the members
were put inside an RwLock. This allows us to use ProgressOutput from
inside multiple threads to emit warnings and errors.

- [ ] You have the latest version of `rustfmt` installed and have your 
```bash
$ rustup override set nightly
```
- [ ] You ran `rustfmt` on the code base before submitting

### DIFF
--- a/src/bindgen.rs
+++ b/src/bindgen.rs
@@ -6,17 +6,16 @@ use PBAR;
 
 pub fn cargo_install_wasm_bindgen(step: &Step) -> Result<(), Error> {
     let msg = format!("{}Installing WASM-bindgen...", emoji::DOWN_ARROW);
-    let pb = PBAR.step(step, &msg);
+    PBAR.step(step, &msg)?;
     let output = Command::new("cargo")
         .arg("install")
         .arg("wasm-bindgen-cli")
         .arg("--force")
         .output()?;
-    pb.finish();
     if !output.status.success() {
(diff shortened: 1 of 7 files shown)


mgattozzi:
Since this is what we had already review before this LGTM

ashleygwilliams:
closing in favor of #181

PR #81, opened 2018-04-13
Don't attempt wasm-bindgen install if it exists in path.

data-pup:
(Fixing issue #51.)

After running into some of the problems in the previous PR [here], I squashed the commits from my previous PR down, and rebased it off of the recent changes to the repo.

[here]: https://github.com/ashleygwilliams/wasm-pack/pull/79

In case there are issues with the Travis build again, here are the rust-fmt and rustc versions I am using.

```
✨  cargo fmt --version
rustfmt 0.4.1-nightly (e784712 2018-04-09)

✨  rustc --version
rustc 1.27.0-nightly (ad610bed8 2018-04-11)
```

Thank you for bearing with me on the linting problems by the way, pardon me.

PR #192, opened 2018-06-19
Coordinating wasm-bindgen versions

### LINKED ISSUE #146: coordinating wasm-bindgen versions
currently we are force installing wasm-bindgen, but this is frustrating because it takes a long time and often doesn't need to happen. i'd love a solution for this. let's use this issue to brainstorm. (related to #51  )

strawman proposal:

- read cargo.toml to find wasm-bindgen lib version
- check currently install wasm-bindgen version
   - if not installed, install version listed in cargo.toml
   - if installed, check version, if version mismatch install otherwise skip

@alexcrichton can you confirm that the lib and cli versions are paired up and that this strategy would be effective?

data-pup:
Fixes #146. This aims to solve the issue of making sure that there is (a) a locally installed `wasm-bindgen` and (b) that it is the same version as specified in the `Cargo.toml` file.

As of now, there are still some rough edges that I want to hone down here. It seemed like a good idea to check in the existing work on this issue that I got done today, to make sure that I am headed in the right direction on this. Any feedback is totally welcome!

I'll also close #81 now, since I believe this will also take care of #51.

I should be able to get this all wrapped up tomorrow, had to spend a little more time than expected today getting up to speed with the changes that have occurred since then. Loving all the stuff people have contributed! :smile: 

---

```bash
$ rustup override set nightly
```

### DIFF
--- a/src/bindgen.rs
+++ b/src/bindgen.rs
@@ -3,57 +3,8 @@
 use emoji;
 use error::Error;
 use progressbar::Step;
-use std::{env, fs, path, process::Command};
+use std::{env, path, process::Command};
 use PBAR;
-
-/// Return a string containing the path to the local `wasm-bindgen`.
-fn local_wasm_bindgen_path_str(crate_path: &str) -> String {
-    #[cfg(not(target_family = "windows"))]
-    return format!("{}/{}", crate_path, "bin/wasm-bindgen");
-    #[cfg(target_family = "windows")]
-    return format!("{}\\{}", crate_path, "bin\\wasm-bindgen");
-}
-
-/// Return a string containing the path to the global `wasm-bindgen`.
-fn global_wasm_bindgen_path_str() -> Result<Option<String>, Error> {
-    #[cfg(target_family = "windows")]
-    let path_sep: &str = ";";
-    #[cfg(not(target_family = "windows"))]
-    let path_sep: &str = ":";
-
-    let path = env::var("PATH")?;
-    for path_dir in path.split(path_sep)
-    {
-        let prog_str = format!("{}/wasm-bindgen", path_dir);
-        if fs::metadata(&prog_str).is_ok() {
-            return Ok(Some(prog_str))
-        }
-    }
-
-    Ok(None)
-}
-
-/// Check if the `wasm-bindgen` dependency is locally satisfied.
-fn wasm_bindgen_version_check(crate_path: &str, dep_version: &str) -> Result<bool, Error> {
-    let path_str = local_wasm_bindgen_path_str(crate_path);
-    let wasm_bindgen = path::Path::new(&path_str);
-
-    if !wasm_bindgen.is_file() {
-        return Ok(false);
-    }
-
-    let output = Command::new(wasm_bindgen).arg("--version").output()?;
-    if output.status.success() {
-        let s = String::from_utf8_lossy(&output.stdout);
-        let installed_version = s.trim();
-        Ok(installed_version == dep_version)
-    } else {
-        let error_msg = "Could not find version of local wasm-bindgen";
-        let s = String::from_utf8_lossy(&output.stderr);
-        let e = Error::cli(error_msg, s);
-        Err(e)
-    }
-}
 
 /// Install the `wasm-bindgen` CLI with `cargo install`.
 pub fn cargo_install_wasm_bindgen(
@@ -62,14 +13,26 @@ pub fn cargo_install_wasm_bindgen(
     install_permitted: bool,
     step: &Step,
 ) -> Result<(), Error> {
-    if wasm_bindgen_version_check(path, version)? {
+    // Determine if the `wasm-bindgen` dependency is already met for the given mode.
+    let dependency_met = if let Some(bindgen_path) = wasm_bindgen_path_str(path, install_permitted)?
+    {
+        wasm_bindgen_version_check(&bindgen_path, version)?
+    } else {
+        false
+    };
+
+    // If the `wasm-bindgen` dependency is already met, print a message and return.
+    if dependency_met {
         let msg = format!("{}WASM-bindgen already installed...", emoji::DOWN_ARROW);
         PBAR.step(step, &msg);
         return Ok(());
     }
 
+    // If the `wasm-bindgen` dependency was not met, and installs are not
+    // permitted, return a configuration error.
     if !install_permitted {
         return Err(Error::crate_config("WASM-bindgen is not installed!"));
+        // FIXUP: This error message can be improved.
     }
 
     let msg = format!("{}Installing WASM-bindgen...", emoji::DOWN_ARROW);
@@ -98,6 +61,7 @@ pub fn wasm_bindgen_build(
     name: &str,
     disable_dts: bool,
     target: &str,
+    install_permitted: bool,
     debug: bool,
     step: &Step,
 ) -> Result<(), Error> {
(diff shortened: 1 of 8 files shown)


src/bindgen.rs
```diff
@@ -1,16 +1,40 @@
...
+    if output.status.success() {
+        let stdout = String::from_utf8_lossy(&output.stdout);
+        let installed_version = stdout.trim();
+        Ok(dep_version == installed_version)
+    } else {
+        let message = "Could not find version of local wasm-bindgen".to_string();
+        let stderr = String::from_utf8_lossy(&output.stderr).to_string();
+        let e = Error::Cli { message, stderr };
+        Err(e)
+    }
```
data-pup: Not a giant fan of this :point_up: Creating the `Error:Cli` manually, rather than through the `cli(..)` function seems a little hacky, but the other approach caused a return type error, since this function returns a `Result<bool, Error>

src/bindgen.rs
```diff
@@ -1,16 +1,40 @@
...
-use std::process::Command;
+use std::{fs, process::Command};
 use PBAR;
 
-pub fn cargo_install_wasm_bindgen(step: &Step) -> Result<(), Error> {
+static LOCAL_BINDGEN_PATH: &str = "bin/wasm-bindgen";
+
+pub fn wasm_bindgen_locally_installed(_path: &str) -> bool {
+    fs::metadata(LOCAL_BINDGEN_PATH).is_ok()
+}
```
data-pup: Should I create an absolute path here? Or is the relative path in `LOCAL_BINDGEN_PATH` enough? I think I saw an issue regarding what happens when `wasm-pack` is run from within the `pkg` directory, so I was thinking that using the crate path could make this a little less fragile.

If relative path is satisfactory, I can remove the parameter here, otherwise an absolute path can be formed with `format!` :)

src/command/init.rs
```diff
@@ -188,9 +188,25 @@ impl Init {
...
+                    install_needed = false;
+                }
+            }
+        }
+
+        if install_needed {
+            info!(&log, "Installing wasm-bindgen-cli...");
+            bindgen::cargo_install_wasm_bindgen(&self.crate_path, &bindgen_version, step)?;
+            info!(&log, "Installing wasm-bindgen-cli was successful.");
+        }
```
data-pup: The control flow here is the main thing I'd like to clean up, not especially happy with these nested `if` statements.

data-pup:
I think this is ready for review now :smile: Let me know if you have any feedback!

ashleygwilliams:
@data-pup awesome! i'll take a look soon- in the meantime it looks like rust fmt needs appeasing lolsob. thanks for putting so much effort into this! it's super appreciated. as mentioned previously- i think i'm down to sneak this in in a point release because it's gonna make everything way more awesome :)

data-pup:
Ahh pardon! I ran `fmt` before I rebased this, that's fixed now :) Happy I could help!

src/bindgen.rs
```diff
@@ -1,16 +1,46 @@
 use emoji;
 use error::Error;
 use progressbar::Step;
-use std::process::Command;
+use std::{path, process::Command};
 use PBAR;
 
-pub fn cargo_install_wasm_bindgen(step: &Step) -> Result<(), Error> {
+static LOCAL_BINDGEN_PATH: &str = "bin/wasm-bindgen";
```
mgattozzi: You should cfg this so that windows is "bin\\wasm-bindgen". While windows can handle it (sort of? windows pathing is wild technically the yen symbol is a valid path separator) it's best to have an option for it.

src/bindgen.rs
```diff
@@ -1,16 +1,46 @@
...
+
+fn wasm_bindgen_exists_locally(crate_path: &str) -> bool {
+    let path_str = format!("{}/{}", crate_path, LOCAL_BINDGEN_PATH);
```
mgattozzi: Same reasoning here for the path.

src/bindgen.rs
mgattozzi: Odd. What was the error?

src/command/init.rs
```diff
@@ -192,9 +192,25 @@ impl Init {
     }
 
     fn step_install_wasm_bindgen(&mut self, step: &Step, log: &Logger) -> Result<(), Error> {
-        info!(&log, "Installing wasm-bindgen-cli...");
-        bindgen::cargo_install_wasm_bindgen(step)?;
-        info!(&log, "Installing wasm-bindgen-cli was successful.");
+        let bindgen_version =
+            manifest::get_wasm_bindgen_version(&self.crate_path)?.ok_or(Error::CrateConfig {
```
mgattozzi: We should just make a `crate_config` function for `Error` and use that here.

src/command/init.rs
```diff
@@ -192,9 +192,25 @@ impl Init {
...
+                message: "Unexpected error while parsing wasm-bindgen dependency version"
+                    .to_string(),
+            })?;
+
+        if bindgen_version.is_empty() {
+            let e = Error::CrateConfig {
```
mgattozzi: And here!

mgattozzi: then you can do `return Err(Error::crate_config("wasm-bindgen version dependency was empty"))`

src/manifest.rs
```diff
@@ -148,6 +148,13 @@ pub fn get_crate_name(path: &str) -> Result<String, Error> {
     Ok(read_cargo_toml(path)?.package.name)
 }
 
+pub fn get_wasm_bindgen_version(path: &str) -> Result<Option<String>, Error> {
+    let version: Option<String> = read_cargo_toml(path)?
```
mgattozzi: you shouldn't need to specify `Option<String>` here. The type is inferred as the return type

src/bindgen.rs
data-pup: If I use the `Error::cli` function here I end up with the following error:

```
expected type `std::result::Result<bool, _>`
found type `std::result::Result<(), _>`
```

This is happening because that function wraps the error up into a result, which is fine for functions that also return a `Result<(), Error>`, but means that it won't work for functions that return something else in the `Ok` case :sad`

```rust
pub fn cli(message: &str, stderr: Cow<str>) -> Result<(), Self> {
```

data-pup:
Added some more commits to fix the issues mentioned in @mgattozzi 's review. Thanks for the feedback! I added some platform specific path strings, and changed the `get_wasm_bindgen_version` function to return a `Result<String, Error>`. 

> then you can do return Err(Error::crate_config("wasm-bindgen version dependency was empty"))

Regarding this point, I realized that there was already a `crate_config` error, but did run into the same issue I'd seen with the other `cli` helper function. These were returning a `Result<(), Error>`, which meant that they aren't usable in a function that _does_ return a value in the success case.

I altered these so that they return `Self`, so that we can use these methods elsewhere too. This felt a little opinionated on my part, but I don't think it affects readability.

data-pup:
> You'll also need to make sure the part that calls wasm-bindgen in the build step uses the local one!

Still need to take care of this part, I'll do that right now though! :smiley:

data-pup:
Ready for another round of review, whenever you have time!

src/bindgen.rs
```diff
@@ -1,24 +1,67 @@
...
+    #[cfg(target_family = "windows")]
+    return format!("{}\\{}", crate_path, LOCAL_BINDGEN_PATH);
+}
+
+pub fn wasm_bindgen_version_check(
+    crate_path: &str,
+    dep_version: &str,
+    step: &Step,
+) -> Result<bool, Error> {
+    let msg = format!("{}Checking WASM-bindgen dependency...", emoji::CHECK);
```
ashleygwilliams: not sure if this should be a step displayed to users- i think we should just do it in the install wasm-bindgen step

src/bindgen.rs
```diff
@@ -1,24 +1,67 @@
 use emoji;
 use error::Error;
 use progressbar::Step;
-use std::process::Command;
+use std::{path, process::Command};
 use PBAR;
 
-pub fn cargo_install_wasm_bindgen(step: &Step) -> Result<(), Error> {
+#[cfg(not(target_family = "windows"))]
+static LOCAL_BINDGEN_PATH: &str = "bin/wasm-bindgen";
```
ashleygwilliams: this is local to the crate- correct? if so we'll want to add it to the .gitignore

src/bindgen.rs
```diff
@@ -1,24 +1,67 @@
...
+        return Ok(false);
+    }
+
+    let output = Command::new(wasm_bindgen).arg("--version").output()?;
+    if output.status.success() {
+        let s = String::from_utf8_lossy(&output.stdout);
+        let installed_version = s.trim();
+        Ok(installed_version == dep_version)
+    } else {
+        let error_msg = "Could not find version of local wasm-bindgen";
```
ashleygwilliams: we may want to give some sort of instruction about how to handle this if it happens. feels a little like a dead end right now :)

data-pup:
I like your points regarding the `--no-installs` flag, that's an important detail I don't think my changes currently account for. Would it make sense in that case to search through the locations in the $PATH environment variable for wasm-bingen?

Overriding the local behavior with a flag makes sense to me, but right now I'm not sure if I have a solid answer to the story about how this works either. I will think about this too, but feel free to follow up if you come to a decision :)

ashleygwilliams:
@data-pup i also don't know exactly what we should do! very open to suggestions. i can take some time this weekend to think on it :)

mgattozzi:
Doesn't no installs imply using a local one anyways? From my understanding it won't be installing a global version. In fact wouldn't it make more sense to just use a sandboxed version anyways rather than relying on a global one? In tools like stack/cabal for Haskell it uses sandboxed versions of everything unless specified to use a global one which seems like the better choice, unless I'm misunderstanding what you mean by --no-installs @ashleygwilliams

data-pup:
Just wanted to follow up on this, do you happen to have any advice for next steps on my end? If there might still be some thought required regarding the interface etc., I'm happy to hold off on this!

I know I can definitely add some extra logic to handle the `--no-installs` flag, but a little unsure what that should generally look like at the moment.

Btw, thanks for being nice maintainers! Appreciate the feedback so far :)

data-pup:
I saw that #146 has been marked as Blocking Rust 2018, so I rebased this off of master to prepare for further work on this. At this point I think all of the changes requested by Michael have been taken care of, and I can take care of some of the minor details mentioned in Ashley's review today.

My current plan is to look for a globally installed version of `wasm-bindgen` that matches the dependency version if `--no-installs` is provided. If that cannot be found in the path, then it will fail rather than installing a new version.

If I should take a different route, let me know!

data-pup:
Took a swing at this today :) Let me know if you spot any problems!

mgattozzi:
@data-pup I'll have to take a look now that I'm back when I get the chance

data-pup:
Sounds good to me! I can also rebase this again, now that the path things from #220 have landed :+1:

src/bindgen.rs
```diff
@@ -34,40 +61,112 @@ pub fn wasm_bindgen_build(
...
+
+    global_wasm_bindgen_path_str()
+}
+
+/// Return a string containing the path to the local `wasm-bindgen`, if it exists.
+fn local_wasm_bindgen_path_str(crate_path: &str) -> Option<String> {
+    #[cfg(not(target_family = "windows"))]
+    let local_path = format!("{}/{}", crate_path, "bin/wasm-bindgen");
+    #[cfg(target_family = "windows")]
+    let local_path = format!("{}\\{}", crate_path, "bin\\wasm-bindgen");
```
fitzgen: If these functions were using `std::path::Path[Buf]`, then we wouldn't need these different code paths. Is there a reason we aren't using them?

src/bindgen.rs
```diff
@@ -3,25 +3,52 @@
...
+pub fn cargo_install_wasm_bindgen(
+    path: &str,
+    version: &str,
+    install_permitted: bool,
+    step: &Step,
+) -> Result<(), Error> {
+    // Determine if the `wasm-bindgen` dependency is already met for the given mode.
+    let dependency_met = if let Some(bindgen_path) = wasm_bindgen_path_str(path, install_permitted)?
+    {
+        wasm_bindgen_version_check(&bindgen_path, version)?
```
fitzgen: I *think* we don't want to propagate errors here, but instead treat them as "false, dependency not met".

src/bindgen.rs
```diff
@@ -34,40 +61,112 @@ pub fn wasm_bindgen_build(
...
+        Err(e)
+    }
+}
+
+/// Return a string to `wasm-bindgen`, if it exists for the given mode.
+fn wasm_bindgen_path_str(
+    crate_path: &str,
+    install_permitted: bool,
+) -> Result<Option<String>, Error> {
+    if install_permitted {
```
fitzgen: If we are not permitting installs, we should still allow *using* a local install, shouldn't we? Just not installing the local version? Or is it more of "force use of the global version and don't install it if missing"?

src/bindgen.rs
```diff
@@ -34,40 +61,112 @@ pub fn wasm_bindgen_build(
...
+fn global_wasm_bindgen_path_str() -> Result<Option<String>, Error> {
+    #[cfg(target_family = "windows")]
+    let path_sep: &str = ";";
+    #[cfg(not(target_family = "windows"))]
+    let path_sep: &str = ":";
+
+    let path = env::var("PATH")?;
+    for path_dir in path.split(path_sep) {
+        let prog_str = format!("{}/wasm-bindgen", path_dir);
+        if path::Path::new(&prog_str).is_file() {
```
fitzgen: I think we should just use https://crates.io/crates/which instead of rolling our own implementation.

src/bindgen.rs
```diff
@@ -34,40 +61,112 @@ pub fn wasm_bindgen_build(
...
-        .arg("./pkg")
-        .arg(dts_arg)
-        .arg(target_arg)
-        .output()?;
-    if !output.status.success() {
+    let output = Command::new(wasm_bindgen).arg("--version").output()?;
+    if output.status.success() {
+        let s = String::from_utf8_lossy(&output.stdout);
+        let installed_version = s.trim();
+        Ok(installed_version == dep_version)
```
fitzgen: Running `wasm-bindgen --version` doesn't produce *just* the version number:

```
$ wasm-bindgen --version
wasm-bindgen 0.2.15
```

So I think we need to do a little bit more parsing here than just `.trim()`, since we are getting the expected version from `Cargo.toml` which *will* just be the version number.

src/error.rs
```diff
@@ -44,23 +49,24 @@ pub enum Error {
 
 impl Error {
     /// Construct a CLI error.
-    pub fn cli(message: &str, stderr: Cow<str>) -> Result<(), Self> {
-        Err(Error::Cli {
+    pub fn cli(message: &str, stderr: Cow<str>) -> Self {
```
fitzgen: These changes to the `Error` constructors are making this PR a bit noisier than it otherwise would be; unless this was requested for this PR, I'd have personally left them for a separate PR.

data-pup:
Thanks for the feedback @fitzgen! I can resolve the merge conflicts today, and take care of some of your inline comments as well. I can have something EOD, since this relates to a 'Blocking Rust 2018' issue :crab:

src/error.rs
data-pup: I agree that these are a little noisy, I can totally remove these changes. The tldr of why they are there is that these methods were causing type errors when I tried to use them from within a function that didn't also return a `Result<(), Error>`, but you are right that making that change separately is probably wiser :)

src/bindgen.rs
data-pup: Ah! Good catch. I can take care of that.

src/bindgen.rs
data-pup: No reason, I hadn't known about those. I can fix this :)

data-pup:
Continuing this work in #244, because the merge conflicts got a little too drastic.

Issue #160, opened 2018-06-07
Add some kind of toml configuration


fitzgen:
As we add more options and flags (#159 #158 #153 etc) it would be great if we didn't have to keep passing the same command line arguments for every build all the time, and instead could canonicalize configurations in a toml file.

ashleygwilliams:
@fitzgen it would be really useful in discussing this if we had a strawman proposal. do you think you could write up an example of what you think it might look like and we can discuss from there? thanks! (of course, anyone else reading this should also feel free to post an example of how they think configuration should work!)

ashleygwilliams:
(also, i updated the title because, while i think we'll likely want to use TOML, i don't want to close the convo off to potentially other formats, assuming there's a good rationale, of course)

fitzgen:
This is what I imagine the defaults from #153 would look like, as described in a strawman toml format

```toml
[build.debug]
debug = true
features = ["console_error_panic_hook"]
rustc-opt-level = 0
wasm-opt = false
wasm-snip = false
wasm-bindgen = true

[build.profiling]
debug = true
features = ["console_error_panic_hook", "wee_alloc"]
rustc-opt-level = "s"
wasm-opt = { opt-level = "z" }
wasm-snip = false
wasm-bindgen = true

[build.production]
debug = false
features = ["wee_alloc"]
rustc-opt-level = "s"
wasm-opt = { opt-level = "z" }
wasm-snip = { snip-rust-fmt-code = true, snip-rust-panicking-code = true }
wasm-bindgen = true
```

steveklabnik:
After trying tools that use an external `.toml` file vs putting configuration in `Cargo.toml`, I would argue that putting it in `Cargo.toml` is much nicer. All of your configuration is in one file, rather than an increasing number of `.toml` files.  See https://github.com/rust-osdev/bootimage#configuration for example; you could put this in

`[package.metadata.wasm-pack.build.debug]`

instead of just

`[build.debug]`.

Issue #188, opened 2018-06-18
split `init` command into `init` and `build`


ashleygwilliams:
we've been *packing* (pun intended) a ton of stuff into the `init` command. this has the benefit of being a one-stop shop, but as we've seen with the intro of the `--mode` flag it's a lot harder to customize, and for those who use wasm-pack as a build tool, a single command that does *everything* is not really what folks want. for 0.5.0 let's break out `init` into a `init` and `build` command.

strawman proposal:

- `init` becomes what `init --mode no-build` is now
- `build` is everything `--mode no-build` skips, build should have a `--no-installs` flag

thoughts?

ashleygwilliams:
related: #185

fitzgen:
SGTM.

Ideally, `--no-installs` will be unnecessary as well, once we can properly determine whether the dependency tools are available and up-to-date or not.

csmoe:
Given the original `init` options,
```rust
pub struct Init {
    crate_path: String,
    scope: Option<String>,
    disable_dts: bool,
    target: String,
    debug: bool,
    crate_name: String,
}
```
#194 introduced the configuration: 
```rust
[package.metadata.wasm-pack.build.<build_mode>]
debug = true
features = ["console_error_panic_hook"]
rustc-opt-level = 0
wasm-opt = false
wasm-snip = false
wasm-bindgen = true
```
In the addition of `build` options above, should I move `init.debug` into `build` too?

ashleygwilliams:
hey @csmoe sorry i missed this comment. let me get the split done (hopefully today or tomorrow) and then we can figure out exactly where things should be. it'll be easier once it's actually in place.

ashleygwilliams:
closed by #216

PR #163, opened 2018-06-11
[WIP] Add support for running wasm-snip

csmoe:
Closes #158

Make sure these boxes are checked! 📦✅

- [X] You have the latest version of `rustfmt` installed and have your 
      cloned directory set to nightly
```bash
$ rustup override set nightly
$ rustup component add rustfmt-preview --toolchain nightly
```
- [X] You ran `rustfmt` on the code base before submitting
- [X] You reference which issue is being closed in the PR text

✨✨ 😄 Thanks so much for contributing to wasm-pack! 😄 ✨✨

PR #194, opened 2018-06-21
[WIP] Add toml like configuration

csmoe:
Address #160 

Config format: (from https://github.com/rustwasm/wasm-pack/issues/160#issuecomment-396413387):
```toml
[package.metadata.wasm-pack.build.<build_mode>]
debug = true
features = ["console_error_panic_hook"]
rustc-opt-level = 0
wasm-opt = false
wasm-snip = false
wasm-bindgen = true
```
`build_mode = debug | profiling | production`


ashleygwilliams:
this is an awesome start @csmoe let me know if you need any help! i think the plan for landing this should be coordinated with the work in #163  and we'll need to get the other commands in there as well. i can start on that (or get others to start on that!) soon!

mgattozzi:
@csmoe when you feel it's good for review just ping us :) I also just wanted to say I'm super stoked for this so thank you for working on it.

fitzgen:
Is this actually still WIP? Or waiting on review?

csmoe:
@fitzgen those changes were introduced into `manifest` locally indeed, but I don't know how to test for them since there is no [build](https://github.com/rustwasm/wasm-pack/issues/188) cmd, maybe I can take care of the splitting stuff, could you take a look at the no-response comment?https://github.com/rustwasm/wasm-pack/issues/188#issuecomment-400230511

ashleygwilliams:
@csmoe if i am correct this is waiting on #216 ? if so, you can rebase onto that branch and then change the PR to be into that PR so we can be ready to merge when that lands (ideally today). thank you so much for all your hard work and especially for your patience!

additionally, i think we'll want some *very basic* docs and tests for this. lemme know if you want some direction on that.

csmoe:
@ashleygwilliams yes, I'm waiting for `build`.
will restart tomorrow since it's midnight in my timezone.

src/manifest.rs
```diff
@@ -129,9 +214,100 @@ impl CargoManifest {
...
+    }
+}
+
+impl WasmPackConfig {
+    fn from_table(table: Table) -> Result<WasmPackConfig, Error> {
+        let mut wasm_pack_config = WasmPackConfig::default();
+        for (key, value) in table {
+            match (key.as_str(), value.clone()) {
+                ("debug", Value::Boolean(b)) => wasm_pack_config.debug = From::from(b),
+                ("features", Value::Array(a)) => {
```
csmoe: how does features work?
@ashleygwilliams  impl From\<Features\> for [CargoTomlFeaturesSection](https://doc.rust-lang.org/cargo/reference/manifest.html#the-features-section)?

PR #195, opened 2018-06-25
Fix code example in README

steveklabnik:
This won't compile if it isn't pub!

Fixes https://github.com/rustwasm/wasm-bindgen/issues/309


mgattozzi:
Thanks @steveklabnik! :D

Issue #189, opened 2018-06-18
publish and pack commands have confusing path expectations


ashleygwilliams:
currently the `publish` and `pack` commands anticipate that you are running them from a directory that contains a `pkg` directory to be published. i have confusingly `cd`'d into the `pkg` directory and tried and gotten an error, e.g.:

```
Ashleys-MBP-2:wasm-pack ag_dubs$ cd tests/fixtures/js-hello-world/pkg/
Ashleys-MBP-2:pkg ag_dubs$ wasm-pack publish
| ⛔   [ERR]: There was an I/O error. Details:
No such file or directory (os error 2)
Caused by: No such file or directory (os error 2)
```

i think it's working fine as is, but perhaps we want to have a custom error message here that explains the situation, and/or have an option to pass a path to the pkg directory, not just the parent directory? unclear exactly what's the best UX here- would love to hear ya'lls thoughts!

Mackiovello:
I would like to give this a try! I think it should work similarly to how [`dotnet publish`](https://docs.microsoft.com/en-us/dotnet/core/tools/dotnet-publish?tabs=netcore21#arguments) works: 

* It uses the current directory by default
* You can optionally set the directory as the first argument
* If there's nothing to publish in the directory, it gives a nice error message: `Specify a project or solution file. The current working directory does not contain a project or solution file.`

It may not be the best comparison, but it's very similar to what you propose and what feels intuitive in my mind. 

I'm relatively new to Rust, but I think I can implement this and create a pull request.

mgattozzi:
Yeah! I believe this is the correct way and kind of what we would want to do. So what happens is that it just uses the I/O error which is "Can't find the file". Best way to make this nicer is the following:

- Before we publish create a `Path` to the `package.json`
- Check that it [exists](https://doc.rust-lang.org/std/path/struct.Path.html#method.exists)
- If it does not keep trying the different methods you mentioned for path resolution
- If it exists then continue with the publishing
- Otherwise throw an error

The key thing here is you'll need to create a new error so that you can have that custom string you mentioned! You can do that in `src/error.rs`

Thanks for taking this! Feel free to ping here or in a WIP PR if you need help :D

Mackiovello:
After looking at this some more, I think it's surprising that the user-entered path should point to the parent of the pkg directory. I don't know any tool that works this way. Would it be alright to change this behavior? 

[...]

PR #198, opened 2018-07-04
Improve UX for pack and publish path expectations

Mackiovello:
Okay, I have to take this piece by piece. I'm creating this as a WIP pull request to some pressure on myself but also to make it easier to ask for help.

If anyone sees anything that can be improved or think should be changed, please tell me. I'm new to Rust, so every little hint helps. I'll put some questions in my own code for my own reference but also for you to answer if you have time. 

This is for #189 

---

__saving this for later__

```bash
$ rustup override set nightly
```


src/command/pack.rs
```diff
@@ -9,7 +9,15 @@ pub fn pack(path: Option<String>, log: &Logger) -> result::Result<(), Error> {
     let crate_path = set_crate_path(path);
 
     info!(&log, "Packing up the npm package...");
-    npm::npm_pack(&crate_path)?;
+    match npm::npm_pack(&crate_path) {
+        Ok(r) => r,
+        Err(Error::Io { .. }) => {
```
Mackiovello: I think this match is too broad. I only want to catch os error 2: directory not found. I don't want to swallow other IO errors. Investigate how to do that.

src/command/pack.rs
```diff
@@ -9,7 +9,15 @@ pub fn pack(path: Option<String>, log: &Logger) -> result::Result<(), Error> {
     let crate_path = set_crate_path(path);
 
     info!(&log, "Packing up the npm package...");
-    npm::npm_pack(&crate_path)?;
+    match npm::npm_pack(&crate_path) {
+        Ok(r) => r,
```
Mackiovello: This should probably just be `Ok(r) => (),`

src/command/pack.rs
```diff
@@ -9,7 +9,15 @@ pub fn pack(path: Option<String>, log: &Logger) -> result::Result<(), Error> {
...
+        Err(Error::Io { .. }) => {
+            return Err(Error::DirNotFound {
+                message: "Unable to find the pkg directory".to_owned(),
```
Mackiovello: Should this be more specific? Should it also print the path that the user entered? Not sure yet.

src/command/pack.rs
```diff
@@ -9,7 +9,15 @@ pub fn pack(path: Option<String>, log: &Logger) -> result::Result<(), Error> {
...
+            });
+        }
+        Err(e) => return Err(e),
```
Mackiovello: These explicit `return` statements feel like code smells. Maybe I can use the `?` operator here?

src/error.rs
```diff
@@ -45,6 +47,7 @@ impl Error {
             Error::CrateConfig { message: _ } => {
                 "There was a crate configuration error. Details:\n\n"
             }
+            Error::DirNotFound { message: _ } => "Unable to find the directory\n\n",
```
Mackiovello: As I understand it, this works as a default message. I considered using a more specific `PkgNotFound` error, but if everyone made specific error like that, it would be a mess to work with. This feels easier to use.

src/command/pack.rs
Mackiovello: I should try something like this instead:

```rust
my_function_call().map_err(|e| match {
  interesting_error => something_else,
  stuff => stuff,
})?
```

Mackiovello:
@mgattozzi could you give this a quick look? Am I on the right path?

Also, how should I do with tests? I don't like shipping code without unit tests. It looks like I have to inject the `npm` functions in order to test. Is there a better way as you see it?

The code is repeated between pack and publish. Do you believe it should be abstracted or do we keep it like this?

ashleygwilliams:
thanks so much for this @Mackiovello ! i'll give a review by EOD today :)

Mackiovello:
This does not work as expected right now. I'll look at it later today. Don't review yet

Mackiovello:
Never mind, it actually works. I was confused since I got `os error 2`, but that was because I didn't have NPM installed. Here's how it works with NPM installed:

```
$ cargo run -- pack myproject/pkg
   Compiling wasm-pack v0.4.1 (file:///home/erlend/rust/wasm-pack)
    Finished dev [unoptimized + debuginfo] target(s) in 2.46s
     Running `target/debug/wasm-pack pack myproject/pkg`

| 🎒  packed up your package!
$ cargo run -- pack myproject
    Finished dev [unoptimized + debuginfo] target(s) in 0.06s
     Running `target/debug/wasm-pack pack myproject`

| 🎒  packed up your package!
$ cargo run -- pack myproject/src/
    Finished dev [unoptimized + debuginfo] target(s) in 0.06s
     Running `target/debug/wasm-pack pack myproject/src/`
Unable to find the pkg directory at path 'myproject/src/', or in a child directory of 'myproject/src/'
erlend@mackiovello:~/rust/wasm-pack$
```

Mackiovello:
Ran `cargo +nightly fmt` with no result:

``` 
$ cargo +nightly fmt
$ git status
On branch issue/189
Your branch is up-to-date with 'origin/issue/189'.
nothing to commit, working directory clean
```

Mackiovello:
@ashleygwilliams this is ready for a review :slightly_smiling_face:

ashleygwilliams:
hey @Mackiovello ! thanks so much for this. so far it looks great! do you think you could add some docs and perhaps a test (let me know if you run into trouble with this one)? since this is about expectations, i think some docs would be great :) should be close to getting this in tho- thanks so much again!

Mackiovello:
Thank you @ashleygwilliams! One question: Where do you want documentation? On the `pack` and `publish` functions or somewhere else? 

Also, should the tests be unit tests with no access to the file system or should it be integration tests that are allowed to use the file system? I think it should be fine with unit tests, but then I have to inject functions like `is_dir` and mock it in the tests. It'll make the code a little messier.

ashleygwilliams:
@Mackiovello so- for the docs, i would add both API docs (not too specific to end user, just add descriptions of the functionality generally), and then and more user-facing specific docs to the docs in the `docs` dir. looks like `pack` and `publish` are separate at the moment, but i would be happy if you combined them since they share this behavior you are adding (and it's by far the most complicated thing about them, so it's probably best to have that doc'd in a single place). 

for the tests, all our tests currently use the filesystem by accessing code in the `tests/fixtures` directory. it's not *the best* but it's fine for now. mocking is such a pain in rust anyways so i think you are right that this is the least messy way :)

lemme know if you have any other questions.

PR #216, opened 2018-07-16
Split init into init and build

csmoe:
@ashleygwilliams Sorry for missing your comment [there](https://github.com/rustwasm/wasm-pack/issues/188) since I was nearly done at that time.


ashleygwilliams:
looking great so far @csmoe thanks for taking this on! this will def require some docs updates! i can get you a review by tomorrow, but in the meantime- if you want to add docs that would be great!

mgattozzi:
Hey @csmoe mind rebasing and fixing the conflicts? I can give you a review as soon as that's done now that I'm back :)

csmoe:
@ashleygwilliams I copied the workflow from your [comment](https://github.com/rustwasm/wasm-pack/issues/188#issue-333360598) there:
   - init becomes what init --mode no-build is now
   - build is everything --mode no-build skips, build should have a --no-installs flag

I was also puzzled by the repeated init-steps in `build`(sorry for not asking for your suggestion at that time), but I just simply left it as it was.

>  i think our goal would be to keep those separate? e.g. if you run wasm-pack build and haven't run init it asks you to run init first?

I love this.

docs/init.md
```diff
@@ -1,9 +1,9 @@
-# wasm-pack init
+# wasm-pack init(Deprecated)
```
xtuc: nit: space before `(`

src/command/build.rs
```diff
@@ -1,113 +1,92 @@
...
+    #[structopt(long = "mode", short = "m", default_value = "normal")]
+    /// Sets steps to be run. [possible values: no-install, normal]
+    pub mode: String,
+
+    #[structopt(long = "no-typescript")]
+    /// By default a *.d.ts file is generated for the generated JS file, but
+    /// this flag will disable generating this TypeScript file.
+    pub disable_dts: bool,
+
+    #[structopt(long = "target", short = "t", default_value = "browser")]
```
xtuc: btw `browser` is not accurate. ESM will be soon available in Nodejs, which isn't a browser.

The correct targets are:
- CJS (commonjs)
- ESM (ECMAScript modules)

csmoe: > ESM will be soon available in Nodejs, which isn't a browser.

so should I leave a FIXME note here or change it to `ESM` right now?

ashleygwilliams: leave whatever the current behavior is, and let's file an issue to update.

README.md
```diff
@@ -29,9 +29,10 @@ visiting that repo!
 
 ## 🎙️ Commands
 
+- [`init`](docs/init.md): [**Deprecated**] Initialize an npm wasm pkg from a rustwasm crate
+- [`build`](docs/build.md): Generate an npm wasm pkg from a rustwasm crate
 - [`init`](docs/init.md): Generate an npm wasm pkg from a rustwasm crate
```
mgattozzi: Looks like this line wasn't removed despite having added the deprecated tag

PR #244, opened 2018-08-01
Coordinate wasm-bindgen versions.

### LINKED ISSUE #146: coordinating wasm-bindgen versions

data-pup:
Fixes #146. This aims to solve the issue of making sure that there is (a) a locally installed wasm-bindgen and (b) that it is the same version as specified in the Cargo.toml file.

Rebasing the previous PR ended up being too much effort after some intensive changes were made in the meantime, so I re-did the work off the current master branch.

While doing so, I fixed *most* of the issues that @fitzgen mentioned in their review. There is one remaining detail to take care of tomorrow, regarding how the output of `wasm-bindgen --version` is parsed. I thought it would be useful to get this new PR open before I take care of that, so the rest of this can be reviewed :)

---

```bash
$ rustup override set nightly
```


src/bindgen.rs
```diff
@@ -42,33 +73,88 @@ pub fn wasm_bindgen_build(
...
+fn wasm_bindgen_version_check(bindgen_path: &PathBuf, dep_version: &str) -> Result<bool, Error> {
+    let wasm_bindgen = Path::new(bindgen_path);
+    let output = Command::new(wasm_bindgen).arg("--version").output()?;
+    if output.status.success() {
+        let s = String::from_utf8_lossy(&output.stdout);
+        let installed_version = s.trim(); // FIXUP: Trim this more thoroughly?
+        Ok(installed_version == dep_version)
+    } else {
+        // FIXUP: This error message can be improved.
+        let message = String::from("Could not find version of local wasm-bindgen");
```
data-pup: As the "FIXUP" comment mentions, I think this error message can be improved. Parsing the output of the `wasm-bindgen --version` command will also be some fixup work I can get finished tomorrow.

src/bindgen.rs
```diff
@@ -42,33 +73,88 @@ pub fn wasm_bindgen_build(
...
+        None
+    }
+}
+
+/// Return a string containing the path to the global `wasm-bindgen`, if it exists.
+fn global_wasm_bindgen_path_str() -> Option<PathBuf> {
+    if let Ok(path) = which("wasm-bindgen") {
+        Some(path)
+    } else {
+        None
```
data-pup: Other details I know I need to take care of: These functions can be collapsed into a single function, and should be named `wasm_bindgen_path` rather than being suffixed `path_str`, since these return `PathBuf` values rather than strings now.

src/manifest.rs
```diff
@@ -199,6 +199,9 @@ fn check_wasm_bindgen(path: &Path) -> Result<(), Error> {
         "Ensure that you have \"{}\" as a dependency in your Cargo.toml file:\n[dependencies]\nwasm-bindgen = \"0.2\"",
         style("wasm-bindgen").bold().dim()
     ))
+    // FIXUP: Use the `get_wasm_bindgen_version` function for this?
+    // let _ = get_wasm_bindgen_version(path)?;
+    // Ok(())
```
data-pup: This also needs to get cleaned up. Pardon!

src/bindgen.rs
```diff
@@ -3,26 +3,57 @@
...
+    version: &str,
+    install_permitted: bool,
+    step: &Step,
+) -> Result<(), Error> {
+    // Determine if the `wasm-bindgen` dependency is already met for the given mode.
+    let dependency_met = if let Some(bindgen_path) = wasm_bindgen_path_str(path) {
+        wasm_bindgen_version_check(&bindgen_path, version).unwrap_or_else(|_| false)
+    } else {
+        false
+    };
```
data-pup: This is a little wonky right now, because the version check returns a result. As mentioned below, I think I can edit that so that it returns a plain boolean value, which should make this much nicer.

data-pup:
I added some various nits that I already see, but if I missed anything do let me know! I should have this cleaned up tomorrow.

data-pup:
This is all cleaned up and ready for review now :smile_cat: Let me know if there are any changes that I should make!

Cargo.toml
```diff
@@ -23,6 +23,7 @@ slog-term = "2.4"
 slog-async = "2.3"
 structopt = "0.2"
 toml = "0.4"
+which = "2.0.0"
```
mgattozzi: I didn't know this existed and now I'm happy to see it does.

src/bindgen.rs
```diff
@@ -3,26 +3,53 @@
...
+
     let msg = format!("{}Installing WASM-bindgen...", emoji::DOWN_ARROW);
     PBAR.step(step, &msg);
     let output = Command::new("cargo")
         .arg("install")
-        .arg("wasm-bindgen-cli")
         .arg("--force")
+        .arg("wasm-bindgen-cli")
+        .arg("--version")
+        .arg(version)
```
mgattozzi: My understanding is that this will keep them pinned to the same version as the package, but if they use a new one in a different project it'll overwrite it?

How do we account for versions such as "0.2" which allows for any version from ">= 0.2 && < 0.3"? which version should be chosen in that case? The one in the lockfile? I only realized we'd have to answer this question just now unfortunately :/

mgattozzi: Oh I see now further below you're specifying to install it locally and to use that one like was discussed. Awesome :D I still wonder about the versioning in the second part of my comment.

src/bindgen.rs
```diff
@@ -42,33 +69,81 @@ pub fn wasm_bindgen_build(
...
-        .arg(dts_arg)
-        .arg(target_arg)
-        .output()?;
-    if !output.status.success() {
-        let s = String::from_utf8_lossy(&output.stderr);
-        Error::cli("wasm-bindgen failed to execute properly", s)
-    } else {
-        Ok(())
-    }
+    local_bindgen_path(crate_path).or_else(global_bindgen_path)
```
mgattozzi: This is a pretty clever usage of closures I really like it :)

src/command/build.rs
```diff
@@ -71,21 +72,27 @@ impl Build {
     pub fn try_from_opts(build_opts: BuildOptions) -> Result<Self, Error> {
         let crate_path = set_crate_path(build_opts.path);
         let crate_name = manifest::get_crate_name(&crate_path)?;
+        let mode = match build_opts.mode.as_str() {
+            "no-install" => BuildMode::Noinstall,
+            "normal" => BuildMode::Normal,
```
mgattozzi: Is there a reason to have this if it always defaults to normal?

src/command/build.rs
```diff
@@ -213,8 +220,19 @@ impl Build {
     }
 
     fn step_install_wasm_bindgen(&mut self, step: &Step, log: &Logger) -> Result<(), Error> {
+        info!(&log, "Identifying WASM-bindgen dependency...");
```
mgattozzi: s/WASM/wasm/g

src/command/build.rs
data-pup: Not really! I had included both cases for aesthetic reasons only, but I can clean that up.

src/bindgen.rs
```diff
@@ -3,26 +3,53 @@
...
+    version: &str,
+    install_permitted: bool,
+    step: &Step,
+) -> Result<(), Error> {
+    // If the `wasm-bindgen` dependency is already met, print a message and return.
+    if wasm_bindgen_path(path)
+        .map(|bindgen_path| wasm_bindgen_version_check(&bindgen_path, version))
+        .unwrap_or(false)
+    {
+        let msg = format!("{}WASM-bindgen already installed...", emoji::DOWN_ARROW);
```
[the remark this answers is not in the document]
data-pup: Can do!

src/bindgen.rs
data-pup: @mgattozzi Regarding the second part of your first comment, would this crate be a sensible solution? https://crates.io/crates/semver

Still thinking about how its `VersionReq` struct would be used to find the most recent version that satisfies a requirement like `>= 0.2 && < 0.3`, but it seems like a promising library for this.

mgattozzi: That is the canonical semver crate. I'm just thinking like:

I build a program using "0.2" as my version
Cargo.lock has it at 0.2.8

A new release is out and wasm-bindgen is at 0.2.9

Which one do you use? The lockfile or new crates.io? I think because it's a library the newest one on crates.io that fits the description, but that also means running cargo update which may not be desired behavior. I'm more asking this question to you and also @ashleygwilliams and whoever else might have a stake in answering this.

fitzgen: I think we should always follow the lock file. Since we do (or will do) a `cargo build` before getting the `wasm-bindgen` CLI, the lock file will always have the exact wasm-bindgen used to build the `.wasm` binary.

data-pup: Following the lock file makes a lot of sense to me! If it makes sense to everybody else, I could try adjusting the `get_wasm_bindgen_version` function so that it looks through the `Cargo.lock` file rather than the `.toml` file.

If I *do* go this route, should `get_wasm_bindgen_version` end up in a new file, i.e. `lockfile.rs` given that we wouldn't be checking the manifest itself?

data-pup:
@mgattozzi, this is ready for review now!

data-pup:
Yay! :tada: Really happy to hear :)

mgattozzi:
@data-pup Thanks for this. I'm opened up an issue in #270

src/manifest.rs
```diff
@@ -212,3 +212,28 @@ fn check_crate_type(path: &Path) -> Result<(), Error> {
       "crate-type must be cdylib to compile to wasm32-unknown-unknown. Add the following to your Cargo.toml file:\n\n[lib]\ncrate-type = [\"cdylib\"]"
     )
 }
+
+/// Get the version of `wasm-bindgen` specified as a dependency.
+pub fn get_wasm_bindgen_version(path: &Path) -> Result<String, Error> {
+    if let Some(deps) = read_cargo_toml(path)?.dependencies {
+        match deps.get("wasm_bindgen") {
```
xmclark: Is this correct? I thought it would be "wasm-bindgen". This line is giving me problems locally.

fitzgen: See https://github.com/rustwasm/wasm-pack/pull/267

Issue #159, opened 2018-06-07
Add support for running wasm-opt


fitzgen:
And configuring:

* (most importantly) its optimization level (ie `-Os` vs `-Oz` vs `-O3` etc)
* what passes are enabled or disabled
* maybe also installing it if necessary?

https://github.com/WebAssembly/binaryen

ashleygwilliams:
i think that we should fundamentally expose this through #153 - in line with our desire to have a more convention over configuration tool to expose pre-configured "build" options to the user. do those three build modes cover the behavior we'd want to see @fitzgen or is there more we would want to expose? i'm very cautious about overwhelming the use with options and configuration (tho #160 will certainly help with that)

fitzgen:
I think we definitely need to expose control over optimization level. The other knobs I am "meh" on.

Regarding pre-configured builds: totally agree! I also don't think that is mutually exclusive with exposing knobs to change the default of some setting within each build. For example, cargo lets you do this with things like:

```
[profile.release]
# By default release doesn't build with debug info, but we are customizing that here.
debug = true
# Tell rustc to optimize for size, rather than its default of optizing for speed.
opt-level = "s"
```

I think we could do the same kind of default build configurations that do the 95% thing out of the box + knobs if you need them for the other 5% approach.

csmoe:
I'll take care of this.

pepyakin:
For what it worth, [binaryen-rs](https://github.com/pepyakin/binaryen-rs) has just gained support for setting optimization levels. This allows for achieving same functionality as with `wasm-opt`.

ashleygwilliams:
now that we've landed toml config in 0.6.0 i think it's time to look at this again and get it into 0.7.0.

decisions to make:

1. how to run `wasm-opt`
    - should we run a child process with a wasm-opt binary release (from a fork of emscripten perhaps? since the release story there is complicated?)
    - should we leverage `binaryen-rs`

    personally if `binaryen-rs` is well-maintained this may be the best option.

2. configuration
    i think we should leverage the build profiles. you should be able to set a key (`wasm-opt` or `opt` perhaps) to a direct set of `wasm-opt` args, or, some preset keywords (e.g. `size`, `speed`)

   i imagine this looking like

   ```toml
   [package.metadata.wasm-pack.profile.dev]
   opt = "size" #alias for the -s flag

   [package.metadata.wasm-pack.profile.dev.wasm-bindgen]
   debug-js-glue = true
   ```

considerations:
[...]

PR #245, opened 2018-08-02
[WIP] Introduce wasm-opt

csmoe:
Address #159 
blocked by #194 ,wasm-opt and [wasm-snip](https://github.com/rustwasm/wasm-pack/pull/163) should be merged at same time.

### DIFF
--- a/src/command/build.rs
+++ b/src/command/build.rs
@@ -1,4 +1,5 @@
 use bindgen;
+use opt;
 use build;
 use command::utils::{create_pkg_dir, set_crate_path};
 use emoji;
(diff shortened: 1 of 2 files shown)


nicksrandall:
Is this still blocked? Looks like #194 has been closed.

nicksrandall:
I wonder if it would be worth integrating directly with https://github.com/pepyakin/binaryen-rs (which now supports opts) rather than shelling out to cli command. That *might* be easier than making sure `wasm-opt` is always installed in the user's machine.

ashleygwilliams:
closing in favor of #625

PR #246, opened 2018-08-02
the infamous docs pr

ashleygwilliams:
screen shot 2018-09-23 at 12 40 03 pm
screen shot 2018-09-23 at 12 40 08 pm


docs/src/tutorial/hello-wasm-pack.md
```diff
@@ -0,0 +1,137 @@
...
+Let's step through this line by line. First up is the list of nightly features. We're enabling this for
+the whole crate. What this means is that we will later tag code with an attribute and this will
+allow Rust to generate code that we don't have to write by hand. In our case it'll use
+`wasm-bindgen.` It should be noted that `#![feature(...)]` implies using the nightly
+compiler. This gated feature will hopefully be stabilized and landed soon so that you won't need it!
+
+`wasm-bindgen` knows how to make code that works well with wasm so we don't have to
+worry about it too much and just write Rust code for the most part. If you want to know the full
+extent of its capabilities check out the README on its repo which can be found
+[here](https://github.com/alexcrichton/wasm-bindgen). For our purposes we need to know that if we
```
fitzgen: Maybe

> If you want to know the full extent of its capabilities, [check out its README](https://github.com/alexcrichton/wasm-bindgen).

docs/src/tutorial/hello-wasm-pack.md
```diff
@@ -0,0 +1,137 @@
...
+want functions to work with wasm easily we'll need it.
+
+The next line says we're importing the `wasm-bindgen` crate and the line after that imports the
+prelude from `wasm-bindgen`. The `extern crate` call lets the compiler know what crates to link in
+and the `prelude` contains all the types and functions that `wasm-bindgen` needs to work properly!
+
+Cool let's import the `alert` function from JS so that we can call it in our Rust code!
```
fitzgen: Suggest comma (or exclamation) after cool:

> Cool, let's import ...

docs/src/tutorial/hello-wasm-pack.md
```diff
@@ -0,0 +1,137 @@
...
+
+```rust
+#[wasm_bindgen]
+extern {
+    fn alert(s: &str);
+}
+```
+
+Alright so we have our external bit of code and we have everything imported so let's write the
+actual `add` function, as well as an `add_alert` function that will use `add` in itself but also
```
fitzgen: In the code it is named `alert_add`, not `add_alert`.

ashleygwilliams:
heh this is all text from the previous tutorial, i just changed the name of the file. i have more incoming on this. thanks for the comments. i just forgot to mark this WIP because i was sneaking it in right before the core team meeting.

docs/src/tutorial/hello-wasm-pack.md
mgattozzi: Maybe also the final line could say this instead:

> For this tutorial all we need to know is that wasm-bindgen will help us generate code that lets us call our wasm code from JS without us needing to know the nitty gritty details.

docs/src/tutorial/hello-wasm-pack.md
```diff
@@ -0,0 +1,137 @@
...
+test it_works ... ok
+test result: ok. 1 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out
+```
+
+Yay it all works! Notice we didn't add a test for `alert_add`. This is because Rust won't know what
+`alert` is unless the wasm code is running in the browser! Don't worry though. Once we package this
+code up and upload it to npm we'll then test out that function to make sure everything works like we
+expect it too!
+
+You can find all of the above code [here](https://github.com/mgattozzi/wasm-add).
```
mgattozzi: Should we move this repo to rustwasm as well? Let me know I'm happy to transfer it and then we can update this line

fitzgen:
> I used this version of the book to get going and it worked well for me. Good work so far. +1

<3 <3 <3

ashleygwilliams:
the tests are failing, but only sometimes and just for the test tests, as filed in #344 gonna merge and then probably fiddle with gh-pages shenanigans cuz i never get it right on the first try

Issue #253, opened 2018-08-10
Links in readme don't point to docs/src


hoodie:
## 🐛 Bug description
The links in the readme don't work on github

#### 🤔 Expected Behavior
I should have played with wasm-pack itself instead.

#### 👟 Steps to reproduce
1. https://github.com/rustwasm/wasm-pack
2. click on commands documentation
3. resist obivan octocatobi trying to convince you that this is "not the web page you are looking for"

#### 🌍 Your environment
Mild 23° on a Friday night, tea and some light music

PR #254, opened 2018-08-10
fix(doc): fix links in README

hoodie:
# The obligatory README-link-fix-PR

```bash
$ rustup override set nightly
```
- [x] You reference which issue is being closed in the PR text #253

### DIFF
--- a/README.md
+++ b/README.md
@@ -29,9 +29,8 @@ visiting that repo!
 
 ## 🎙️ Commands
 
-- [`init`](docs/init.md): Generate an npm wasm pkg from a rustwasm crate
-- [`build`](docs/build.md): Generate an npm wasm pkg from a rustwasm crate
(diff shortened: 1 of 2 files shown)


migerh:
While you're at it, can you please fix the two links below [Prerequisites](https://github.com/rustwasm/wasm-pack#-prerequisities) as well? All it takes is a set of well placed `src` :slightly_smiling_face:

ashleygwilliams:
hey folks! this is actually the result of a sloppy PR to migrate the docs to an mdbook that didn't deploy to gh-pages that happened right before i got sick :( really sorry about that. going to fix it now, so this PR likely won't be needed, but i will leave it open until the problem is fixed.

fitzgen:
Going to merge this, since it is better to have the links working even while we are waiting for the incoming mdbook :)

Issue #247, opened 2018-08-02
reorg and fix .travis.yml to account for both release tasks and book deploy tasks


ashleygwilliams:
similar to what @alexcrichton recently did for wasmbindgen: https://github.com/rustwasm/wasm-bindgen/blob/master/.travis.yml

xmclark:
I made #259

mgattozzi:
Closed in #259! Thanks @xmclark!

PR #259, opened 2018-08-23
Reorg travis deployment

xmclark:
Addresses #247 

1. New github deploy macro in travis
2. Deploy artifacts and book in their own job
3. Use job name environment variable
4. Copy notification settings from wasm-bindgen travis
5. Cache cargo for faster builds

---

```bash
$ rustup override set nightly
```

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,76 +1,77 @@
 language: rust
 sudo: false
+cache: cargo
+
+DEPLOY_TO_GITHUB: &DEPLOY_TO_GITHUB
+  before_deploy:
+    - git config --local user.name "Ashley Williams"
+    - git config --local user.email "ashley666ashley@gmail.com"
+    - name="wasm-pack-$TRAVIS_TAG-$TARGET"
+    - mkdir $name
+    - cp target/$TARGET/release/wasm-pack $name/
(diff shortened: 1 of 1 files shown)


xmclark:
A few comments:

I was tempted to start the discussion earlier, but I felt that showing something, then starting discussion would be more valuable.

I think this PR is what the ticket was asking for. Co-locate deployment with building artifacts in each travis job.

I also borrowed a few other patterns used in the wasm-bindgen travis file like using named jobs (this helps when looking at the travis jobs)  and turning some notifications off.

I added cargo caching. I noticed that wasm-bindgen was **not** doing this. I am open to reverting that commit if the community feels it is more valuable to install dependencies every time. I added it because I like faster builds and right now the builds are kind of long.

Also, I could not get latest cargofmt to work on my mac. It hangs with no output 😢 Thankfully no rust changes here, but I will need to get it working at some point.

fitzgen:
@xmclark thanks for diving into this! Would you mind rebasing on master, to get the fix for slog that just merged into master? And while you're at it, could you squash the commits into a single one?

Thanks!

xmclark:
Ok, I had to stretch my git knowledge for that one. Rebased and squashed. Good to go! ✅

PR #283, opened 2018-09-07
Add listing for precompiled Windows binaries

alexcrichton:
They're fixed on wasm-bindgen's CI, so let's download/install them!


src/bindgen.rs
```diff
@@ -99,12 +99,15 @@ fn curl(url: &str) -> Result<Vec<u8>, failure::Error> {
...
     let macos = cfg!(target_os = "macos");
+    let windows = cfg!(windows);
     let x86_64 = cfg!(target_arch = "x86_64");
 
     let target = if linux && x86_64 {
         "x86_64-unknown-linux-musl"
     } else if macos && x86_64 {
         "x86_64-apple-darwin"
+    } else if windows && x86_64 {
+        "x86_64-pc-windows-msvc"
```
fitzgen: can you also update the tests that are `cfg`ed to match whether we can download prebuilt binaries? `tests/all/bindgen`

alexcrichton: Sure thing, done!

fitzgen:
Looks like this ran into the same bug I fixed in the wasm-pack test PR: https://ci.appveyor.com/project/ashleygwilliams/wasm-pack-071k0/build/1.0.213#L256

alexcrichton:
Oh actually that looks like it's missing a `.exe` for windows

tests/all/bindgen.rs
```diff
@@ -4,12 +4,18 @@ use wasm_pack::bindgen;
...
-    all(target_os = "macos", target_arch = "x86_64")
+    all(target_os = "macos", target_arch = "x86_64"),
+    all(windows, target_arch = "x86_64"),
 ))]
 fn can_download_prebuilt_wasm_bindgen() {
+    use std::env;
     let dir = tempfile::TempDir::new().unwrap();
-    bindgen::download_prebuilt_wasm_bindgen(dir.path(), "0.2.19").unwrap();
-    assert!(dir.path().join("bin").join("wasm-bindgen").is_file());
+    bindgen::download_prebuilt_wasm_bindgen(dir.path(), "0.2.21").unwrap();
```
ashleygwilliams: curious about updating the version number here- that's maintenance we aren't going to want to be doing...

alexcrichton: Oh this is just because the 0.2.19 release didn't have a windows release due to a bug in wasm-bindgen's configuration, but 0.2.21 has it fixed. This shouldn't need to change again.

ashleygwilliams: 0.2.19 hit a bug hence the number update. this won't need to be maintained. resolved.

ashleygwilliams:
ping @alexcrichton looks like this is still failing on appveyor - lemme know if i can help!

Issue #288, opened 2018-09-07
Colors in rustc's error mesages are lost


alexcrichton:
## 🐛 Bug description

When compiling code with `wasm-pack build` the compiler's error messages are not printed with colors.

#### 🤔 Expected Behavior

The compiler's colors come through!

#### 👟 Steps to reproduce

If you create a blank `src/lib.rs` with only one token in it, `x` (invalid rust code), then `wasm-pack build` will show the compiler's error but without colors

ashleygwilliams:
this was solved by the PR that moved us toward our new output strategy :) closing.

Issue #287, opened 2018-09-07
Show the output of Cargo's build


alexcrichton:
## 💡 Feature description

Compiling a crate can often take quite a long time and it's nice to know about the progress within the build when checking up on it. Would it be possible to avoid suppressing the output of `cargo build` and have it stream to the console?

ashleygwilliams:
i have also wanted this! i think it's possible, and i'd want it to be default, *i think*, though i do think we might want to give a flag option to turn it off. any one else have thoughts?

fitzgen:
Personally, I'd prefer always displaying all output by default, but having a `-q` flag to `wasm-pack` which makes it not display subprocess's outputs.

It would be neat if we prefixed the output with the command that was run, eg:

```text
cargo build (stderr): a warning message or something...
cargo build (stderr): some more warnings or whatever...
cargo build (stdout): something from stdout
[4/8] Compiling to wasm...
```

PR #293, opened 2018-09-08
Always ‘color’ with cargo build.

jtomschroeder:
Always color cargo's output - #288

### DIFF
--- a/src/build.rs
+++ b/src/build.rs
@@ -57,6 +57,7 @@ pub fn cargo_build_wasm(path: &Path, debug: bool, step: &Step) -> Result<(), Err
             cmd.arg("--release");
         }
         cmd.arg("--target").arg("wasm32-unknown-unknown");
+        cmd.arg("--color").arg("always");
         cmd.output()?
     };
 


alexcrichton:
Thanks for this! Unfortunately though this won't fix the issue for a number of use cases :(

While this will force cargo/rustc to turn on colors, because the output is still captured via a pipe it's always colored using ANSI escape codes. That means that Windows console users won't get colors but rather seemingly garbled text (as the escape codes are printed). Additionally on Unix if wasm-pack's output is piped to a file then colors should also be disabled.

This bug may be solved as well if https://github.com/rustwasm/wasm-pack/issues/287 is implemented

jtomschroeder:
Oh yeah, that makes sense. #287 would be great.

ashleygwilliams:
@alexcrichton does this solve enough to be worth merging (i understand it won't solve everything, but how good of an improvement is it?)

alexcrichton:
I would personally think no because this makes the compiler experience worse on Windows by accident

PR #299, opened 2018-09-11
feat(publish): add --access flag to publish command

### LINKED ISSUE #297: support --access flag to publish command
## 🐛 Bug description
Scoped packages default to private publishing and we currently do not expose any `--access` flag. As a result folks who want to make a scoped package can only publish as a private package, which requires a paid account.

ashleygwilliams:
fixes #297

### DIFF
--- a/src/command/mod.rs
+++ b/src/command/mod.rs
@@ -3,14 +3,15 @@
 pub mod build;
 mod login;
 mod pack;
-mod publish;
+/// Data structures and functions for publishing a package.
+pub mod publish;
 pub mod test;
 pub mod utils;
 
 use self::build::{Build, BuildOptions};
 use self::login::login;
 use self::pack::pack;
-use self::publish::publish;
+use self::publish::{access::Access, publish};
 use self::test::{Test, TestOptions};
 use error::Error;
 use slog::Logger;
@@ -36,6 +37,10 @@ pub enum Command {
     #[structopt(name = "publish")]
     /// 🎆  pack up your npm package and publish!
     Publish {
+        /// The access level for the package to be published
+        #[structopt(long = "access", short = "a")]
+        access: Option<Access>,
+
         /// The path to the Rust crate.
         #[structopt(parse(from_os_str))]
         path: Option<PathBuf>,
@@ -92,10 +97,10 @@ pub fn run_wasm_pack(command: Command, log: &Logger) -> result::Result<(), Error
             info!(&log, "Path: {:?}", &path);
             pack(path, &log)
         }
-        Command::Publish { path } => {
+        Command::Publish { path, access } => {
             info!(&log, "Running publish command...");
             info!(&log, "Path: {:?}", &path);
-            publish(path, &log)
+            publish(path, access, &log)
         }
         Command::Login {
             registry,
--- a/src/npm.rs
+++ b/src/npm.rs
@@ -1,5 +1,6 @@
 //! Functionality related to publishing to npm.
 
+use command::publish::access::Access;
 use error::Error;
 use std::process::{Command, Stdio};
 
@@ -18,13 +19,23 @@ pub fn npm_pack(path: &str) -> Result<(), Error> {
 }
 
 /// Run the `npm publish` command.
-pub fn npm_publish(path: &str) -> Result<(), Error> {
-    let output = Command::new("npm")
-        .current_dir(path)
-        .arg("publish")
-        .stdin(Stdio::inherit())
-        .stdout(Stdio::inherit())
-        .output()?;
+pub fn npm_publish(path: &str, access: Option<Access>) -> Result<(), Error> {
+    let output = match access {
+        Some(a) => Command::new("npm")
+            .current_dir(path)
+            .arg("publish")
+            .arg(&format!("{}", a.to_string()))
+            .stdin(Stdio::inherit())
+            .stdout(Stdio::inherit())
+            .output()?,
+        None => Command::new("npm")
+            .current_dir(path)
+            .arg("publish")
+            .stdin(Stdio::inherit())
+            .stdout(Stdio::inherit())
+            .output()?,
+    };
+
     if !output.status.success() {
         let s = String::from_utf8_lossy(&output.stderr);
         Error::cli("Publishing to npm failed", s)
--- a/src/command/publish/access.rs
+++ b/src/command/publish/access.rs
@@ -1,0 +1,35 @@
+use error::Error;
+use std::fmt;
+use std::str::FromStr;
+
+/// Represents access level for the to-be publish package. Passed to `wasm-pack publish` as a flag, e.g. `--access=public`.
+#[derive(Debug)]
+pub enum Access {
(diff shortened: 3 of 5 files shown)


src/command/publish.rs
```diff
@@ -1,15 +1,47 @@
...
+}
+
+impl FromStr for Access {
+    type Err = Error;
+
+    fn from_str(s: &str) -> Result<Self, Error> {
+        return match s {
+      "public" => Ok(Access::Public),
+      "restricted" => Ok(Access::Restricted),
+      "private" => Ok(Access::Restricted),
```
ashleygwilliams: doing something a little unconventional here but i seriously cannot believe that it's "restricted" and not "private" so i kiiiiiinda want to have this in here, but if ya'll object i'm cool with removing it

src/command/publish.rs
```diff
@@ -1,15 +1,47 @@
...
+      _ => Err(Error::Unsupported { message: format!("{} is not a supported access level. See https://docs.npmjs.com/cli/access for more information on npm package access levels.", s)}),
+    };
+    }
```
ashleygwilliams: also really not sure what's up with the formating here, i ran cargo fmt but this is nasty

src/command/publish.rs
```diff
@@ -1,15 +1,47 @@
...
+}
+
+impl FromStr for Access {
+    type Err = Error;
+
+    fn from_str(s: &str) -> Result<Self, Error> {
+        match s {
+      "public" => Ok(Access::Public),
+      "restricted" => Ok(Access::Restricted),
+      "private" => Ok(Access::Restricted),
```
ashleygwilliams: doing something a little unconventional here but i seriously cannot believe that it's "restricted" and not "private" so i kiiiiiinda want to have this in here, but if ya'll object i'm cool with removing it

src/command/publish.rs
```diff
@@ -1,15 +1,47 @@
...
+      _ => Err(Error::Unsupported { message: format!("{} is not a supported access level. See https://docs.npmjs.com/cli/access for more information on npm package access levels.", s)}),
+    }
```
ashleygwilliams: also really not sure what's up with the formating here, i ran cargo fmt but this is nasty

ashleygwilliams:
i actually just realized that if you pass `--access=""` npm gets mad. i was hoping it was just do it's default. do we want to change how the access level defaults? if yes, i can make a small change, if no i can make a bigger one. i'm thinking probably no, so this will require a small amount of work before it can be merged.

fitzgen:
> do we want to change how the access level defaults? if yes, i can make a small change, if no i can make a bigger one. i'm thinking probably no, so this will require a small amount of work before it can be merged.

While I personally think that access should default to public, I don't think it is worth creating inconsistencies with `npm` over.

src/command/mod.rs
```diff
@@ -35,6 +35,10 @@ pub enum Command {
     /// 🎆  pack up your npm package and publish!
     Publish {
         /// The path to the Rust crate.
+        #[structopt(long = "access", short = "a")]
+        access: Option<Access>,
+
+        /// The access level for the package to be published
```
fitzgen: This comment should be on the field above, and the comment for the field above should be here :-p

src/command/publish.rs
fitzgen: So the npm wording is "restricted" but this is creating an alias to that named "private"? I'd prefer just matching what npm does.

fitzgen: The indentation here is super funky too

src/command/publish.rs
fitzgen: Maybe if you manually add some `\` followed by a newline in the middle of the format string?

src/command/publish.rs
```diff
@@ -1,15 +1,47 @@
...
+    log: &Logger,
+) -> result::Result<(), Error> {
     let crate_path = set_crate_path(path);
+    let access = match access {
+        Some(access) => match access {
+            Public => "public",
+            Restricted => "restricted",
+        },
+        None => "",
+    };
```
fitzgen: How about instead of making this stringly typed again, we implement `Display` for `Access`, implement `Default` for `Access`, and then right here do

```rust
access.unwrap_or_default()
```

src/npm.rs
```diff
@@ -18,10 +18,11 @@ pub fn npm_pack(path: &str) -> Result<(), Error> {
 }
 
 /// Run the `npm publish` command.
-pub fn npm_publish(path: &str) -> Result<(), Error> {
+pub fn npm_publish(path: &str, access: &str) -> Result<(), Error> {
```
fitzgen: And then this argument would be `access: Access` and we know that we can't ever pass a bad access here

src/command/publish.rs
mgattozzi: I like the alias personally. wow cargo fmt what is you doing.

src/command/publish.rs
mgattozzi: I agree with the use of `Display` to not have to worry about strings and have it be centralized to one spot so that calls to `to_string()` will be consistent as well

ashleygwilliams: so this actually isn't stringly typed- or it is, only insofar as struct opt needs to read strings from the CLI. once it's read from the CLI, it's treated as the enum, and if anything is passed that doesn't work, it's rejected. we can be sure that we never pass npm publish a bad string. that being said i can refactor this a bit, but just wanted to be clear it's not stringly typed. additionally, assuming we don't want to override npm defaults, we can't actually set a default for access because it will be based on whether the package is scoped which we don't track (and is very difficult to track since the package.json can be edited at anytime between init/build and publish)...

ashleygwilliams: (i do like the idea of implement Display for it tho, i just don't think that removes having to implement the from string because of struct opt)

src/command/publish.rs
ashleygwilliams: i talked with nick and am gonna file an issue, it's a bug in cargo fmt, likely from the long string literal

src/command/mod.rs
ashleygwilliams: d'oh. good catch.

src/command/publish.rs
ashleygwilliams: the one thing to note is that we can't implement a default- this is because the default changes depending on whether or not the package is scoped, which we don't currently track. i need to think about whether i think it's a good idea to do so, but i had a plan for how to do the default that didn't involve tracking whether it was scoped so i may try that first.

ashleygwilliams:
failing re doc comments now- gonna fill those out but i think i have one more refactor to make this nicer. it's too bad that the npm access flag doesn't take a default, e.g. `--access=default`, but i have a plan, that also cleans up the duplication in `src/npm.rs`

src/command/mod.rs
fitzgen: This is still unaddressed ;)

mgattozzi:
This LGTM great job! Let's get it merged

PR #301, opened 2018-09-12
Remove `branch: master` from AppVeyor deploy

### LINKED ISSUE #147: debug build windows appveyor releases
we currently make release binaries for linux and mac using travis. we should also do this for windows with appveyor!

alexcrichton:
fixes #147 

It looks like Windows releases aren't getting build for wasm-pack tags
unfortunately and taking a look at the [last logs][logs] it looks like the
reason is that the `master` branch is required. I'm not actually entirely sure
what's required here but by removing this it's closer to `wasm-bindgen`'s
configuration which I think is working!

[logs]: https://ci.appveyor.com/project/ashleygwilliams/wasm-pack-071k0/build/1.0.72

### DIFF
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -12,7 +12,7 @@ install:
 build: false
 
 test_script:
-  - cargo test --release --locked -- --test-threads 1
+  - cargo test --release --locked
 
 before_deploy:
   - ps: |
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -31,5 +31,4 @@ deploy:
   autho_token:
     secure: iHsRUqwGf/Zh7OuYpHOWQL8buaOL+c8/6kXLRly8V2j0LCUo7CcDs0NxQ0vl2bhZ
   on:
-    branch: master
     appveyor_repo_tag: true
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -8,7 +8,7 @@ install:
 build: false
 
 test_script:
-  - cargo test --locked -- --test-threads 1
+  - cargo test --release --locked -- --test-threads 1
 
 before_deploy:
   - ps: |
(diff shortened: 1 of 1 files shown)


alexcrichton:
Er actually realized that the deploy wasn't actually getting its release artifacts built as well so pushed up a commit for that as well as making the binaries a bit more compatible with more windows versions.

.appveyor.yml
```diff
@@ -8,7 +12,7 @@ install:
 build: false
 
 test_script:
-  - cargo test --locked -- --test-threads 1
+  - cargo test --release --locked -- --test-threads 1
```
ashleygwilliams: i think we can also remove the test-threads req here now

ashleygwilliams: should we add `--release` to `.travis.yml` as well? https://github.com/rustwasm/wasm-pack/blob/master/.travis.yml#L39

.appveyor.yml
```diff
@@ -1,3 +1,7 @@
+environment:
+  global:
+    RUSTFLAGS: -C target-feature=+crt-static
```
ashleygwilliams: TIL!

.appveyor.yml
alexcrichton: Nah that's fine to be non-release as it's not producing a final binary. AppVeyor right now doesn't have separate jobs (like Travis does) for building releases and testing, and from my experience at least that tends to work out alright.

In that sense this is just needed here for AppVeyor because we're both testing a binary and publishing it, whereas Travis either publishes it or tests it. Testing without `--release` is generally preferred for Travis because it gets you results more quickly

alexcrichton: Also certainly, can remove `--test-threads`!

ashleygwilliams:
🎉

PR #307, opened 2018-09-14
Add an experimental installer for wasm-pack

### LINKED ISSUE #295: idea: "Installers" for `wasm-pack`
Currently one of the main installations methods for `wasm-pack` is `cargo install`, and this has a few great benefits to it:

* Users don't have to know what platform they're running on, they just need `cargo` which is already pretty easy to install.
* Users don't need to set up `PATH` to get `wasm-pack`-the-subcommand because Cargo already takes care of those warnings.
* It's a pretty seamless experience where everything is downloaded for you.

The downsides, however, of `cargo install` are:

* Compilation of Rust code isn't always speedy, so it can take some time. Furthermore as `wasm-pack` continues to grow in functionality it'll pick up dependencies and slow down to compile.
* Over time `wasm-pack` may pick up hard-to-compile dependencies. For example it recently picked up `curl` as a crate, but that needs to compile C code and deals with linkage of C libraries and such. That can often trip up a lot of newcomers who don't always have pristinely configured environments.

I think, though, that we may be able to solve these downsides while still preserving the benefits of `cargo install` today. We could employ a rustup-like solution to install `wasm-pack` but leave automatic updates for a future iteration. We've already got precompiled binaries (that are "quite compatible") for wasm-pack, so we just need a better way to get them to users.

I'm imagining something like:

* There's a site like https://rustup.rs/ (hosted on gh-pages most likely) which users visit to install `wasm-pack`. This website does platform detection to tell users what installer to install.
* On Unix we'd provide a `curl | sh` solution with a shell script in this repository (similar to `rustup.sh`). I don't think this'd be too hard to implement, as it'd largely just download a tarball from github (the precompiled release) and then extract the `wasm-pack` binary into `$HOME/.cargo/bin`.
* On Windows we'd provide `wasm-pack-init.exe` (or something like that), which could actually simply be `wasm-pack.exe` itself. When run for the first time `wasm-pack.exe` would move itself into `.cargo/bin` as well, which should already be in `PATH`.

Overall this would free us from the above two downsides (long build times and sometimes complicated native dependencies) while also keeping wasm-pack as easy to install as it is today. One perhaps snag in this plan is that updates are a bit easier with Cargo ("run this command" vs "visit this website"), but updates with `cargo install` already aren't that great so it's perhaps not that bad!

alexcrichton:
This commit kicks off the addition of installers for wasm-pack, with the main
goals of being:

* Users should have a one-click (ideally) solution to download and install
  wasm-pack.
* We should ideally not have to worry about picking up "heavy" dependencies in
  wasm-pack like C++ or C code (if necessary).

The general idea here is very similar to rustup (and in fact a good deal of code
is copied from them!). The installation worklow for wasm-pack in theory after
this commit looks like:

1. First, a users visits the installer landing page. A preview is available at
   https://alexcrichton.github.io/wasm-pack/installer/. Clearly this page needs
   a better stylesheet!

2. The user performs the instructions presented on the page. The website
   automatically detects what platform you're running on, basically giving you a
   curl/sh script or a Windows installer. There's various options for seeing
   other installers as well, and the fallback of `cargo install` is presented if
   we don't recognize the platform (or if we don't think we have precompiled
   binaries).

3a. On Unix, users execute a curl/sh script like:

    ```
    curl https://alexcrichton.github.io/wasm-pack/installer/init.sh -sSf | sh
    ```

    This command will download the shell script included in this PR, which will
    in turn do some further platform detection (like OSX vs Linux) and then
    download the most recent version of `wasm-pack` from GitHub releases. This
    is then extracted, the `wasm-pack` binary is renamed to `wasm-pack-init`,
    and then that binary is run. The binary will refuse by default to overwrite
    a previous installation, but that can be overridden with the `-f` flag.

3b. On Windows users download a binary. This binary, when run, will pop up a
    console window. This window will have the same output as the shell script
    installer, and will wait at the end of its execution to ensure the user has
    time to read everything.

And... that's it! The deployment process for all this looks like so:

* All CI builds will update the website and shell script (published to
  gh-pages). A small script at `docs/installer/build-installer.rs` fills in the
  current version number inferred from `Cargo.toml`.

* Tagged CI builds will publish releases, as usual.

* "Pushing a release" is done by bumping the version number in `Cargo.toml`.
  When bumped all online installers will now point to the new release. There
  will be a window of time, though, when the version number is bumped and the
  release hasn't finished building on CI. In this case users will get errors
  (unfortunately).

This is all still a work-in-progress and feedback is definitely appreicated and
welcome on all this!

Closes #295


ashleygwilliams:
hey @alexcrichton this is awesome- quick question before i review- could you move the installer to the rustwasm org? i am very happy to improve the stylesheet :)

.travis.yml
```diff
@@ -59,18 +59,20 @@ matrix:
     - (test -x $HOME/.cargo/bin/mdbook || cargo install --vers "^0.1" mdbook)
     - cargo install-update -a
     script:
-    - cd docs && mdbook build
+    - (cd docs && mdbook build)
```
ashleygwilliams: what do the parans do here? am curious

docs/installer/build-installer.rs
```diff
@@ -0,0 +1,30 @@
+use std::fs;
```
ashleygwilliams: curious why this is in the docs dir? (not necessarily opposed just not entirely clear why)

docs/installer/index.html
```diff
@@ -0,0 +1,106 @@
+<!DOCTYPE html>
```
ashleygwilliams: if we land this i would love to do something where it's not just an installer page but actually just the landing page for wasmpack, e.g. links to the docs, shows an example... no need to actually change this but thought i'd mention

src/main.rs
```diff
@@ -1,31 +1,43 @@
...
 #[macro_use]
 extern crate human_panic;
 extern crate structopt;
 extern crate wasm_pack;
+extern crate which;
 
-use failure::Fail;
 use std::env;
 use structopt::StructOpt;
-use wasm_pack::{command::run_wasm_pack, error::Error, logger, Cli};
```
ashleygwilliams: curious about what you are doing with Error here... can you explain?

.travis.yml
alexcrichton: Ah this is because travis generates a literal shell script and without the parens the `cd` changes the directory for the rest of the script, so by using parens here the cwd is changed for just this one step

docs/installer/build-installer.rs
alexcrichton: Ah just threw it in here as it was deployed to gh-pages, would be fine to have it anywhere!

docs/installer/index.html
alexcrichton: 👍

PR #312, opened 2018-09-18
modules modules modules

### LINKED ISSUE #309: use `module` key for ES6 generated assets
and add "sideEffects": "false" so tree shaking can happen

ashleygwilliams:
fixes #309 plus rather large refactor which is in the first few commits

strategy here is to define several struct for types of packages folks will want to make (common js only, es6 only, es6 + commonjs to start), we'll serialize the Cargo.toml into the format based on what the target specified is

### DIFF
--- a/src/manifest/mod.rs
+++ b/src/manifest/mod.rs
@@ -101,33 +101,22 @@ impl CargoManifest {
         let filename = self.package.name.replace("-", "_");
         let wasm_file = format!("{}_bg.wasm", filename);
         let js_file = format!("{}.js", filename);
+        let mut files = vec![wasm_file];
 
-        let dts_file = if disable_dts == true {
-            None
-        } else {
-            Some(format!("{}.d.ts", filename))
-        };
-
-        let js_bg_file = Some(format!("{}_bg.js", filename));
+        let js_bg_file = format!("{}_bg.js", filename);
+        files.push(js_bg_file.to_string());
 
         if let Some(s) = scope {
             self.package.name = format!("@{}/{}", s, self.package.name);
         }
-        let mut files = vec![wasm_file];
 
-        match dts_file {
-            Some(ref dts_file) => {
-                files.push(dts_file.to_string());
-            }
-            None => {}
-        }
-
-        match js_bg_file {
-            Some(ref js_bg_file) => {
-                files.push(js_bg_file.to_string());
-            }
-            None => {}
-        }
+        let dts_file = if disable_dts == false {
+            let file = format!("{}.d.ts", filename);
+            files.push(file.to_string());
+            Some(file)
+        } else {
+            None
+        };
 
         check_optional_fields(
             &self.package.description,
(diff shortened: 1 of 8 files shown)


src/manifest/mod.rs
```diff
@@ -169,7 +142,46 @@ impl CargoManifest {
...
+        let mut files = vec![wasm_file];
+
+        match dts_file {
+            Some(ref dts_file) => {
+                files.push(dts_file.to_string());
+            }
+            None => {}
+        }
+
+        NpmPackage::ES6Package(ES6Package {
```
fitzgen: nitpick: I'd probably name this `EsModulePackage` since I think modules didn't come until after es6 and es6 is also actually es2015 or some such... or something like that, right?

ashleygwilliams: yeah that actually makes a bunch of sense, good catch

src/manifest/mod.rs
```diff
@@ -169,7 +148,53 @@ impl CargoManifest {
...
+        let dts_file = if disable_dts == true {
+            None
+        } else {
+            Some(format!("{}.d.ts", filename))
+        };
+
+        if let Some(s) = scope {
+            self.package.name = format!("@{}/{}", s, self.package.name);
+        }
+        let mut files = vec![wasm_file, js_file.clone()];
```
ashleygwilliams: this `clone`. i hate it but i also love not having a ton of type annontations everywhere. i think it's fine but if we hate it let's talk about ways to make it as clean as possible.

fitzgen: I don't think this single allocation that isn't in a loop (or even within any code that we know to be hot) is a problem. :+1:

src/manifest/mod.rs
```diff
@@ -169,7 +148,53 @@ impl CargoManifest {
...
+        let wasm_file = format!("{}_bg.wasm", filename);
+        let js_file = format!("{}.js", filename);
+
+        let dts_file = if disable_dts == true {
+            None
+        } else {
+            Some(format!("{}.d.ts", filename))
+        };
+
+        if let Some(s) = scope {
```
alexcrichton: This scope handling logic looks duplicated between this and the above?

src/manifest/mod.rs
```diff
@@ -169,7 +148,53 @@ impl CargoManifest {
...
+
+    fn into_esmodules(mut self, scope: &Option<String>, disable_dts: bool) -> NpmPackage {
+        let filename = self.package.name.replace("-", "_");
+        let wasm_file = format!("{}_bg.wasm", filename);
+        let js_file = format!("{}.js", filename);
+
+        let dts_file = if disable_dts == true {
+            None
+        } else {
+            Some(format!("{}.d.ts", filename))
```
alexcrichton: This dts logic looks duplicated with the above?

src/manifest/mod.rs
ashleygwilliams: this is all from master but yeah you are right i think, file an issue?

src/manifest/mod.rs
ashleygwilliams: this is all from master but yeah you are right i think, file an issue?

ashleygwilliams:
@alexcrichton correct, this is just to lay the groundwork for #313

src/manifest/mod.rs
ashleygwilliams: i guess i can try to fix it in this PR, might as well

mgattozzi:
Great job @ashleygwilliams!

PR #320, opened 2018-09-20
chore(deps): cargo update

ashleygwilliams:


ashleygwilliams:
@alexcrichton in theory this should also fix #284 - would you be willing to give it a test run and let me know?

PR #323, opened 2018-09-20
tests: Make tests run faster by using Fixture builders

### LINKED ISSUE #310: investigate tests running incredibly slow on mac os

fitzgen:
fixes #310 

Instead of copying directories, just build the fixtures directly in the temporary directory.

-------------------------

```bash
$ rustup override set nightly
```


tests/all/utils/fixture.rs
```diff
@@ -17,66 +17,123 @@ pub struct Fixture {
...
-    );
+impl Fixture {
+    /// Create a new test fixture in a temporary directory.
+    pub fn new() -> Fixture {
+        // Make sure that all fixtures end up sharing a target dir, and we don't
+        // recompile wasm-bindgen and friends many times over.
+        static SET_TARGET_DIR: Once = ONCE_INIT;
+        SET_TARGET_DIR.call_once(|| {
+            env::set_var(
+                "CARGO_TARGET_DIR",
```
alexcrichton: I'd personally recommend just always setting this env var for subprocesses, but I think `Once` will probably do the trick as well. I'm always just sort of weirdly paranoid about changing env vars while running.

tests/all/utils/fixture.rs
```diff
@@ -87,14 +144,29 @@ impl Fixture {
...
+        assert!(shared_wasm_bindgen_test_runner.is_file());
+
+        wasm_pack::binaries::ensure_local_bin_dir(&self.path).unwrap();
+
+        fs::copy(
+            &shared_wasm_bindgen,
+            wasm_pack::binaries::local_bin_path(&self.path, "wasm-bindgen"),
+        ).expect("could not copy `wasm-bindgen` to fixture directory");
+
+        fs::copy(
```
alexcrichton: For an even speedier test I'd recommend using `hard_link` first and falling back to `copy` only if that fails

tests/all/utils/fixture.rs
fitzgen: FWIW, this part is just code motion and we were already doing this.

alexcrichton:
Seems fine by me!

I personally like having the fixtures next to the test itself (so you don't have to jump around too much), but that may just be me and also wouldn't want to cause too much churn for this!

tests/all/utils/fixture.rs
alexcrichton: 👍 no worries, feel free to move on

fitzgen: (Also, the actual subprocess spawning happens deep within wasm-pack, not in the tests :-/ I agree with the general principal, however!)

tests/all/utils/fixture.rs
fitzgen: done!

Issue #270, opened 2018-08-25
Handle versions of wasm-bindgen in lockfiles


mgattozzi:
## 💡 Feature description
We have a basic implementation of keeping up to date with wasm-bindgen. Ideally though with the local install we'd tie the version to whatever is in the lock file to avoid issues with things like `Cargo.toml` using `wasm-bindgen = 0.2` as what version of `wasm-bindgen-cli` does this mean? Having a lock file look up would solve this issue.

fitzgen:
Finding the lock file can be interesting if the crate is in a workspace. We can leverage https://crates.io/crates/cargo_metadata to help.

data-pup:
Digging into this now :smile_cat: 

https://github.com/oli-obk/cargo_metadata/blob/master/src/lib.rs#L156

It does look like the `Cargo.toml` dependency is what is given, rather than the lockfile. This crate should still be useful for finding the lock file though :)

data-pup:
Sounds good to me :o) I'll follow up if I have other questions!

data-pup:
I think I have a lot of the work on this done :smile_cat: I should be able to open a PR shortly.

As an aside, is there anywhere where the fields in `Cargo.lock` are officially specified? The closest I could find was [this](https://doc.rust-lang.org/cargo/guide/cargo-toml-vs-cargo-lock.html), in the Cargo book.

fitzgen:
I think @ashleygwilliams or @alexcrichton would know the answer to that question.

alexcrichton:
Ah they're not necessarily officially specified but when using TOML+Serde you can deserialize into a structure like:

```rust
#[derive(Deserialize)]
struct Lockfile {
    packages: Vec<Package>,
}

#[derive(Deserialize)]
struct Package {
    name: String,
    version: String,
    source: Option<String>,
}
```

and that should be enough to match on `wasm-bindgen`!

data-pup:
Yay! Thanks for the help @alexcrichton, that lines up with what I had done so far. My struct looked like this:

```rust
#[derive(Deserialize)]
struct Package {
    name: String,
    version: String,
    source: Option<String>,
    dependencies: Option<Vec<String>>
}
```

Should I keep the `dependencies` field out of this, following your example?

ashleygwilliams:
closed by the last release.

PR #302, opened 2018-09-12
(WIP) Use `Cargo.lock` to identify wasm-bindgen version.

data-pup:
Fixes #270. This aims to find the wasm-bindgen version using the lockfile.

Note: This is still a WIP, and should probably land after #271. While working on this, I ran into some of the same problems that were solved by that PR, so I tried to remove any duplicate work. I can rebase this once that lands, but review/feedback is welcome in the meantime :)

---

Make sure these boxes are checked! 📦✅

- [x] You have the latest version of `rustfmt` installed and have your 
      cloned directory set to nightly
```bash
$ rustup override set nightly
$ rustup component add rustfmt-preview --toolchain nightly
```
- [x] You ran `rustfmt` on the code base before submitting
- [x] You reference which issue is being closed in the PR text

✨✨ 😄 Thanks so much for contributing to wasm-pack! 😄 ✨✨

PR #324, opened 2018-09-20
feat(lockfiles): Use `Cargo.lock` to identify wasm-bindgen versions

fitzgen:
fixes https://github.com/rustwasm/wasm-pack/issues/270, #344 

This lets us leverage `cargo` for semver finding and then ensure that we get the exact same version of the CLI that cargo selected. It also lets us support fuzzy dependencies like "0.2" instead of exact dependencies like "0.2.21" again.

Additionally, this lets us remove the `_` vs `-` normalization, since `cargo` also handles that for us.

This is a updated version of #302. In addition to a rebase, this also removes the old `Cargo.toml` checking code, migrates callers to this new version, and ensures that we never check the wasm-bindgen version until after a `Cargo.lock` exists.

------------------------------------------

```bash
$ rustup override set nightly
```

### DIFF
--- a/src/command/test.rs
+++ b/src/command/test.rs
@@ -4,10 +4,11 @@ use super::build::BuildMode;
 use bindgen;
 use build;
 use command::utils::set_crate_path;
+use console::style;
 use emoji;
 use error::Error;
 use indicatif::HumanDuration;
-use lockfile;
+use lockfile::Lockfile;
 use manifest;
 use progressbar::Step;
 use slog::Logger;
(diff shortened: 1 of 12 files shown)


fitzgen:
Crap, I forgot to add these test cases:

> More cases we can handle now and should have tests for:
> 
> * [ ] running `wasm-pack build` in each crate within an umbrella workspace:
> 
> ```
> umbrella/
> - Cargo.toml                        # just have `[workspace] members = [..]`
> - foo/
> - Cargo.toml
> - src/
> - lib.rs
> - bar/
> - Cargo.toml
> - src/
> - lib.rs
> ```
> 
> * [ ] a dependency uses `wasm-bindgen`, but the root crate does not:
> 
> ```rust
> // child/src/lib.rs
> extern crate wasm_bindgen;
> use wasm_bindgen::prelude::*;
> 
> #[wasm_bindgen]
> pub fn hello() -> u32 { 42 }
> 
> // parent/src/lib.rs
> extern crate child;
> pub use child::*;
> ```

fitzgen:
Added the missing test cases.

src/command/test.rs
```diff
@@ -183,9 +184,9 @@ impl Test {
         }
         match self.mode {
             BuildMode::Normal => steps![
-                step_check_crate_config,
                 step_add_wasm_target,
                 step_build_tests,
+                step_check_crate_config,
```
fitzgen: Thinking about it some more, we might want to still check the crate config as the first thing we do, but have checking crate config *not* check wasm-bindgen versions. Then when we go to install wasm-bindgen, we look at the lockfile at tjhat time.

ashleygwilliams: yes i think that's a good call. crate config should def be one of the first things we do because it's a quick fix for folks who it fails for.

src/lockfile.rs
```diff
@@ -0,0 +1,91 @@
...
+    File::open(lock_path)?.read_to_string(&mut lockfile)?;
+    toml::from_str(&lockfile).map_err(Error::from)
+}
+
+/// Given the path to the crate that we are buliding, return a `PathBuf`
+/// containing the location of the lock file, by finding the workspace root.
+fn get_lockfile_path(crate_path: &Path) -> Result<PathBuf, Error> {
+    // Identify the crate's root directory, or return an error.
+    let manifest = crate_path.join("Cargo.toml");
+    let crate_root = cargo_metadata::metadata(Some(&manifest))
```
alexcrichton: This is a bit of a weighty step in terms of execution time, and I think it's executed once for wasm-bindgen and once for wasm-bindgen-test? Could the result be cached per crate?

ashleygwilliams:
closing in favor of https://github.com/rustwasm/wasm-pack/pull/365.  hopefully for the last time <3<3

PR #327, opened 2018-09-21
support no-modules

### LINKED ISSUE #317: `--no-modules` for wasm-pack
## 💡 Feature description
`--no-modules` comes with `wasm-bindgen-cli` and it's needed for some bundlers.

ashleygwilliams:
fixes #317 

note: 

this PR puts the js file generated by bindgen in the `browser` key in `package.json` and includes it in `files`. i am not sure this is the right call- so if folks could weigh in that would be excellent.

### DIFF
--- a/src/bindgen.rs
+++ b/src/bindgen.rs
@@ -136,6 +136,7 @@ pub fn wasm_bindgen_build(
         };
         let target_arg = match target {
             "nodejs" => "--nodejs",
+            "no-modules" => "--no-modules",
             _ => "--browser",
         };
         let bindgen_path = Path::new(&wasm_bindgen_path);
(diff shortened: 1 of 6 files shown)


src/bindgen.rs
```diff
@@ -136,6 +136,7 @@ pub fn wasm_bindgen_build(
         };
         let target_arg = match target {
             "nodejs" => "--nodejs",
+            "no-modules" => "--no-modules",
```
fitzgen: Not in this PR, but as a follow up issue, this seems like a good candidate for an `enum` with `FromStr` and `Display` impls like how we do with `BuildMode` and `Access`

PR #336, opened 2018-09-22
Pass `--debug` to `wasm-bindgen` in debug mode

alexcrichton:
In debug mode `wasm-bindgen` emits more JS glue to check more invariants
and otherwise provide more sanity checks. This is omitted by default
with wasm-pack as it runs in release mode by default, but if `--debug`
is passed let's be sure to include it!


alexcrichton:
Uh... I think that's because the test suite is broken? It's picking up a newer 0.2.22 version from crates.io but still installing 0.2.21 binaries, and the lock file isn't being read still so `=0.2.21` doesn't work...

Issue #339, opened 2018-09-23
The cdylib's name can be overriden by lib.name 


konstin:
## 🐛 Bug description

Currently, wasm-pack assumes the cdylib's name is the crate name with underscores and `.wasm` as extension, while this value can be overwritten by `lib.name` in Cargo.toml.

#### 👟 Steps to reproduce

`Cargo.toml`

```toml
[package]
name = "foo"
version = "0.1.0"
authors = ["konstin <konstin@mailbox.org>"]

[lib]
crate-type = ["cdylib"]
name = "bar" # <- "Renames" the artifact to bar.wasm

[dependencies]
wasm-bindgen = "0.2.22"
```

`src/lib.rs`

```rust
#[cfg(test)]
mod tests {
    #[test]
    fn it_works() {
        assert_eq!(2 + 2, 4);
    }
}
```

Output of `wasm-pack build --mode no-install --debug`

```
  [1/6] Checking crate configuration...
  [2/6] Compiling to WASM...
  [3/6] Creating a pkg directory...
  [4/6] Writing a package.json...
  :-) [WARN]: Field 'description' is missing from Cargo.toml. It is not necessary, but recommended
  :-) [WARN]: Field 'repository' is missing from Cargo.toml. It is not necessary, but recommended
  :-) [WARN]: Field 'license' is missing from Cargo.toml. It is not necessary, but recommended
  [5/6] Copying over your README...
  :-) [WARN]: origin crate has no README
| [6/6] Running WASM-bindgen...
wasm-bindgen failed to execute properly. stderr:

error: failed to read `target/wasm32-unknown-unknown/debug/foo.wasm`
	caused by: No such file or directory (os error 2)
```

#### :desktop_computer: Relevant code

https://github.com/rustwasm/wasm-pack/blob/120236f1a93e5e96b9b7692e9630aad93fcdd70d/src/manifest/mod.rs#L264-L267

And [wasm_bindgen_build](https://github.com/rustwasm/wasm-pack/blob/64c5a22ff6f20a9b20790b50480dceb1c5e1d28d/src/bindgen.rs#L109-L160)

#### 🤔 Expected Behavior

While checking for `lib.name` would fix this particular issue, it's definitely not future proof and I assume there are already cases today where this still wouldn't work.

One solution would be asking cargo through `cargo metadata`. Relevant excerpt:

```json
{
  "packages": [
    {
      "name": "c",
      "version": "0.1.0",
[...]

PR #340, opened 2018-09-23
[WIP] Use `--message-format=json` to fix #333

### LINKED ISSUE #288: Colors in rustc's error mesages are lost

### LINKED ISSUE #305: Ensure that running `wasm-pack build` in two crates in the same workspace works
Follow up to https://github.com/rustwasm/wasm-pack/issues/252

We want test coverage that ensure that if we have

```
workspace/
  |- crate_a/
  `- crate_b/
```

running `wasm-pack build` in `crate_a` works, and running `wasm-pack build` in `crate_b` right afterwards also works.

### LINKED ISSUE #333: Use `--message-format=json` for getting the location of the generated files
## 💡 Feature description

Currently, wasm-pack tries to guess the location of the asset from the Cargo.toml. With  `--message-format=json` we can get the location directly from cargo. 

This should also fix #252.

#### 💻 Basic example

```json
{
  "features": [],
  "filenames": [
    "/home/konsti/pyo3-pack/points/target/debug/libpoints.so"
  ],
  "fresh": true,
  "package_id": "points 0.1.0 (path+file:///home/konsti/pyo3-pack/points)",
  "profile": {
    "debug_assertions": true,
    "debuginfo": 2,
    "opt_level": "0",
    "overflow_checks": true,
    "test": false
  },
  "reason": "compiler-artifact",
  "target": {
    "crate_types": [
      "cdylib"
    ],
    "edition": "2015",
    "kind": [
      "cdylib"
    ],
    "name": "points",
    "src_path": "/home/konsti/pyo3-pack/points/src/lib.rs"
  }
}
```

Implementation in pyo3-pack: https://github.com/PyO3/pyo3-pack/blob/2f0c8a926bf1b607c48589a2393604de322f7f40/src/compile.rs#L221-L232

cargo-metadata has a [pending pull request](https://github.com/oli-obk/cargo_metadata/pull/53) which wraps cargo's messages in a nice api we could use.

konstin:
Parse the `--message-format=json` output with cargo_metadata to get the location of the wasm file from cargo.

This PR is marked WIP because it depends on https://github.com/oli-obk/cargo_metadata/pull/53. 7484faf1fc878d0883c8c05fe54e75a4cf265081 is because I otherwise had test failures about wrong versions (same as for @alexcrichton in https://github.com/rustwasm/wasm-pack/pull/336#issuecomment-423716747). With that commit `it_can_run_browser_tests` is still failing though. I guess this will be fixed before merging and I can rebase the commit away.

Adding `wasm_file` to `Build` is a bit of an ugly solution, but I couldn't come up with something better without rewriting the whole steps code. Currently I'm also ignoring the diagnostics messages and instead return the captured stderr in case of an error. (Not capturing stderr and setting `-q` ~~would fix #288 and with the other messages we~~ could easily do #287.  We'd only need to make sure how this interacts with indicatif)

      cloned directory set to nightly (well, 2018-09-16)

Closes #333
Fixes #305 
Relevant for #339 

### DIFF
--- a/src/build.rs
+++ b/src/build.rs
@@ -1,11 +1,17 @@
 //! Building a Rust crate into a `.wasm` binary.
 
+use cargo_metadata;
+use cargo_metadata::Message;
 use emoji;
 use error::Error;
 use progressbar::Step;
+use std::io::Read;
 use std::path::Path;
+use std::path::PathBuf;
 use std::process::Command;
+use std::process::Stdio;
 use std::str;
+use tempfile::NamedTempFile;
 use PBAR;
 
 /// Ensure that `rustc` is present and that it is >= 1.30.0
@@ -16,20 +22,20 @@ pub fn check_rustc_version(step: &Step) -> Result<String, Error> {
     match local_minor_version {
         Some(mv) => {
             if mv < 30 {
-              return Err(Error::RustcVersion {
-                message: format!(
-                  "Your version of Rust, '1.{}', is not supported. Please install Rust version 1.30.0 or higher.",
-                  mv.to_string()
-                ),
-                local_minor_version: mv.to_string(),
-              })
+                return Err(Error::RustcVersion {
+                    message: format!(
+                        "Your version of Rust, '1.{}', is not supported. Please install Rust version 1.30.0 or higher.",
+                        mv.to_string()
+                    ),
+                    local_minor_version: mv.to_string(),
+                });
             } else {
-              Ok(mv.to_string())
+                Ok(mv.to_string())
             }
-      },
-      None => Err(Error::RustcMissing {
-        message: "We can't figure out what your Rust version is- which means you might not have Rust installed. Please install Rust version 1.30.0 or higher.".to_string(),
-      }),
+        }
+        None => Err(Error::RustcMissing {
+            message: "We can't figure out what your Rust version is- which means you might not have Rust installed. Please install Rust version 1.30.0 or higher.".to_string(),
+        }),
     }
 }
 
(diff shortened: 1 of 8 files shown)


src/build.rs
```diff
@@ -47,27 +52,76 @@ fn ensure_nightly() -> Result<(), Error> {
...
-        let s = String::from_utf8_lossy(&output.stderr);
-        Error::cli("Compilation of your program failed", s)
+    let mut cmd = Command::new("cargo");
+    cmd.current_dir(path)
+        .arg("+nightly")
+        .arg("build")
+        .arg("--lib")
+        .arg("--target=wasm32-unknown-unknown")
+        .arg("--message-format=json")
+        .stderr(Stdio::piped())
```
alexcrichton: For this I believe it will need to pessimistically spawn a new thread to read stderr. Othrewise the blocking reads on stdout could cause wasm-pack to deadlock when it's waiting for info on stdout but the child is waiting to write info to stderr.

konstin: Afaik we're spawning a new process here and stderr is new pipe from the os, so the child can write to stderr no matter what the parent process does (except broken pipe errors and the like). Not an expert though.

alexcrichton: Pipes have a fixed capacity, however, and won't infinitely buffer. If Cargo dumps a lot of information on stderr it'll eventually block, and then it won't make progress until the pipe is read, which wasm-pack doesn't do until the process has otherwise finished.

konstin: Ah thanks, I didn't think about that! I've replaced the pipe with a tempfile.

konstin:
I've rebased onto master. I've realized that rustc and therefore cargo can't do both `human` and `json` messages at the same time, so we need to get the human messages out of the json. 

Because both `PBAR` and `Logger` errors aren't shown to the user, I simply eprintln'd the compiler messages. This also means the user is now shown the warning even if the compilation succeeds.

```
  [1/9] Checking `rustc` version...
  [2/9] Checking crate configuration...
  [3/9] Adding WASM target...
| [4/9] Compiling to WASM...
error[E0425]: cannot find function `fail` in this scope
 --> src/lib.rs:6:5
  |
6 |     fail();
  |     ^^^^ not found in this scope


warning: unused import: `std::ptr`
 --> src/lib.rs:1:5
  |
1 | use std::ptr;
  |     ^^^^^^^^
  |
  = note: #[warn(unused_imports)] on by default

\ [4/9] Compiling to WASM...
warning: unused import: `std::mem::replace`
 --> src/lib.rs:2:5
  |
2 | use std::mem::replace;
  |     ^^^^^^^^^^^^^^^^^


error: aborting due to previous error


For more information about this error, try `rustc --explain E0425`.

Compilation of your program failed. stderr:

   Compiling a v0.1.0 (/home/konsti/wasm-pack/a)
error: Could not compile `a`.

To learn more, run the command again with --verbose.
```

konstin:
This was made impossible by #392.

Issue #351, opened 2018-09-23
check that rust is installed, is above 1.30, or err, before all steps


ashleygwilliams:
now with our fancy installers, people might install wasm-pack... and not have rust!

i would like to land this and then remove the enforce nightly. wasm-pack should check ur rust version and error if it won't work, giving instructions to folks on how to use rustup to fix.

PR #353, opened 2018-09-24
feat(build): add rustc check

ashleygwilliams:
fixes https://github.com/rustwasm/wasm-pack/issues/351

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -16,7 +16,7 @@ name = "ansi_term"
 version = "0.11.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 dependencies = [
- "winapi 0.3.5 (registry+https://github.com/rust-lang/crates.io-index)",
+ "winapi 0.3.6 (registry+https://github.com/rust-lang/crates.io-index)",
 ]
 
 [[package]]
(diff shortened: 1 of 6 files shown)


Cargo.toml
```diff
@@ -19,6 +19,7 @@ indicatif = "0.9.0"
 lazy_static = "1.1.0"
 openssl = { version = '0.10.11', optional = true }
 parking_lot = "0.6"
+rustc_version = "0.2.3"
```
alexcrichton: FWIW using this crate can be problematic sometimes until https://github.com/Kimundi/rustc-version-rs/issues/11 has been fixed, and in general determining just the minor version of rustc is [pretty lightweight](https://github.com/alexcrichton/proc-macro2/blob/79e40a113b51836f33214c6d00228934b41bd4ad/build.rs#L44-L61)

ashleygwilliams:
chatted with @alexcrichton and we are gonna punt on testing this at the moment. it's tricky and it's probably best to just test the binary which will require a bunch of work and prevent the release today. i'll test manually to ensure it works as expected and prioritize the testing strat for this week and next release

ashleygwilliams:
currently i can only get it to this:

```
| [1/9] 🎯  Checking `rustc` version...
We can't figure out what your Rust version is- which means you might not have Rust installed. Please install Rust version 1.30.0 or higher.
```

test strategy - compile in beta or nightly, `override set stable`, `./target/debug/wasm-pack build`

ashleygwilliams:
ok, this works locally for me

```
| [1/9] 🦀  Checking `rustc` version...
Your version of Rust, '29', is not supported. Please install Rust version 1.30.0 or higher.
```

ashleygwilliams:
i might fix it to log the full version cuz that 29 looks weird

reneleonhardt:
Since the old edition 2021 is being used, Rust 1.56 should be required.
https://doc.rust-lang.org/edition-guide/rust-2021/index.html

But after looking through the long list of dependencies, the MSRV is probably much higher.
https://github.com/foresterre/cargo-msrv

PR #367, opened 2018-09-24
feat(buildmode): allow --mode force to skip rustc check

### LINKED ISSUE #366: allow -f to skip the rust version check
maybe you have an old version of rust nightly you really love. i hope you know what you're doing.

ashleygwilliams:
fixes #366

### DIFF
--- a/src/command/build.rs
+++ b/src/command/build.rs
@@ -38,6 +38,8 @@ pub enum BuildMode {
     /// Don't install tools like `wasm-bindgen`, just use the global
     /// environment's existing versions to do builds.
     Noinstall,
+    /// Skip the rustc version check
+    Force,
 }
 
 impl Default for BuildMode {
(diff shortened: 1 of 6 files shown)


src/command/build.rs
```diff
@@ -38,6 +38,8 @@ pub enum BuildMode {
     /// Don't install tools like `wasm-bindgen`, just use the global
     /// environment's existing versions to do builds.
     Noinstall,
+    /// Skip the rustc version check
```
fitzgen: I can imagine there are cases where I want to use the existing wasm-bindgen and whatever rustc I have, and just do dang it.

By making `force` a build mode, are we making this mutually exclusive with `no-install`?

ashleygwilliams: yeah so my thought is "dont check my stuff" = "force"
don't friggin install *anything* = "no-install"
i guess if you want both.. we don't let that happen at the moment, we'll see if someone asks for it?

ashleygwilliams: in my mind, no install is for docker users. i think docker users still want checks.

Issue #380, opened 2018-09-27
Relative path crate dependencies are canonicalized badly on windows


LaylConway:
## 🐛 Bug description
I was setting up a project using `parcel-plugin-wasm.rs`, which uses wasm-pack in the background.
When building the project, I got the following error.
```
C:\Users\Layl\Desktop\artist-page\services\webapp\client [master ↑ +2 ~5 -4 !]
λ  yarn run build
yarn run v1.9.4
$ parcel build index.html --public-url ./dist/
×  C:\Users\Layl\Desktop\artist-page\services\webapp\client\crate\src\lib.rs: Command failed: wasm-pack initCompilation of your program failed. stderr:

error: failed to read `\\?\C:\Users\Layl\Desktop\artist-page\libraries/simply-frontend\Cargo.toml`

Caused by:
  The filename, directory name, or volume label syntax is incorrect. (os error 123)
```
The specified path of this dependency is:
```
simply-frontend = { path = "../../../../libraries/simply-frontend" }
```
It seems wasm-pack is canonicalizing this path, which results in a UNC Path on windows, which most windows API calls do not accept, resulting in an error.

#### 🤔 Expected Behavior
The path should probably be handled either as-is, perhaps pushed after the crate's path.

#### 👟 Steps to reproduce
1. Create a crate that will be built with wasm-pack
2. Add a dependency for that crate at a relative path
3. `wasm-pack build`

#### 🌍 Your environment
Include the relevant details of your environment.
wasm-pack version: 0.5.0
rustc version: rustc 1.30.0-nightly (5c875d938 2018-09-24)

xmclark:
This has caused me some trouble too. 

I have been working around the problem by first running the raw cargo build command e.g. 

```
> cargo build --target wasm32-unknown-unknown
...
> wasm-pack build
```

xmclark:
I think this issue has been resolved with #389 . Is it time to close the issue?

LaylConway:
I don't currently have a test build to try it out but that looks like it should fix it

PR #389, opened 2018-10-04
fix canonical paths windows

xmclark:
This PR addresses problems with canonical paths on windows that were reported in #380. I tested this by building a wasm app with a few dependencies on windows.

I would love somebody to check the work and try running it on another windows box targeting another wasm app 😄 .

Implementation:
Remove the usage of std::fs::canonicalize.

Unit test:
I added a unit test based on reported failures:
- a project with local transitive dependencies, see #413

### DIFF
--- a/src/command/utils.rs
+++ b/src/command/utils.rs
@@ -4,19 +4,13 @@ use emoji;
 use failure;
 use progressbar::Step;
 use std::fs;
-use std::io;
 use std::path::{Path, PathBuf};
 use PBAR;
 
 /// If an explicit path is given, then use it, otherwise assume the current
 /// directory is the crate path.
-pub fn set_crate_path(path: Option<PathBuf>) -> io::Result<PathBuf> {
-    let crate_path = match path {
-        Some(p) => p,
-        None => PathBuf::from("."),
-    };
-
-    crate_path.canonicalize()
+pub fn set_crate_path(path: Option<PathBuf>) -> Result<PathBuf, failure::Error> {
+    Ok(path.unwrap_or(PathBuf::from(".")))
 }
 
 /// Construct our `pkg` directory in the crate.
--- a/tests/all/utils/fixture.rs
+++ b/tests/all/utils/fixture.rs
@@ -437,3 +437,156 @@ pub fn wbg_test_node() -> Fixture {
         );
     fixture
 }
+
+pub fn transitive_dependencies() -> Fixture {
+    fn project_main_fixture(fixture: &mut Fixture) {
+        fixture.file(PathBuf::from("main/README"), "# Main Fixture\n");
+        fixture.file(
+            PathBuf::from("main/Cargo.toml"),
+            r#"
+            [package]
+            authors = ["The wasm-pack developers"]
+            description = "so awesome rust+wasm package"
+            license = "WTFPL"
+            name = "main_project"
+            repository = "https://github.com/rustwasm/wasm-pack.git"
+            version = "0.1.0"
+
+            [lib]
+            crate-type = ["cdylib"]
+
+            [dependencies]
+            wasm-bindgen = "=0.2.21"
+            project_a = { path = "../project_a" }
+            project_b = { path = "../project_b" }
+
+            [dev-dependencies]
+            wasm-bindgen-test = "=0.2.21"
+        "#,
+        );
+        fixture.file(
+            PathBuf::from("main/src/lib.rs"),
+            r#"
+                extern crate wasm_bindgen;
+                use wasm_bindgen::prelude::*;
+
+                // Import the `window.alert` function from the Web.
+                #[wasm_bindgen]
+                extern {
+                    fn alert(s: &str);
+                }
+
+                // Export a `greet` function from Rust to JavaScript, that alerts a
+                // hello message.
+                #[wasm_bindgen]
+                pub fn greet(name: &str) {
+                    alert(&format!("Hello, {}!", name));
+                }
+            "#,
+        );
+    }
+
+    fn project_a_fixture(fixture: &mut Fixture) {
+        fixture.file(
+            PathBuf::from("project_a/README"),
+            "# Project Alpha Fixture\n",
+        );
+        fixture.file(
+            PathBuf::from("project_a/Cargo.toml"),
+            r#"
+            [package]
+            authors = ["The wasm-pack developers"]
+            description = "so awesome rust+wasm package"
+            license = "WTFPL"
+            name = "project_a"
+            repository = "https://github.com/rustwasm/wasm-pack.git"
+            version = "0.1.0"
+
+            [lib]
+            crate-type = ["cdylib"]
+
+            [dependencies]
+            wasm-bindgen = "=0.2.21"
+            project_b = { path = "../project_b" }
+
+            [dev-dependencies]
+            wasm-bindgen-test = "=0.2.21"
+        "#,
+        );
+        fixture.file(
+            PathBuf::from("project_a/src/lib.rs"),
+            r#"
+                extern crate wasm_bindgen;
+                // extern crate project_b;
+                use wasm_bindgen::prelude::*;
+
+                // Import the `window.alert` function from the Web.
+                #[wasm_bindgen]
+                extern {
+                    fn alert(s: &str);
+                }
+
+                // Export a `greet` function from Rust to JavaScript, that alerts a
+                // hello message.
+                #[wasm_bindgen]
+                pub fn greet(name: &str) {
+                    alert(&format!("Hello, {}!", name));
+                }
+            "#,
+        );
+    }
+
+    fn project_b_fixture(fixture: &mut Fixture) {
+        fixture.file(
+            PathBuf::from("project_b/README"),
+            "# Project Beta Fixture\n",
+        );
+        fixture.file(
+            PathBuf::from("project_b/Cargo.toml"),
+            r#"
+            [package]
+            authors = ["The wasm-pack developers"]
+            description = "so awesome rust+wasm package"
+            license = "WTFPL"
+            name = "project_b"
+            repository = "https://github.com/rustwasm/wasm-pack.git"
+            version = "0.1.0"
+
+            [lib]
+            crate-type = ["cdylib"]
+
+            [dependencies]
+            wasm-bindgen = "=0.2.21"
+
+            [dev-dependencies]
+            wasm-bindgen-test = "=0.2.21"
+        "#,
+        );
+        fixture.file(
+            PathBuf::from("project_b/src/lib.rs"),
+            r#"
+                extern crate wasm_bindgen;
+                use wasm_bindgen::prelude::*;
+
+                // Import the `window.alert` function from the Web.
+                #[wasm_bindgen]
+                extern {
+                    fn alert(s: &str);
+                }
+
+                // Export a `greet` function from Rust to JavaScript, that alerts a
+                // hello message.
+                #[wasm_bindgen]
+                pub fn greet(name: &str) {
+                    alert(&format!("Hello, {}!", name));
+                }
+            "#,
+        );
+    }
+
+    let mut fixture = Fixture::new();
+    project_b_fixture(&mut fixture);
+    project_a_fixture(&mut fixture);
+    project_main_fixture(&mut fixture);
+    fixture
+}
--- a/tests/all/build.rs
+++ b/tests/all/build.rs
@@ -125,3 +125,16 @@ fn renamed_crate_name_works() {
     .unwrap();
     fixture.run(cli.cmd).unwrap();
 }
+
+#[test]
+fn it_should_build_nested_project_with_transitive_dependencies() {
+    let fixture = utils::fixture::transitive_dependencies();
+    fixture.install_local_wasm_bindgen();
+    let cli = Cli::from_iter_safe(vec![
+        "wasm-pack",
+        "build",
+        &fixture.path.join("main").display().to_string(),
+    ])
+    .unwrap();
+    fixture.run(cli.cmd).unwrap();
+}


steveklabnik:
@xmclark what's an easy way to reproduce the bug?

xmclark:
Compiler errors on linux/mac. I think I am not correctly conditionally compiling on windows. I am welcome to suggestions, otherwise I will dig into this tonight. I was pretty sure I tried this on my mac after I built it on my windows...but maybe I made a mistake somewhere 🙃 

@steveklabnik the easiest way is creating a wasm-pack app that has rust/crate dependencies. Simply calling `wasm-pack build` might produce a build error because windows doesn't do well with UNC paths.

I will dig into it later this evening after I get over with my day job. Thanks for jumping on this everyone 😂

steveklabnik:
I spun up a new app via `cargo-generate`, added `semver-parser` to its `Cargo.toml`, modified the code to this:

```rust
extern crate cfg_if;
extern crate wasm_bindgen;
extern crate semver_parser;

mod utils;

use wasm_bindgen::prelude::*;

#[wasm_bindgen]
extern {
    fn alert(s: &str);
}

#[wasm_bindgen]
pub fn greet() {
    alert(&format!("{:?}", semver_parser::range::parse("1.0.0").unwrap()));
}
```

and here's the output:

```
> wasm-pack build

  [1/9] Checking `rustc` version...
  [2/9] Checking crate configuration...
  [3/9] Adding WASM target...
  [4/9] Compiling to WASM...
  [5/9] Creating a pkg directory...
  [6/9] Writing a package.json...
  :-) [WARN]: Field 'description' is missing from Cargo.toml. It is not necessary, but recommended
  :-) [WARN]: Field 'repository' is missing from Cargo.toml. It is not necessary, but recommended
  :-) [WARN]: Field 'license' is missing from Cargo.toml. It is not necessary, but recommended
  [7/9] Copying over your README...
  [8/9] wasm-bindgen already installed...
  [9/9] Running WASM-bindgen...
  :-) Done in 0 seconds
| :-) Your wasm pkg is ready to publish at "\\\\?\\C:\\Users\\Steve Klabnik\\tmp\\wasm-bug\\pkg".
```

So, I can't reproduce...

ashleygwilliams:
ok so- taking a look at this PR i'm starting to have the suspicion that this is not a wasm-pack bug but potentially something that needs to be fixed either in the plugin or in parcel itself. 

the first step to sorting this that i would want to see is a test for wasm-pack that can reproduce the bug you've found. we have our tests run on appveyor and there may very well be a gap in our tests, but fundamentally i think the approach we have for paths on windows is reasonable.

to say a bit more here, we have to dig into the differences between `canonicalize` and `canonicalize_path`. my understanding (again could be wrong) is that canonicalize asks your OS for the path and here `canonicalize` will, on a windows machine, return a UNC path (e.g. `\\?` which u can see in Steve's output above), notably removing the drive name, e.g. `C:`. `canonicalize_path` will return a non UNC path, which is to say it will use `C:`. So, my understanding is that this non UNC path could be wrong, which is why `canonicalize` is what we use here.

my main curiosity is exactly why the UNC path doesn't work in some cases. my assumption is that there are more constraints be added in the tooling you've layered here, but i genuinely am not sure.

hopefully this was vaguely helpful? again, starting with a test that can reproduce this issue is the best step for moving forward i think. i also would like to dive into the plugin and Parcel code, so if you have links to particular bits of those code bases that could help illuminate how these paths are being used that'd be great!

xmclark:
@steveklabnik thanks for the detailed example. I will be a bit more explicit in my example next time and provide a similar example.

I have tried a few more things on my windows box with both git bash and cmd:

I also was *not* able to reproduce the error using your example. The default dependencies in the template seem fine on windows.

I am able to reproduce a problem by installing a different package: `crc`.

I followed the same steps:
1. Install latest wasm-pack with git url,
2. `cargo generate`
3. Add `crc = "1.8.1"` to Cargo.toml
4. `wasm-pack build`

```
Mac@DESKTOP-NNBM40V MINGW64 ~/dev/reproduce-wasm-pack-pr-389 (master)
$ wasm-pack build

  [1/9] Checking `rustc` version...
  [2/9] Checking crate configuration...
  [3/9] Adding WASM target...
/ [4/9] Compiling to WASM...
Process exited with exit code: 101: Compilation of your program failed. stderr:

   Compiling crc v1.8.1
error: couldn't read "\\\\?\\C:\\Users\\Mac\\dev\\reproduce-wasm-pack-pr-389\\target\\wasm32-unknown-unknown\\release\\build\\crc-087346432c34fa56\\out/crc16_constants.rs": The filename, directory name, or volume label syntax is incorrect. (os error 123)
 --> C:\Users\Mac\.cargo\registry\src\github.com-1ecc6299db9ec823\crc-1.8.1\src\crc16.rs:8:1
  |
8 | include!(concat!(env!("OUT_DIR"), "/crc16_constants.rs"));
  | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

error: aborting due to previous error

error: Could not compile `crc`.

To learn more, run the command again with --verbose.
```

Running `cargo build --target wasm32-unknown-unknown --release --lib` will complete successfully.

```
Mac@DESKTOP-NNBM40V MINGW64 ~/dev/reproduce-wasm-pack-pr-389 (master)
$ cargo build --target wasm32-unknown-unknown --release --lib
   Compiling crc v1.8.1
   Compiling reproduce-wasm-pack-pr-389 v0.1.0 (C:\Users\Mac\dev\reproduce-wasm-pack-pr-389)
warning: function is never used: `set_panic_hook`
  --> src\utils.rs:11:9
   |
11 |         pub fn set_panic_hook() {}
   |         ^^^^^^^^^^^^^^^^^^^^^^^
   |
   = note: #[warn(dead_code)] on by default

    Finished release [optimized] target(s) in 0.76s 
```

steveklabnik:
So! I can reproduce this now, but it's only the crc crate

```
> wasm-pack build

  [1/9] Checking `rustc` version...
  [2/9] Checking crate configuration...
  [3/9] Adding WASM target...
- [4/9] Compiling to WASM...
Compilation of your program failed. stderr:

    Updating crates.io index
   Compiling build_const v0.2.1
   Compiling crc v1.8.1
error: couldn't read "\\\\?\\C:\\Users\\Steve Klabnik\\tmp\\wasm-bug\\target\\wasm32-unknown-unknown\\release\\build\\crc-d224ab13f0f73b57\\out/crc16_constants.rs": The filename, directory name, or volume label syntax is incorrect. (os error 123)
 --> C:\Users\Steve Klabnik\.cargo\registry\src\github.com-1ecc6299db9ec823\crc-1.8.1\src\crc16.rs:8:1
  |
8 | include!(concat!(env!("OUT_DIR"), "/crc16_constants.rs"));
  | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```

Compiling crc locally, both work:

```text
~\.cargo\registry\src\github.com-1ecc6299db9ec823\crc-1.8.1> cargo +nightly build --target=wasm32-unknown-unknown
   Compiling build_const v0.2.1
   Compiling crc v1.8.1 (C:\Users\Steve Klabnik\.cargo\registry\src\github.com-1ecc6299db9ec823\crc-1.8.1)
    Finished dev [unoptimized + debuginfo] target(s) in 1.48s
~\.cargo\registry\src\github.com-1ecc6299db9ec823\crc-1.8.1> cargo +nightly build
   Compiling crc v1.8.1 (C:\Users\Steve Klabnik\.cargo\registry\src\github.com-1ecc6299db9ec823\crc-1.8.1)
    Finished dev [unoptimized + debuginfo] target(s) in 0.36s
```

I have no idea what's going on here.

xmclark:
Yeah this is an especially odd one! 

I think the main reason is how the std `canonicalize` function works on windows. It is pretty well documented in [this issue](https://github.com/rust-lang/rust/issues/42869).

Calling `fs::canonicalize` on windows produces a path with a few extra leading characters e.g. `\\?C:\blah`. This kind of path is not understood by cargo (and many other programs on windows). It is also the same cause for [another issue](https://github.com/rustwasm/wasm-pack/issues/390) reported recently about the pkg path not being printed correctly. 

This PR implements a solution that follows a strategy suggested by a few others in that rust-lang issue.

xmclark:
Agreed. A separate issue is the display as a UNC path. 

If it were un-escaped, it would look like this:
```| :-) Your wasm pkg is ready to publish at "\\?\C:\Users\Steve Klabnik\tmp\wasm-bug\pkg"```

I would have also expected to to be printed without the leading `\\?\`. 

Wow! This comment is starting to sound a lot like an issue report... [I made this fix as well in this PR](https://github.com/rustwasm/wasm-pack/pull/389/files#diff-3b7d580454f5224e58f3687e4eb6568fL144). If the project would prefer that be a separate PR, I'm up for removing that extra revision.

**Comment by ashleygwilliams** [comment]
hey! so- this issue PR convo is indeed getting long. @xmclark in order to accept this PR, i need a test case that demonstrates this bug. let me know if you need help or have questions about that!

**Comment by ashleygwilliams** [comment]
@xmclark following up- are you around on a sync channel anywhere? would it be helpful to discuss a bit?

**Comment by xmclark** [comment]
@ashleygwilliams hey sorry for late reply. Long week.

I apologize if there has been some miscommunication on my part. I would like to do whatever I need to help reviewers. I was under the impression that I had provided all the needed information for this to be reviewed. The description references the issue number that this PR fixes. [This comment](https://github.com/rustwasm/wasm-pack/pull/389#issuecomment-427230272) demonstrates how to reproduce the bug. 

@technetos was super helpful and got me onto the rust-lang and the rust programming language discords last week. Are those good channels for discussion? Thanks!

**Comment by steveklabnik** [comment]
That comment gives us some kind of reproduction but it doesn’t translate into a test that we could use to ensure that it won’t regress, and it also is “too big” to really sort out what the root issue actually is. In my understanding a unit test is what’s being sought here.

File: src/command/utils.rs [hunk 1]
```diff
@@ -44,3 +42,29 @@ pub fn find_pkg_directory(path: &Path) -> Option<PathBuf> {
...
+            if vec_chars[0..4] == [92, 92, 63, 92] {
+                Ok(Path::new(&OsString::from_wide(&vec_chars[4..])).to_owned())
+            }
+            else {
+                Ok(canonical)
+            }
+        }
+    }
+    else {
+        /// Strips UNC from canonical path on Windows.
```
raggy-rs: this comment is missplaced

raggy-rs:
Maybe the repro of https://github.com/rustwasm/wasm-pack/issues/413 could be made into a unit test. The underlying issue is that when you call cargo from a std::process::Command with a UNC Path cargo cannot handle it.

ashleygwilliams:
@xmclark you can find me on most rust platforms as @ag_dubs if you'd like to talk sync- and no worries on the delay- it's important to fix this but i have also been traveling :)

rpjohnst:
(Summarizing https://github.com/rustwasm/wasm-pack/issues/413#issuecomment-433196674 here:)

A better solution than stripping `\\?\` from the beginning of paths (which is incorrect in general, since it can change the meaning of the path) would be to call `GetFullPathNameW`. This will behave more like we expect `fs::canonicalize` to behave.

xmclark:
@rpjohnst hey that is a really good point! I was never a super fan of stripping the extra characters...it feels kludgy. I will look into `GetFullPathNameW`.

xmclark:
I'm not sure what's up with the appveyor build. This code is up for review! I have decided on a different implementation which avoids some of the problems with `std::canonicalize`. I also added two tests that capture the reported fail cases.

fitzgen:
> I'm not sure what's up with the appveyor build.

If you rebase on master, the appveyor builds should be fixed now. See #417 for details.

fitzgen:
This PR looks good to me. I'm unfamiliar with the absolutize crate, but if it fixes the problem then that sgtm. Do either of @ashleygwilliams or @alexcrichton want to verify that this is the correct fix?

tests/all/utils/fixture.rs
```diff
@@ -474,3 +474,207 @@ pub fn wbg_test_node() -> Fixture {
...
+            name = "proj"
+            repository = "https://github.com/rustwasm/wasm-pack.git"
+            version = "0.1.0"
+
+            [lib]
+            crate-type = ["cdylib"]
+
+            [dependencies]
+            wasm-bindgen = "=0.2.21"
+            crc = "*"
```
steveklabnik: this is *extremely* fragile

I'm not sure about this strategy overall

ashleygwilliams: yeah i will need to talk a look at this a bit closer but i am not in favor of * deps in general, and i would prefer a more general use case test

xmclark: Yeah I also agree, and I'm glad you both point it out.

In this case, I was having a problem installing `crc = "1.8.1"`. I left it as * because pinning the version makes sense today, but if there was a change tomorrow in crc, we wouldn't catch it.

I can pin the version if that is preferred 😄

Cargo.toml
```diff
@@ -21,6 +21,7 @@ indicatif = "0.9.0"
 lazy_static = "1.1.0"
 openssl = { version = '0.10.11', optional = true }
 parking_lot = "0.6"
+path-absolutize = "1.1.1"
```
rpjohnst: Was there a particular reason for using `path-absolutize`, which implements its own custom logic, rather than just calling `GetFullPathName` on Windows? I expect the latter would be more robust.

alexcrichton:
I'm unfortunately away for windows for awhile now so I can't verify this, but sorry if this has already been asked by why is canonicalization done? Could the original path just be plumbed around?

xmclark:
@alexcrichton that is a good point. I removed absolutize (and canonicalize) and it just works. It seems that we do not need to canonicalize the path at all.

alexcrichton:
Ok great! 

Is this ready to go in that case? Or are there final points to take care of?

xmclark:
Thanks!

I think I could use some feedback on how to fix my unit test that uses the crc-rs crate. The test is valid, but I have received feedback that not pinning the dependency version is bad. I am also a little concerned that we are NOT testing for the underlying problem. I think I understand that fundamental issue, but recreating it in a test fixture would could be really complicated. The crc-rs crate has a custom build step that generates some rust code in the out_dir. When included, the path to the generated file is in the windows UNC path format, and cargo chokes on it. This only happens when building with wasm-pack and this PR fixes that issue. 

I am interested in any suggestions on this!

xmclark:
Thanks @steveklabnik @alexcrichton! I have removed the test.

alexcrichton:
👍

PR #392, opened 2018-10-04
Child process and output management

### LINKED ISSUE #287: Show the output of Cargo's build

fitzgen:
Fix #287 but also all the instances of this that we might ever run into.

### DIFF
--- a/src/bindgen.rs
+++ b/src/bindgen.rs
@@ -1,8 +1,10 @@
 //! Functionality related to installing and running `wasm-bindgen`.
 
 use binaries::{self, bin_path, install_binaries_from_targz_at_url};
+use child;
 use emoji;
 use error::Error;
+use failure::{self, ResultExt};
 use progressbar::Step;
 use slog::Logger;
 use std::path::{Path, PathBuf};
@@ -22,7 +24,7 @@ pub fn install_wasm_bindgen(
     install_permitted: bool,
     step: &Step,
     log: &Logger,
-) -> Result<(), Error> {
+) -> Result<(), failure::Error> {
     // If the `wasm-bindgen` dependency is already met, print a message and return.
     if wasm_bindgen_path(log, root_path)
         .map(|bindgen_path| wasm_bindgen_version_check(&bindgen_path, version, log))
@@ -37,7 +39,7 @@ pub fn install_wasm_bindgen(
     // permitted, return a configuration error.
     if !install_permitted {
         let msg = format!("wasm-bindgen v{} is not installed!", version);
-        return Error::crate_config(&msg);
+        return Err(Error::crate_config(&msg).into());
     }
 
     let msg = format!("{}Installing wasm-bindgen...", emoji::DOWN_ARROW);
@@ -48,7 +50,7 @@ pub fn install_wasm_bindgen(
             log,
             "could not download pre-built `wasm-bindgen`: {}. Falling back to `cargo install`.", e
         );
-        cargo_install_wasm_bindgen(root_path, version)
+        cargo_install_wasm_bindgen(log, root_path, version)
     })
 }
 
@@ -81,28 +83,23 @@ pub fn download_prebuilt_wasm_bindgen(root_path: &Path, version: &str) -> Result
 
 /// Use `cargo install` to install the `wasm-bindgen` CLI locally into the given
 /// crate.
-pub fn cargo_install_wasm_bindgen(crate_path: &Path, version: &str) -> Result<(), Error> {
-    let output = Command::new("cargo")
-        .arg("install")
+pub fn cargo_install_wasm_bindgen(
+    logger: &Logger,
+    crate_path: &Path,
+    version: &str,
+) -> Result<(), failure::Error> {
+    let mut cmd = Command::new("cargo");
+    cmd.arg("install")
         .arg("--force")
         .arg("wasm-bindgen-cli")
         .arg("--version")
         .arg(version)
         .arg("--root")
-        .arg(crate_path)
-        .output()?;
-    if !output.status.success() {
-        let message = "Installing wasm-bindgen failed".to_string();
-        let s = String::from_utf8_lossy(&output.stderr);
-        Err(Error::Cli {
-            message,
-            stderr: s.to_string(),
-            exit_status: output.status,
-        })
-    } else {
-        assert!(binaries::local_bin_path(crate_path, "wasm-bindgen").is_file());
-        Ok(())
-    }
+        .arg(crate_path);
+
+    child::run(logger, cmd, "cargo install").context("Installing wasm-bindgen with cargo")?;
+    assert!(binaries::local_bin_path(crate_path, "wasm-bindgen").is_file());
+    Ok(())
 }
 
 /// Run the `wasm-bindgen` CLI to generate bindings for the current crate's
@@ -116,7 +113,7 @@ pub fn wasm_bindgen_build(
     debug: bool,
     step: &Step,
     log: &Logger,
-) -> Result<(), Error> {
+) -> Result<(), failure::Error> {
     let msg = format!("{}Running WASM-bindgen...", emoji::RUNNER);
     PBAR.step(step, &msg);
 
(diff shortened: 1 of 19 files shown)


fitzgen:
First build, that has to build the wasm-bindgen macro:

wasm-pack-build

Rebuilding again, which is super fast:

wasm-pack-build-again

src/child.rs
```diff
@@ -0,0 +1,176 @@
...
+
+            if exit.success() {
+                return Ok(stdout);
+            } else {
+                let msg = format!("`{}` did not exit successfully", command_name);
+                return Err(Error::cli(&msg, stderr.into(), exit).into());
+            }
+        }
+
+        thread::yield_now();
```
alexcrichton: I think we'll definitely want a synchronization solution before this lands to ensure that we're not just eating a core spin-waiting while cargo tries to finish. We don't actually have to `try_wait` or wait on the process at all until stdout/stderr are closed, so we can wait until those are done to do that and otherwise just loop around receiving on the channel (which we can do with a timeout as well)

src/child.rs
```diff
@@ -0,0 +1,176 @@
...
+    let (send, recv) = mpsc::channel();
+    let stdout_send = send.clone();
+    let stderr_send = send;
+
+    // Because pipes have a fixed-size buffer, we need to keep reading stdout
+    // and stderr on a separate thread to avoid potential dead locks with
+    // waiting on the child process.
+
+    let stdout_handle = thread::spawn(move || {
+        for line in stdout.lines() {
```
alexcrichton: I've had unfortunately bad experiences in the past line-buffering stdout. An example of the bad behavior is that libtest in quiet mode used to simply print a `.`-per-test and then at the very end it'd print a newline. This logic means that no output would happen until the very very end where everything comes out all at once. We [fixed this bug for Rust](https://github.com/alexcrichton/stamp-rs/pull/1) in the past, and I think a similar fix could be applied here where this just streams data as it comes, and the data processor specially handles newlines in the stream to append output.

src/child.rs
```diff
@@ -0,0 +1,176 @@
...
+    Stderr(S),
+}
+
+fn print_child_output<S>(out: Result<Output<S>, io::Error>, command_name: &str)
+where
+    S: AsRef<str>,
+{
+    let message = match out {
+        Ok(Output::Stdout(line)) => format!("{} (stdout): {}", command_name, line.as_ref()),
+        Ok(Output::Stderr(line)) => format!("{} (stderr): {}", command_name, line.as_ref()),
```
alexcrichton: I'm not sure we want to prefix things with `(stderr)` and/or `(stdout)` as it may quickly add noise to the output. Do you think it's worth differentiating? (people may be alarmed, for example, that Cargo always prints to stderr despite it typically not generating errors).

Additionally, the `command_name` may be best to print at the very beginning? Something like `Running ...` like in Cargo so it's there if necessary and otherwise not taking up too much space.

src/child.rs
```diff
@@ -0,0 +1,176 @@
...
+            stderr_send.send(line.map(Output::Stderr)).unwrap();
+        }
+    });
+
+    let mut stdout = String::new();
+    let mut stderr = String::new();
+    let mut is_long_running = false;
+
+    loop {
+        if !is_long_running && taking_too_long(since) {
```
alexcrichton: I'm not sure I'm 100% convinced this strategy of swallowing the first 200ms of output will work out in the long run. In the worst case a tool finishes very quickly but some crucial information it prints to stdout/stderr is lost. I don't think that wasm-pack invokes many tools like that today though? This is largely just in terms of future concerns.

In general I feel we should depend on tools responsibly not generating lots of output that's not necessary. Like adding a wasm target with rustup which is already there should produce zero output, but I'm not sure that we should gauge that based on the 200ms delay here. (similarly Cargo's "fresh" output of "you compiled this in 0.2s" is something we'd want to hide, but there isn't a great way to do that in Cargo right now unfortunately).

src/child.rs
```diff
@@ -0,0 +1,176 @@
...
+            // Join the threads reading the child's output to make sure the
+            // finish OK.
+            stdout_handle.join().unwrap();
+            stderr_handle.join().unwrap();
+
+            if exit.success() {
+                return Ok(stdout);
+            } else {
+                let msg = format!("`{}` did not exit successfully", command_name);
+                return Err(Error::cli(&msg, stderr.into(), exit).into());
```
alexcrichton: This'll want to ferry along stdout as well, want to be sure to not lose anything whenever we print failure messages!

I think it'd also be good to carry along the full command line here as opposed to just the name of the command to ensure that all the arguments are in the error message too

src/child.rs
fitzgen: So you're saying you think we should always display the child's output unconditionally?

alexcrichton: After talking over lunch I think the conclusion was that we should probably remove this yeah, and have the main thread block on dispatching output received from the channel

fitzgen: For context:

* We *really* don't want to swallow warnings from rustc, even when cargo build exits super quickly.

* That general scenario could play out with other CLI tools too

* The noisiest parts of the `wasm-pack build` output are actually things like "checking if you have wasm32-u-u installed" when 99% of the time the user already does. For this particular case, it is wasm-pack itself printing this out, not a child process, so this module is unaffected.

Therefore, we should always print all output, regardless of time the child takes to run.

For that last bullet point, and any kind of "install X if it isn't already here", we only want to print in the case where X isn't already installed and we are going to commence with installing it.

alexcrichton: Er and to elaborate a bit! I think it's fine to leave this in if it's just hypothetical things I'm worried about and I can always file a follow-up issue, but one concrete use case we discovered which may turn out badly is a `rustc` invocation which prints a compiler warning (like a diagnostic) but finishes in under 200ms. I think the logic here would swallow up the warning and avoid it getting printed.

alexcrichton: Er oops sorry, double post at the same time!

fitzgen:
Ok, new commits make it so that

* we don't do the 200ms thing, and instead always send child output to the console,
* we don't line buffer the child's stdout and stderr
* we maintain the child's stdout in the `Error::Cli` variant

fitzgen:
Note: we will want follow up PRs to allow making certain commands run in "quiet" mode where we don't dump their output to our console, but (a) I think having these switched defaults is better in the general case, and (b) we can do that as a follow up issue.

fitzgen:
Notably, we still want "quiet" commands to have their output end up in the logs, just not the console.

src/bindgen.rs
```diff
@@ -81,28 +83,23 @@ pub fn download_prebuilt_wasm_bindgen(root_path: &Path, version: &str) -> Result
...
 /// crate.
-pub fn cargo_install_wasm_bindgen(crate_path: &Path, version: &str) -> Result<(), Error> {
-    let output = Command::new("cargo")
-        .arg("install")
+pub fn cargo_install_wasm_bindgen(
+    logger: &Logger,
+    crate_path: &Path,
+    version: &str,
+) -> Result<(), failure::Error> {
+    let mut cmd = Command::new("cargo");
```
ashleygwilliams: is this mut here necessary?

src/bindgen.rs
```diff
@@ -153,27 +150,20 @@ pub fn wasm_bindgen_build(
...
 /// Check if the `wasm-bindgen` dependency is locally satisfied.
 fn wasm_bindgen_version_check(bindgen_path: &PathBuf, dep_version: &str, log: &Logger) -> bool {
-    Command::new(bindgen_path)
-        .arg("--version")
-        .output()
-        .ok()
-        .filter(|output| output.status.success())
-        .map(|output| {
-            String::from_utf8_lossy(&output.stdout)
+    let mut cmd = Command::new(bindgen_path);
```
ashleygwilliams: is this mut here necessary?

src/build.rs
```diff
@@ -54,28 +57,23 @@ fn rustc_minor_version() -> Option<u32> {
...
-        let s = String::from_utf8_lossy(&output.stderr);
-        Error::cli(
-            "Adding the wasm32-unknown-unknown target failed",
-            s,
-            output.status,
-        )
-    } else {
-        Ok(())
-    }
+    let mut cmd = Command::new("rustup");
```
ashleygwilliams: is this mut here necessary?

src/npm.rs
```diff
@@ -1,56 +1,51 @@
...
-pub fn npm_pack(path: &str) -> Result<(), Error> {
-    let output = Command::new("npm").current_dir(path).arg("pack").output()?;
-    if !output.status.success() {
-        let s = String::from_utf8_lossy(&output.stderr);
-        Error::cli("Packaging up your code failed", s, output.status)
-    } else {
-        Ok(())
-    }
+pub fn npm_pack(log: &Logger, path: &str) -> Result<(), failure::Error> {
+    let mut cmd = Command::new("npm");
```
ashleygwilliams: is this mut necessary?

fitzgen:
The `mut`s are necessary because the methods for adding arguments etc to a command take `&mut self`.

PR #394, opened 2018-10-05
Change to INFO when alerting users about missing fields Cargo.toml

### LINKED ISSUE #393: Warnings about missing Cargo.toml keys are scarier than need be
When watching a stream of someone going through our tutorial, they got surprised/flustered/worried by these warning, so I think we can tone it down a bit somehow.

Perhaps this can just be "info" instead of all-caps and all-bold **WARN**? Perhaps we can collapse all the missing keys warnings into a single warning instead of multiple?

mstallmo:
Closes #393 

I switched the level displayed in the progress bar from `PBAR.warn` to `PBAR.info` and changed the format closure from `warn_format` to `info_format` for consistency.

How the new terminal output looks:
info_output

```bash
$ rustup override set nightly
```

### DIFF
--- a/src/manifest/mod.rs
+++ b/src/manifest/mod.rs
@@ -38,7 +38,7 @@ struct CargoPackage {
 
 impl CargoPackage {
     fn check_optional_fields(&self) {
-        let warn_fmt = |field| {
+        let info_fmt = |field| {
             format!(
                 "Field '{}' is missing from Cargo.toml. It is not necessary, but recommended",
                 field
@@ -46,13 +46,13 @@ impl CargoPackage {
         };
 
         if self.description.is_none() {
-            PBAR.warn(&warn_fmt("description"));
+            PBAR.info(&info_fmt("description"));
         }
         if self.repository.is_none() {
-            PBAR.warn(&warn_fmt("repository"));
+            PBAR.info(&info_fmt("repository"));
         }
         if self.license.is_none() {
-            PBAR.warn(&warn_fmt("license"));
+            PBAR.info(&info_fmt("license"));
         }
     }
 }


ashleygwilliams:
@fitzgen do you think this is enough to make it less scary, or should we also try to make it into a single info line? 

(thanks @mstallmo this looks good to me but we may want to take it just a bit further- if you're up for it we can do it on this PR, otherwise i can accept this and do it today)

mstallmo:
@ashleygwilliams No problem, I'm up for updating this PR to reflect whatever we think is the less scary way to inform users of the missing fields!

mstallmo:
Will do! I can probably get to this in the next couple of hours over lunch. 

For formatting I’m assuming we just want to have the missing fields separated by commas in the INFO line?

ashleygwilliams:
@mstallmo yeah, separated by commas sounds good to me, e.g. "Optional fields missing from Cargo.toml: 'x', 'x', 'x'. These are not necessary but recommended."

fitzgen:
> @fitzgen do you think this is enough to make it less scary, or should we also try to make it into a single info line?

I think switching both to INFO and to single line is good :+1:

mstallmo:
I've updated to combine the INFO print out to one line with fields separated by commas.  I've attached pictures how the print out looks with the various combinations of printout:

one

two

three

Side note: It looks like running rustfmt on the mod.rs file picked up a change with how `.map` was formatted on line 84. If we don't want to include that here I can reset the commit and undo that change.

ashleygwilliams:
@mstallmo appreciate the note on the diff but i think it's fine :) this all looks awesome- thanks for your efforts. waiting on CI to be green but i think we can merge this!

mstallmo:
Welp, just ran cargo fmt and it looks like the formatting for visual studio on my Mac was not configured correctly. I fixed the issue and updated.

PR #396, opened 2018-10-05
bugfix(command/build.rs): Cancel Escape

### LINKED ISSUE #390: Windows path output is unneccesarily escaped
## 🐛 Bug description

When I run wasm-pack, I get a final output that looks like this

```text
| :-) Your wasm pkg is ready to publish at "\\\\?\\C:\\Users\\Steve Klabnik\\tmp\\wasm-bug\\pkg".
```

#### 🤔 Expected Behavior

It should look like this:

```text
| :-) Your wasm pkg is ready to publish at "\\?\C:\Users\Steve Klabnik\tmp\wasm-bug\pkg".
```

#### 👟 Steps to reproduce

Build the hello world template on Windows.

#### 🌍 Your environment

wasm-pack version: wasm-pack 0.5.0
rustc version: rustc 1.31.0-nightly (5597ee8a6 2018-10-03)

huangjj27:
fix #390

```bash
$ rustup override set nightly
```


fitzgen:
> so PathBuf doesn't have a default display on purpose- you'll likely want to use `to_str` or `to_string_lossy` (the previous can fail, the former is likely the better option here, though may sometimes print weird things).

I believe this is the preferred method for formatting paths: https://doc.rust-lang.org/nightly/std/path/struct.Path.html#method.display

src/command/build.rs
```diff
@@ -133,15 +133,19 @@ impl Build {
...
         PBAR.message(&format!("{} Done in {}", emoji::SPARKLE, &duration));
 
         PBAR.message(&format!(
-            "{} Your wasm pkg is ready to publish at {:#?}.",
+            "{} Your wasm pkg is ready to publish at {}.",
             emoji::PACKAGE,
-            self.out_dir.canonicalize().unwrap_or(self.out_dir.clone())
+            self.out_dir
+                .canonicalize()
+                .unwrap_or(self.out_dir.clone())
```
csmoe: consider lazily evaluated `unwrap_or_else` here

ashleygwilliams:
@huangjj27 thanks for this! one ask- could you please include test(s)?

huangjj27:
@ashleygwilliams I would be glad to! but what kinds of tests should I add? Is it OK to just add a unit test to check the modified function?

csmoe:
@huangjj27 add a test case [here](https://github.com/rustwasm/wasm-pack/blob/master/tests/all/build.rs) to check whether windows path output is unneccesarily escaped as this PR intended to fix.

ashleygwilliams:
@huangjj27 if you would prefer i can also grab this PR and add the tests for you! just let me know what you'd like :)

huangjj27:
Hi! I added a test for the change, but the test case seems to keep failing on installing wasm-bindgen, but the source code of test crate generated can be built using normal call of `wasm-pack build`, is that caused by something special in `command::run_wasm_pack`?

huangjj27:
@ashleygwilliams  thanks! I'm working on the test(s) and glad to do the job, but I need some hints and times to finish it (for I could use only about an hour at night during workdays)

huangjj27:
It seems the test case triggered an panic
, which is not found at local

huangjj27:
It's so weird that I passed the test on my windows but failed on appveyor. I even can't get the logs from appveyor. What should I do next?

huangjj27:
It's so strange that the result of `cargo fmt` is totally different. The code pushed is formatted in local

huangjj27:
```
---- build::it_format_out_dir_on_windows stdout ----
Created fixture at C:\projects\wasm-pack-071k0\target\t\.tmpReZwQh\wasm-pack
   [ERR]: There was a crate configuration error. Details:
thread 'build::it_format_out_dir_on_windows' panicked at 'directories in wasm-pack.log should be well formatted', tests\all\build.rs:56:5
stack backtrace:
failures:
    build::it_format_out_dir_on_windows
test result: FAILED. 34 passed; 1 failed; 0 ignored; 0 measured; 0 filtered out
error: test failed, to rerun pass '--test all'
Command exited with code 101
```

Hi, @ashleygwilliams , I get such error that never occur at local and I'm not sure if it's the appveyor that causes the problem. Would you re-trigger the CI?

fitzgen:
I fixed the appveyor CI on master, so if you rebase, then it should start passing and we can merge this!

fitzgen:
err, I suppose this hasn't had another round of review yet, so maybe not merge immediately after rebase :-p

huangjj27:
Repeatable Error of `There was a crate configuration error.`. Confusing.

src/command/build.rs
```diff
@@ -133,15 +133,19 @@ impl Build {
...
 
         PBAR.message(&format!("{} Done in {}", emoji::SPARKLE, &duration));
 
         PBAR.message(&format!(
-            "{} Your wasm pkg is ready to publish at {:#?}.",
+            "{} Your wasm pkg is ready to publish at {}.",
             emoji::PACKAGE,
-            self.out_dir.canonicalize().unwrap_or(self.out_dir.clone())
+            self.out_dir
+                .canonicalize()
```
alexcrichton: I think this may actually be a case where the call to `canonicalize()` could be omitted entirely? This could in that case simply ber `self.out_dir.display()` perhaps?

alexcrichton:
Ah it looks like the failure on Windows is due to changes on master, and I think the test may equally apply to Unix now perhaps?

huangjj27:
@alexcrichton it happened on windows with msvc target, but I forgot to rebase
, which cause the failure. Local test failed again and is still to be fixed

huangjj27:
@alexcrichton Can I update the wasm-bindgen version from `=0.2.21` to `=0.2.27` ? I'm affraid that there is some bugs about msvc fixed in `wasm-bindgen 0.2.23` or later version but not in  `wasm-bindgen 0.2.21`.

huangjj27:
@alexcrichton Can I update the wasm-bindgen version of fixtures from `=0.2.21` to `=0.2.27` ? I'm affraid that there is some bugs about msvc fixed in `wasm-bindgen 0.2.23` or later version but not in  `wasm-bindgen 0.2.21`.

Besides, I also need someone to re produce what've happened on Appveyor CI. the test just passed on my local machine perfectly. Would you offer me a hand, @fitzgen?

alexcrichton:
It should be fine to update `wasm-bindgen`, yeah, but I suspect that's not the issue at hand here. You'll probably want to update the test to print out the contents of wasm-pack.log on CI so you can debug why it doesn't contain the string in question. I believe it's probably related to the fact that AppVeyor checks out code in a different directory

fitzgen:
> Besides, I also need someone to reproduce what've happened on Appveyor CI. the test just passed on my local machine perfectly. Would you offer me a hand, @fitzgen?

You could try pushing an empty commit. I don't have access to the appveyor ci, so I can't trigger a re-run myself.

+cc @ashleygwilliams

huangjj27:
trying to just print the wasm_pack.log on CI seems useless, I'll dive in the code more to figure out why.

> You could try pushing an empty commit. I don't have access to the appveyor ci, so I can't trigger a re-run myself.

@fitzgen Sorry I don't make it clear. I means that maybe you can pull this branch to local, to see what happens to the test and its log. I passed the test at my local, so I pushed the commit, but the CI complains that my test does always fail, but hasn't give me anything about the log. It seems the decision of not outputing what a example did causes some trouble here

huangjj27:
I once thought that it's using git bash in my vscode that makes the test failed. but after I change to powershell as integrated terminal, all test passed again.

huangjj27:
I finally figure out there is difference of the path  between ch and local, and
I shouldn’t hard code much.

PR #408, opened 2018-10-13
test(command/build): add a test for build command

huangjj27:
this PR is to check if a minimal example `fixture::js_hello_world` can be built in the test.

it provides precondition for #396 

---

```bash
$ rustup override set nightly
```

### DIFF
--- a/tests/all/build.rs
+++ b/tests/all/build.rs
@@ -20,3 +20,20 @@ fn build_in_non_crate_directory_doesnt_panic() {
     let err_msg = result.unwrap_err().to_string();
     assert!(err_msg.contains("missing a `Cargo.toml`"));
 }
+
+#[test]
+fn it_should_build_js_hello_world_example() {
+    let fixture = utils::fixture::js_hello_world();
+    let cli = Cli::from_iter_safe(vec![
+        "wasm-pack",
+        "build",
+        &fixture.path.display().to_string(),
+    ])
+    .unwrap();
+    let logger = logger::new(&cli.cmd, cli.verbosity).unwrap();
+    let result = command::run_wasm_pack(cli.cmd, &logger);
+    assert!(
+        result.is_ok(),
+        "running wasm-pack in a js-hello-world directory should succeed.",
+    );
+}


huangjj27:
As is expected, this PR would fail because we set `CARGO_TARGET_DIR` to `wasm-pack/target`,
so when we `run_wasm_pack` with a temporary proejct(wasm-pack/target/t/.tmpXXXXXX/wasm-pack), the  outputed files that should be in tmp_project/wasm32-unknown-unknown would located at wasm-pack/target/wasm32-unknown-unknown. Pending #396 until this test is passed

huangjj27:
And I notice that, using `command::run_wasm_pack` to test **never share** the wasm-binggen at all

huangjj27:
I'm sorry to make thing complex. this PR is no need to use.

huangjj27:
Besides,  here is a question: would it be able to modify the same CARGO_TARGET_DIR when multiple tests are running?

huangjj27:
I found the related code: https://github.com/rustwasm/wasm-pack/blob/master/src/bindgen.rs#L131, it is dead code to a certain path and will not change according to $CARGO_TARGET_DIR. 

one more concern is that if we share the same target directory, will it possible to change the target files when there is multiple test running?

fitzgen:
```
  error: failed to read `target/wasm32-unknown-unknown/release/js_hello_world.wa
| 	caused by: No such file or directory (os error 2)
test build::it_should_build_js_hello_world_example ... FAILED
```

I hope this is not actually trying to read a `.wa` file, and this is just the output getting cut off and it really is a `.wasm` file.

> Besides, here is a question: would it be able to modify the same CARGO_TARGET_DIR when multiple tests are running?

Cargo ensures that target dirs are used in a mutually exclusive way, so we are safe from races. Is that your concern?

> I found the related code: https://github.com/rustwasm/wasm-pack/blob/master/src/bindgen.rs#L131, it is dead code to a certain path and will not change according to $CARGO_TARGET_DIR.

I don't see how it is dead code, can you explain some more?

To make this test pass, we should be using https://docs.rs/cargo_metadata/0.6.0/cargo_metadata/struct.Metadata.html#structfield.target_directory to get the target directory. We are already using that crate for getting hte workspace root: https://github.com/rustwasm/wasm-pack/blob/60bdb80947cac4df9fe2778685cca857e6961854/src/lockfile.rs#L69-L73

huangjj27:
> I hope this is not actually trying to read a .wa file

It's a mis-copying.

> I don't see how it is dead code, can you explain some more?

I means that the command try to find wasm files in the relative target directory while the wasms have been outputed at `$CARGO_TARGET_DIR/wasm32-unknown-unknown`

> To make this test pass, we should be using https://docs.rs/cargo_metadata/0.6.0/cargo_metadata/struct.Metadata.html#structfield.target_directory to get the target directory.

I've used `std::env::var("CARGO_TARGET_DIR")` to get the target directory as is described in [cargo-book/environment-variables](https://doc.rust-lang.org/cargo/reference/environment-variables.html). It seems work fine. Must I change for cargo metadata?

fitzgen:
> I've used `std::env::var("CARGO_TARGET_DIR")` to get the target directory as is described in [cargo-book/environment-variables](https://doc.rust-lang.org/cargo/reference/environment-variables.html). It seems to work fine. Must I change for cargo metadata?

Hm, I'm actually not sure. But @alexcrichton would know.

alexcrichton:
I think `CARGO_TARGET_DIR` is only read by cargo, not written, so it's not guaranteed to be set. Reading it from `cargo metadata` would be the most reliable strategy!

huangjj27:
@alexcrichton 
> so it's not guaranteed to be set.
Yes, so in the codes I match if the environment variable exists.

huangjj27:
it seems that the build outputs too much and takes a long time. How to solve the problem ?

fitzgen:
It looks like it is hung at the end of the CI run. Will try poking CI and seeing if that fixes it.

Ideally we would detect when stdout is not a real terminal and avoid doing the spinner and fancy output and just do the normal output. We have an issue on file somewhere...

huangjj27:
@fitzgen

> Ideally we would detect when stdout is not a real terminal and avoid doing the spinner and fancy output and just do the normal output.

I think we can redirect the stderr to a null, for all test result output to stdout and corespondent error will output there too

huangjj27:
> I'm not sure I follow your proposal, do you mean from the script command in .travis.yml or inside wasm-pack code itself?

I mean the script command.  for example, `cargo test --locked 2>test.log`, if we don't need that output, I think `cargo test --locked 2>/dev/null` will work fine too.

Issue #403, opened 2018-10-09
wasm-pack should check if it's the latest version and warn if not


ashleygwilliams:
when you run build, wasm-pack should check and see what version is latest on crates.io and if the current installed version is latest. if not, it should suggest the user update.

daubaris:
I would love to work on this one!

drager:
@daubaris: Hey, I discussed this very briefly on IRC with @ashleygwilliams. One approach to this is to check the behavior of the [npm client](https://npme.npmjs.com/docs/cli/installation.html) or [yarn](https://yarnpkg.com) and see what they output for outdated packages. Npm outputs something like: "npm WARN ..." with a yellow background and black text. Yarn outputs something like: "warning ..." in yellow text. So maybe, the text could be something like: "warning (in yellow) :warning: There's a newer version of wasm-pack available, the new version is x.x.x and you're using version x.x.x". 

Regarding the approach to this issue. I first had the idea of using something like [hyper](https://crates.io/crates/hyper) for this. Make a request to the [crates api](https://crates.io/api/v1) and fetch the latest version of the crate and then compare that to installed one. However, `wasm-pack` already depends on the [curl crate](https://crates.io/crates/curl) so I think that's a better choice. So, use curl to fetch the latest version of the crate from the crates api and then compare that version with the locally installed one, then output a warning if there's a newer one available.

This is my understanding of the feature request, so anyone, please correct me if I'm wrong.

ashleygwilliams:
closed by #409

PR #409, opened 2018-10-13
added wasm-pack feature which compares local and latest wasm-pack versions

daubaris:
This PR refers to issue #403. Added all the needed stuff, although not quite sure if some choices that I made were correct, would be nice if somebody could check this and explain further! Throw them comments at me! :smile:


fitzgen:
Let me know if you have any questions :)

daubaris:
Okay @fitzgen, i have few questions:
1. For creating a file i'd use std::env::temp_dir, is that appropriate for this kind of situation?
2. If the temporary file was created few months ago and user wants to use wasm-pack when she/he has no internet connection what should it display?
3. The thread should be spawned as soon as wasm-pack was started, right? (never worked with threads :( )
4. By unwrapping errors you mean the part where I call to the API? because indeed that's dumb of me...

fitzgen:
> For creating a file i'd use std::env::temp_dir, is that appropriate for this kind of situation?

Perhaps? I suppose if the OS has cleaned up the temp dir and removed our file, that is a signal that it has likely been long enough since we last checked for updates that we can do it again. I was originally thinking of some file in `~/.wasm-pack` or soemthing, but since we don't maintain such a directory yet, we would have to lay out a bunch of infrastructure in this PR and it isn't 100% clear to me that this is the right thing to do. Perhaps @alexcrichton has thoughts on this?

> If the temporary file was created few months ago and user wants to use wasm-pack when she/he has no internet connection what should it display?

If there is no internet connection, we should just not put anything on the channel, and therefore the main thread won't report anything.

> The thread should be spawned as soon as wasm-pack was started, right? (never worked with threads :( )

Probably the best place to spawn the thread is at the top of `run_wasm_pack` in `src/command/mod.rs` and then check for the possible "upgrade is available" message on the channel at the end of `run_wasm_pack`.

This should hopefully get you started:

```diff
diff --git a/src/command/mod.rs b/src/command/mod.rs
index ee418da..226282a 100644
--- a/src/command/mod.rs
+++ b/src/command/mod.rs
@@ -83,8 +83,25 @@ pub enum Command {
     Test(TestOptions),
 }
 
+// Checks for updates to wasm-pack in the background. If an update is found, the
+// semver of the latest version is sent over the returned channel.
+fn background_check_for_updates() -> mpsc::Receiver<(u32, u32, u32)> {
+    let (sender, receiver) = mpsc::channel();
+
+    let _detached_thread = thread::spawn(move || {
+        // ...
+        if we_found_an_update {
+            sender.send((major, minor, patch));
+        }
+    });
+
+    receiver
+}
+
 /// Run a command with the given logger!
 pub fn run_wasm_pack(command: Command, log: &Logger) -> result::Result<(), failure::Error> {
+    let update_chan = background_check_for_updates();
+
     // Run the correct command based off input and store the result of it so that we can clear
     // the progress bar then return it
     let status = match command {
@@ -126,7 +143,11 @@ pub fn run_wasm_pack(command: Command, log: &Logger) -> result::Result<(), failu
     };
 
     match status {
-        Ok(_) => {}
+        Ok(_) => {
+            if let Ok(update_available) = update_chan.try_recv() {
+                // TODO: print that there is an update available here
+            }
+        }
         Err(ref e) => {
             error!(&log, "{}", e);
             for c in e.iter_chain() {
```

> By unwrapping errors you mean the part where I call to the API? because indeed that's dumb of me...

I mean that *in general*, things like

```rust
something.fallible_operation().unwrap();
```

should become

```rust
something.fallible_operation()?;
```

which will propagate the error instead of panicking.

However, inside the thread spawned in `background_check_for_updates` above, we want to do *neither* of these things. Instead, if we encounter an error (say because of bad network), we just want to log the error, abandon checking for updates, and let the thread finish. This way, if you don't have a network connection, it doesn't prevent you from running `wasm-pack build`.

--------------------------

Hopefully this answers your questions! If anything I wrote is unclear, please ask for clarification :)

daubaris:
Yep, great answer, almost everything is clear, except that part about the file, so I'll wait on others suggestions

alexcrichton:
Until a `~/.wasm-pack` infrastructure is in place the easiest thing to do may be to have a `wasm-pack.stamp` file which lives next to `wasm-pack.exe` perhaps?

alexcrichton:
With https://github.com/rustwasm/wasm-pack/pull/426 a global `Cache` object can be used to maintain a file in a global cache that keeps track of various pieces of metadata and such

daubaris:
Alright, so first of all, sorry for the delay, secondly, some notes on this update. I couldn't really figure out how the global `Cache` object works, so any example regarding it would be great. Next I see my own tendency of using `if let` too much which concerns me a bit. Similarly, I'm in doubt of my architecture decisions - see `src/manifest/mod.rs`. Lastly, I wanted to handle `return_stamp_file_values` function errors instead of just unwrapping as it shouldn't fail, but if does, then we have a problem, so any input on that matter would be also great.

ashleygwilliams:
hey @daubaris ! are you still working on this? sorry for the delay in response. we'd really like to have this for 0.8.0. please let me know if you'd like to continue working on it!

daubaris:
Sure do! have to remember it though. Will complete it asap :+1:

src/build.rs
```diff
@@ -47,6 +48,42 @@ fn rustc_minor_version() -> Option<u32> {
...
+    macro_rules! otry {
+        ($e:expr) => {
+            match $e {
+                Some(e) => e,
+                None => return None,
+            }
+        };
+    }
+
+    let output = otry!(Command::new("wasm-pack").arg("--version").output().ok());
```
chinedufn: Sweet!

But instead of spawning a child process to find this out (which has some overhead) I think we’d want to just include the version the binary at compile time.

So maybe use env!(“CARGO_MANIFEST_VERSION”) here

src/main.rs
```diff
@@ -47,5 +66,10 @@ fn run() -> Result<(), failure::Error> {
 
     let args = Cli::from_args();
     run_wasm_pack(args.cmd)?;
+
+    if let Ok(update_available) = update_available.try_recv() {
+        println!("There's a newer version of wasm-pack available, the new version is: {}, you are using: {}", update_available.1, update_available.0);
```
chinedufn: Maybe link to the releases page and/or suggest cargo install wasm-pack -f

src/build.rs
chinedufn: Would also be more flexible - for example in the (probably uncommon) case that someone is running a wasm-pack binary that isn’t in their $PATH

src/build.rs
```diff
@@ -47,6 +48,42 @@ fn rustc_minor_version() -> Option<u32> {
...
+                Some(latest) => Ok((local, latest)),
+                None => Ok((local, "".to_string()))
+            }
+        },
+        None => bail!("We can't figure out what your wasm-pack version is, make sure the installation path is correct.")
+    }
+}
+
+fn wasm_pack_local_version() -> Option<String> {
+    macro_rules! otry {
```
chinedufn: 🍟 Small fries but the macro feels like overkill to me here 🍟. As I was reading this over I was confused about what otry was until I came and found it.

fitzgen: This is also unnecessary these days, since `?` can work with options, e.g the following compiles:

```rust
pub fn foo(maybe_x: Option<u32>) -> Option<String> {
    let x = maybe_x?;
    Some(x.to_string())
}
```

But yeah, what we should really be doing here is including the version in the binary like `env!("CARGO_MANIFEST_VERSION")` as suggested elsewhere.

src/main.rs
fitzgen: Yes, let's definitely link to https://rustwasm.github.io/wasm-pack/installer/ and maybe even inline the install command directly.

@ashleygwilliams @alexcrichton How stable do we expect the install command to be? Should we just print it directly right here?

src/manifest/mod.rs
```diff
@@ -112,6 +117,116 @@ struct CargoWasmPackProfileWasmBindgen {
...
+    fn return_wasm_pack_file() -> Option<String> {
+        if let Ok(path) = which::which("wasm-pack") {
+            if let Ok(file) = fs::read_to_string(path.with_extension("stamp")) {
+                return Some(file);
+            }
+        }
+        None
+    }
+
+    fn call_for_wasm_pack_version() -> Option<String> {
```
fitzgen: Some doc comments on these functions would be helpful for readers -- I'm not 100% sure what "call for" is supposed to mean in this context.

src/main.rs
alexcrichton: I think for now somewhat unstable enough that we should probably direct to the website

src/manifest/mod.rs
daubaris: Alrighty then, will fix these tommorow! Thanks for the feedback @fitzgen @chinedufn

src/command/build.rs
```diff
@@ -255,6 +255,12 @@ impl Build {
         Ok(())
     }
 
+    /// Returns local and latest wasm-pack versions.
+    pub fn return_wasm_pack_versions() -> Result<(String, String), Error> {
```
drager: I realize I am late to the party but personally I'd like too see this function return a struct that contains `local` and `latest`. Because I think it's not clear for the caller which `String` is which. So maybe something like:

```rust
struct WasmPackVersion {
  local: String,
  latest: String,
}
```

src/build/mod.rs
```diff
@@ -48,6 +49,24 @@ fn rustc_minor_version() -> Option<u32> {
...
+/// Checks and returns local and latest versions of wasm-pack
+pub fn check_wasm_pack_versions() -> Result<(String, String), Error> {
+    match wasm_pack_local_version() {
+        Some(local) => {
+            match Crate::return_wasm_pack_latest_version() {
+                Some(latest) => Ok((local, latest)),
+                None => Ok((local, "".to_string()))
+            }
+        },
+        None => bail!("We can't figure out what your wasm-pack version is, make sure the installation path is correct.")
```
drager: @fitzgen @ashleygwilliams Should we really bail here if we can't determine the user's wasm-pack version? I mean, they can't use the tool at all of this feature fails.

src/main.rs
```diff
@@ -3,17 +3,34 @@ extern crate env_logger;
...
 use wasm_pack::{command::run_wasm_pack, Cli};
 
 mod installer;
 
+fn background_check_for_updates() -> mpsc::Receiver<(String, String)> {
+    let (sender, receiver) = mpsc::channel();
+    let _detached_thread = thread::spawn(move || {
+        if let Ok((local, latest)) = Build::return_wasm_pack_versions() {
+            if !local.is_empty() && !latest.is_empty() && local != latest {
+                sender.send((local, latest)).unwrap();
```
drager: Do we really want to `unwrap()` here? cc @fitzgen @ashleygwilliams

src/manifest/mod.rs
```diff
@@ -113,6 +118,119 @@ struct CargoWasmPackProfileWasmBindgen {
...
+    max_version: String,
+}
+
+impl Crate {
+    /// Returns latest wasm-pack version
+    pub fn return_wasm_pack_latest_version() -> Option<String> {
+        let current_time = chrono::offset::Local::now();
+        Crate::return_wasm_pack_file().and_then(|contents| {
+            let last_updated = Crate::return_stamp_file_value(&contents, "created")
+                .and_then(|t| Some(DateTime::parse_from_str(t.as_str(), "%+").unwrap()));
```
drager: Wouldn't it better to make this function return a `Result` or turn the `parse_from_str(t.as_str(), "%+").unwrap()` into `parse_from_str(t.as_str(), "%+").ok()`? cc @fitzgen @ashleygwilliams

src/manifest/mod.rs
```diff
@@ -113,6 +118,119 @@ struct CargoWasmPackProfileWasmBindgen {
...
+    fn override_stamp_file(current_time: DateTime<offset::Local>, version: &String) {
+        if let Ok(path) = which::which("wasm-pack") {
+            let file = fs::OpenOptions::new()
+                .read(true)
+                .write(true)
+                .append(true)
+                .create(true)
+                .open(path.with_extension("stamp"));
+
+            if let Ok(()) = file.as_ref().unwrap().set_len(0) {
```
drager: Maybe make this function return a `Result` and favor `?` instead of calling `unwrap()`. cc @fitzgen @ashleygwilliams

src/command/build.rs
fitzgen: That seems like a good improvement, but also somethign that could potentially happen in a follow up PR (if that is easier for @daubaris)

src/build/mod.rs
fitzgen: This happens in a background thread, and failure just means that we won't print the potential updates message. It crucially should not prevent any other functionality.

src/main.rs
fitzgen: It doesn't really matter since we don't rely on the background thread to complete successfully.

It may be slightly tidier to do 

```rust
let _ = sender.send((local, latest));
```

to indicate that this is a best-effort attempt to notify the main thread, but it might have won the race to finish the command before the background check finished. But I think the point is fairly academic, given that if the main command finishes, then the main thread is shutting down and the whole process along with it, meaning that this background thread is going to get killed either way.

src/manifest/mod.rs
fitzgen: Yeah, it might be tidier to propagate errors and have the background thread's outermost spawned function be the thing that handles the we-don't-care-about-errors-here behavior in a single place.

src/manifest/mod.rs
```diff
@@ -113,6 +118,119 @@ struct CargoWasmPackProfileWasmBindgen {
...
+
+    fn return_api_call_result(current_time: DateTime<offset::Local>) -> Option<String> {
+        Crate::return_latest_wasm_pack_version().and_then(|v| {
+            Crate::override_stamp_file(current_time, &v);
+            Some(v)
+        })
+    }
+
+    fn override_stamp_file(current_time: DateTime<offset::Local>, version: &String) {
+        if let Ok(path) = which::which("wasm-pack") {
```
fitzgen: Just noticed: this should probably be `std::env::current_exe`. Fine as a follow up PR since it isn't critical, IMO.

src/manifest/mod.rs
fitzgen: Same opinion as above regarding propagating errors to the one central we-don't-care location.

src/command/build.rs
drager: Yeah, I agree.

src/build/mod.rs
drager: Ah! Didn't think of this. Thanks for clarifying!

drager:
@daubaris Thank you so much for all your hard work!

daubaris:
@drager thanks for the help! Sorry for not completing this, real life stuff... 😞 Anyhow, it was great working on this feature, oh and feedback here is 👌

PR #411, opened 2018-10-14
Copy license file(s) to out directory

### LINKED ISSUE #407: wasm-pack build should copy over the license file
Currently, license files aren't copied over. It's not entirely clear to me how best to do this, given that license files can have different names. Perhaps do (the equivalent of) `cp LICENSE* pkg`?

mstallmo:
Closes #407 

Added a new step to `build` that will look in the Cargo.toml file to see if a license key has been set. If one has been set it will glob for `LICENSE*` and copy those file(s) to the specified `out` directory. If a license key has not been set this step will skip with the message `No LICENSE found in Cargo.toml skipping...`. If a key is set in the Cargo.toml but the glob doesn't match any files we will break out of the copying license step and print `License key is set in Cargo.toml but no LICENSE file(s) were found; Please add the LICENSE file(s) to your project directory` to the console.

I would love some feed back on the file globing implementation. I am pretty new to Rust but as far as I can tell the best way to account for the different ways `LICENSE` files can be named was to implement a glob that looks for `LICENSE*` in the crate directory. I thought about using the actual value set in the `License` key in the Cargo.toml but due to the non-standard ways licenses can be referred to I figured it would be best to just check for it's existence rather than relying on the value set for the key itself. This seemed like the most robust way to go about this but I would love to know if there is a better way that I'm not aware of! 

I know it's common in the Rust community to dual license Apache 2.0 and MIT but since that pattern is less common in the JS community I'm not sure if we would want give the ability for users to select which license they want copied if more than one is found. Related, I'm not even sure if it would be desired or appropriate for a user to change the licensing of the wasm-packed output as compared to the Rust source code but it was just a thought I had while implementing this feature.

```bash
$ rustup override set nightly
```

### DIFF
--- a/src/license.rs
+++ b/src/license.rs
@@ -68,7 +68,7 @@ pub fn copy_from_crate(path: &Path, out_dir: &Path, step: &Step) -> Result<(), f
             }
         }
         None => {
-            PBAR.step(step, "No LICENSE found in Cargo.toml skipping...");
+            PBAR.step(step, "No LICENSE found in Cargo.toml, skipping...");
         }
     };
 
--- a/src/license.rs
+++ b/src/license.rs
@@ -1,0 +1,76 @@
+//! Copy `LICENSE` file(s) for the packaged wasm.
+
+use failure;
+use std::fs;
+use std::path::Path;
+
+use emoji;
+use glob::glob;
+use manifest;
+use progressbar::Step;
+use PBAR;
+
+fn get_license(path: &Path) -> Option<String> {
+    match manifest::get_crate_license(path) {
+        Ok(license) => license,
+        Err(_) => None,
+    }
+}
+
+fn glob_license_files(path: &Path) -> Result<Vec<String>, failure::Error> {
+    let mut license_files: Vec<String> = Vec::new();
+    for entry in glob(path.join("LICENSE*").to_str().unwrap())? {
+        match entry {
+            Ok(globed_path) => {
+                license_files.push(String::from(
+                    globed_path.file_name().unwrap().to_str().unwrap(),
+                ));
+            }
+            Err(e) => println!("{:?}", e),
+        }
+    }
+    Ok(license_files)
+}
+
+/// Copy the crate's license into the `pkg` directory.
+pub fn copy_from_crate(path: &Path, out_dir: &Path, step: &Step) -> Result<(), failure::Error> {
+    assert!(
+        fs::metadata(path).ok().map_or(false, |m| m.is_dir()),
+        "crate directory should exist"
+    );
+
+    assert!(
+        fs::metadata(&out_dir).ok().map_or(false, |m| m.is_dir()),
+        "crate's pkg directory should exist"
+    );
+
+    match get_license(path) {
+        Some(_) => {
+            let msg = format!("{}Copying over your LICENSE...", emoji::DANCERS);
+            PBAR.step(step, &msg);
+            let license_files = glob_license_files(path);
+
+            match license_files {
+                Ok(files) => {
+                    if files.len() == 0 {
+                        PBAR.info("License key is set in Cargo.toml but no LICENSE file(s) were found; Please add the LICENSE file(s) to your project directory");
+                        return Ok(());
+                    }
+                    for license_file in files {
+                        let crate_license_path = path.join(&license_file);
+                        let new_license_path = out_dir.join(&license_file);
+                        if let Err(_) = fs::copy(&crate_license_path, &new_license_path) {
+                            PBAR.info("origin crate has no LICENSE");
+                        }
+                    }
+                }
+                Err(_) => PBAR.info("origin crate has no LICENSE"),
+            }
+        }
+        None => {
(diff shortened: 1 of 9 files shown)


src/license.rs
```diff
@@ -0,0 +1,76 @@
...
+fn glob_license_files(path: &Path) -> Result<Vec<String>, failure::Error> {
+    let mut license_files: Vec<String> = Vec::new();
+    for entry in glob(path.join("LICENSE*").to_str().unwrap())? {
+        match entry {
+            Ok(globed_path) => {
+                license_files.push(String::from(
+                    globed_path.file_name().unwrap().to_str().unwrap(),
+                ));
+            }
+            Err(e) => println!("{:?}", e),
```
fitzgen: Instead of matching on the entry and ignoring errors, we should do something like this:

```rust
let path = entry?;
```

src/license.rs
```diff
@@ -0,0 +1,76 @@
...
+use PBAR;
+
+fn get_license(path: &Path) -> Option<String> {
+    match manifest::get_crate_license(path) {
+        Ok(license) => license,
+        Err(_) => None,
+    }
+}
+
+fn glob_license_files(path: &Path) -> Result<Vec<String>, failure::Error> {
```
fitzgen: This should return a `Result<Vec<PathBuf>, failure::Error>` instead of `Result<Vec<String>, failure::Error>`.

src/license.rs
```diff
@@ -0,0 +1,76 @@
...
+                Ok(files) => {
+                    if files.len() == 0 {
+                        PBAR.info("License key is set in Cargo.toml but no LICENSE file(s) were found; Please add the LICENSE file(s) to your project directory");
+                        return Ok(());
+                    }
+                    for license_file in files {
+                        let crate_license_path = path.join(&license_file);
+                        let new_license_path = out_dir.join(&license_file);
+                        if let Err(_) = fs::copy(&crate_license_path, &new_license_path) {
+                            PBAR.info("origin crate has no LICENSE");
```
fitzgen: I don't think that this error means that there is no license file. But either way, we should propagate the error here with `?` instead of swallowing it.

src/license.rs
```diff
@@ -0,0 +1,76 @@
...
+                        }
+                    }
+                }
+                Err(_) => PBAR.info("origin crate has no LICENSE"),
```
fitzgen: Again: we should propagate errors, not swallow them.

src/license.rs
mstallmo: Updated to propagate errors if any happen!

src/license.rs
mstallmo: Updated to return PathBuf instead of String!

src/license.rs
mstallmo: I've updated this to use `?` as well so we get the error instead of printing the message!

src/license.rs
mstallmo: Updated to propagate errors instead of printing out message!

fitzgen:
Urg, this is hitting the cargo bug on windows too.

mstallmo:
Just rebased and pushed. Should be good to go shortly!

src/license.rs
```diff
@@ -0,0 +1,76 @@
...
+            }
+        }
+        None => {
+            PBAR.step(step, "No LICENSE found in Cargo.toml skipping...");
```
ashleygwilliams: we should have some punctuation here between, e.g. ""No LICENSE found in `Cargo.toml`, skipping...""

tests/all/utils/fixture.rs
```diff
@@ -74,6 +74,44 @@ impl Fixture {
...
 
+    /// Add `WTFPL LICENSE` file to the fixture.
+    pub fn wtfpl_license(&self) -> &Self {
+        self.file(
+            "LICENSE-WTFPL",
+            r#"
+                DO WHAT THE FUCK YOU WANT TO PUBLIC LICENSE
+                    Version 2, December 2004
+
+                Copyright (C) 2004 Sam Hocevar <sam@hocevar.net>
```
ashleygwilliams: can we just remove the text of this license? i know it's not something you personally wrote and is the text of the license but it's something i just don't want in this code base. feel free to replace it with a neutral sentence or just a copyright line.

mstallmo: Yeah no problem, I was on the fence about this to begin with I'll go ahead and reword it!

ashleygwilliams: thank you!

src/license.rs
mstallmo: Just updated to include a comma to the No license found message!

tests/all/utils/fixture.rs
mstallmo: I changed all the lines containing this wording to use`WHATEVER` instead!

mstallmo:
@ashleygwilliams thanks for the merge! I'll start on the refactor for this over the next couple of weeks. I'm off work for the holidays so I'll have a some good free time to focus on OSS stuff!

PR #415, opened 2018-10-15
refactor: Move binary installation to it's own crate

### LINKED ISSUE #384: refactor binary installation
i think we'll want to use this for inclusion of cargo-generate #373 and tbh it needs it's own tests, etc. this issue will get all the files associated with binary installation generalized and in it's own dir. moving it completely out of this project can come later

drager:
This PR intends to fix #384.


drager:
@ashleygwilliams Status is that I'm having some problems with cargo workspaces for the tests. Running `cargo test` results in some failures because of the workspaces. I get the following output that makes the tests fail: `current package believes it's in a workspace when it's not.`

However, if I remove the [workspace] section the tests pass...

I'm hoping to resolve this as soon as possible and then add some tests for the functions in the `binary_installation` crate.

drager:
@ashleygwilliams: I have now resolved the issue with workspaces and I added some tests for the path functions as well.

I suppose you could review the code I have pushed now. However, I think some more tests could be added. Please let me know what you think. :blush:

fitzgen:
@drager heads up: if you rebase, then appveyor CI should be fixed :+1:

drager:
@fitzgen: Thank you! :blush:

Cargo.toml
```diff
@@ -35,6 +39,7 @@ tar = "0.4.16"
 toml = "0.4"
 which = "2.0.0"
 zip = "0.4.2"
```
alexcrichton: I think some crates in this list (like `zip` here) may no longer be necessary, mind pruning this of unneeded crates?

Cargo.toml
```diff
@@ -9,6 +9,10 @@ readme = "README.md"
 categories = ["wasm"]
 documentation = "https://rustwasm.github.io/wasm-pack/"
 
+[workspace]
+members = ["binary-install"]
```
alexcrichton: Since there's a dependency on `binary-install` I think this line can be removed

binary-install/Cargo.toml
```diff
@@ -0,0 +1,17 @@
+[package]
+name = "binary-install"
```
alexcrichton: This crate will need to be published to crates.io to publish `wasm-pack` itself to crates.io, so could this perhaps be named `wasm-pack-binary-install` to namespace it?

binary-install/Cargo.toml
```diff
@@ -0,0 +1,17 @@
...
+[dependencies]
+curl = "0.4.13"
+dirs = "1.0.4"
+failure = "0.1.2"
+flate2 = "1.0.2"
+hex = "0.3"
+siphasher = "0.2.3"
+slog = "2.3"
+slog-term = "2.4"
+slog-async = "2.3"
```
alexcrichton: I think not all of these dependencies (like this one) may be needed here, mind pruning this list of those which aren't needed?

Cargo.toml
drager: Ah, my bad. Will fix that!

Cargo.toml
drager: Ah yes!

binary-install/Cargo.toml
drager: Sounds great, will rename it.

binary-install/Cargo.toml
drager: I will remove them.

ashleygwilliams:
ack- i triaged in the wrong order today and looks like they are back :( i'm sorry, i'll be sure to get this in today

drager:
No worries, give me a few minutes and I will resolve them again.

ashleygwilliams:
🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉

PR #420, opened 2018-10-25
[WIP] Add support for missing README files and other README file extensions

mstallmo:
Based on feedback from @fitzgen on #411 I went back and updated the copying of the README.md file to propagate the error as well if one occurs. I had referenced that functionality in writing the license copy so I figured it would be good to go back and update that one as well.

---

```bash
$ rustup override set nightly
```

### DIFF
--- a/src/readme.rs
+++ b/src/readme.rs
@@ -23,8 +23,6 @@ pub fn copy_from_crate(path: &Path, out_dir: &Path, step: &Step) -> Result<(), f
     PBAR.step(step, &msg);
     let crate_readme_path = path.join("README.md");
     let new_readme_path = out_dir.join("README.md");
-    if let Err(_) = fs::copy(&crate_readme_path, &new_readme_path) {
-        PBAR.warn("origin crate has no README");
-    };
+    fs::copy(&crate_readme_path, &new_readme_path)?;
     Ok(())
 }


mstallmo:
Build is failing due to `cargo fmt` but it looks like that's coming from master. I fixed the formatting in #411 but figured I should hold off doing it here again to avoid things getting too confusing.

fitzgen:
Sure thing -- sounds good to me! I updated the title of the PR to reflect this change in direction.

ashleygwilliams:
one thing to share- we should not error if the readme is missing. only warn.

ashleygwilliams:
After reading the associated issue I want to warn that npm does not support these Readme files so i'm not entirely sure that we should? I am curious why you thought this was a bug @fitzgen

fitzgen:
I had no idea npm doesn't support non-md READMEs! This was motivated by github's support for other kinds of READMEs, which means that some projects use that support. It is perhaps worth reconsidering if npm doesn't allow it, though!

mstallmo:
Sorry for the delay but I have everything ready to be pushed up and reviewed now. Is this something we still want or should we close this for now since NPM doesn't support anything except for .md files?

ashleygwilliams:
for the moment i think we should close this as npm doesn't support other README file extensions. if there's a usecase for supporting other README file extensions, and a way to make sure folks aren't confused about npm not supporting the other formats, i'm open to hearing it!

Issue #427, opened 2018-11-04
Allow the user to fully customize the package name


stristr:
## 💡 Feature description
Currently, the NPM package name generated by `wasm-pack` is the (optionally scoped) name of the Rust crate being packed. There is utility in being able to choose the name of the package; for example, I might want to publish multiple packages to the NPM repository using different `--target` options.

#### 💻 Basic example
I propose a command line flag, `--package-name`, for this purpose, which should work in conjunction with `--scope`.

For example:

`wasm-pack build examples/js-hello-world --scope test --package-name custom-name`

should result in an NPM package named `@test/custom-name`.

fitzgen:
While I agree that we want to allow this functionality, I don't think adding a new flag is *quite* the fix we want.

https://github.com/rustwasm/rfcs/pull/4 will likely end up allowing crates to write package.json snippets that are then merged into the package.json that wasm-pack creates. I think using this package.json merging would be the right way to customize the package name.

In the meantime, one can always edit the `pkg/package.json` manually.

stristr:
Thanks for the response @fitzgen! I wanted to explain a bit more about the use case I imagine, because I don't think snippet-merging would help here and runtime renaming is actually preferable.

1. I am building a Rust crate, `bar`. It is meant to be used anywhere I can use a Rust crate (e.g. I can build a native binary, I can compile it into an iOS or Android app, etc.).
2. Separately, I would like to use `wasm-bindgen` to publish _multiple different_ WebAssembly packages for different targets. For example, I might want to publish in the `@foo` scope twice, once with `--target=browser` (`@foo/bar`) and once with `--target=commonjs` (`@foo/bar-node`).
3. Finally and perhaps most crucially, I don't want to deviate from standard tooling in order to manually edit `package.json` for these two different use cases.

As a basic example of how a runtime option would let me have it both ways: the excellent [`@wasm-tool/wasm-pack-plugin`](https://www.npmjs.com/package/@wasm-tool/wasm-pack-plugin) could be used like this, so that one `webpack` build produces both desired outputs:

```
module.exports = {
  plugins: [
    new WasmPackPlugin({
      crateDirectory: path.resolve(__dirname, "crate"),
      extraArgs: "--scope=foo",
    }),
    new WasmPackPlugin({
      crateDirectory: path.resolve(__dirname, "crate"),
      extraArgs: "--scope=foo --package-name=bar-node --out-dir=pkg-node",
    }), 
  ],
};
```

[...]

Issue #313, opened 2018-09-19
give option target=all which generates pkg supporting commonjs and esm


ashleygwilliams:
groundwork in #312 once merged, i'll start on it. in the meantime- if we want to bikeshed the target argument, let's

stristr:
Hey @ashleygwilliams, is this up for grabs, or underway? Happy to help out.

zrzka:
@ashleygwilliams seems that #312 is merged. Do you have any plans to fix this?

This would be really awesome. Then I'll be able to trash the following script, which packs for both targets and then merges these two packages into one.

```bash
#!/usr/bin/env bash

set -e

NAME="balena_temen"

if [ -d "pkg" ]; then
    rm -rf pkg
fi

if [ -d "pkg-node" ]; then
    rm -rf pkg-node
fi

wasm-pack build -t nodejs -d pkg-node
wasm-pack build -t browser -d pkg
cp "pkg-node/${NAME}.js" "pkg/${NAME}_main.js"
sed "s/require[\(]'\.\/$NAME/require\('\.\/${NAME}_main/" "pkg-node/${NAME}_bg.js" > "pkg/${NAME}_bg.js"
jq ".files += [\"${NAME}_bg.js\"]" pkg/package.json \
    | jq ".main = \"${NAME}_main.js\"" > pkg/temp.json
mv pkg/temp.json pkg/package.json
rm -rf pkg-node
```

drager:
I should be able to start working on this.

ashleygwilliams:
from #469, @KSXGitHub:

When I write an NPM package, I rarely think of bundlers as they can comprehend `require()` syntax, so as long as I only use `require()` to load module, I don't worry about bundlers. This is not the case with Node.js WASM modules generated by `wasm-bindgen` as it uses `fs.readFileSync` (which cannot be polyfilled by bundlers) to load `.wasm` file. 

`wasm-pack build` currently supports 2 targets:
  * `--target browser` support bundlers (today's browsers cannot load the output script/module directly) but not Node.js.
  * `--target nodejs` support Node.js runtime but not bundlers/browsers.

This is inconvenient, if I want to support both targets, I have to write my own wrapper that calls `wasm-bindgen` twice.

## Suggestion

* Without options, generate 4 files for 4 targets:
  * `index.bundler.js` for JavaScript bundlers, set `"browser"` field in `package.json` to point to this file.
[...]

PR #428, opened 2018-11-04
Add optional --package-name argument to override package name.

stristr:
This PR attempts to address #427 by introducing an optional `--package-name` build-time flag.


stristr:
Per discussion in #427, closing in favor of finishing #313.

PR #434, opened 2018-11-05
Replace `slog` with `log`

### LINKED ISSUE #425: Avoid the need to pass `&Logger` internally everywhere
I've noticed when working on `wasm-pack` that `&Logger` from `slog` is passed around in a lot of locations. This unfortunately makes refactoring painful, however, and is otherwise somewhat unergonomic to pass in so man locations.

Perhaps the `log` crate or some other solution could be used where we still get to log everything but `&Logger` doesn't have to be passed as an argument?

alexcrichton:
This commit replaces the `slog` family of crates used by `wasm-pack`
with the `log` crate plus `env_logger`. This also means that by default
`wasm-pack` also won't create a `wasm-pack.log` file in the current
directory. Enabling logging will now be done through
`RUST_LOG=wasm_pack` instead of `-v` flags.

Closes #425


otavio:
I am looking for a logging solution and were looking at this repository as an example of the use of `slog`.

Even though I noticed you are not using structured logging, I wonder if it doesn't make sense to look at it. Considering `wasm-pack` library, the use of `slog` opens some nice opportunities to use custom `Drain` if embedding it inside another tool or service.

I may be completely mistaken here but would like to know your opinion about it. Mind to share your view on this?

huangjj27:
Hi @otavio, I think wasm-pack is used for development scene rather than embedded production. So, I think it’s ok not to use logger with embed feature.

alexcrichton:
I've rebased this on master, should be good to go now

huangjj27:
AppVeyor fails on a test base on `wasm-pack.log` which now is not generated by default. Add some params to the test to generate the log.

ashleygwilliams:
closing in favor of https://github.com/rustwasm/wasm-pack/pull/491

PR #436, opened 2018-11-06
Replace internal `Error` with `failure::Error`

alexcrichton:
This commit removes the internal `Error` type in favor of exclusively
using the `failure::Error` type, simplifying error construction in a
number of locations and reducing the number of error types in play.


src/child.rs
```diff
@@ -171,6 +170,28 @@ pub fn run(
...
         return Ok(stdout);
     } else {
-        return Err(Error::cli(command_name, stdout.into(), stderr.into(), exit).into());
+        let mut msg = String::new();
+        if stdout.len() > 0 {
+            msg.push_str("\nstdout:\n");
+            for line in stdout.lines() {
+                msg.push_str("  ");
+                msg.push_str(line);
+                msg.push_str("\n");
```
fitzgen: This is going to re-create the double-output thing, since the output is already re-printed as the command is run.

alexcrichton: Oops, indeed!

PR #440, opened 2018-11-06
Add three build profiles and infrastructure for their toml config

### LINKED ISSUE #153: create flags for the rustwasm working group's three "build modes"
the rust wasm working group is consider 3 modes of building a wasm project as seen here: https://github.com/rustwasm/rust_wasm_template/issues/6

this seems highly related to #127 

are we interested in making a flag for each of these? i think it seems like a reasonable idea. would love to hear all ya'lls thoughts!

### LINKED ISSUE #160: Add some kind of toml configuration

fitzgen:
Fixes #153
Fixes #160

### DIFF
--- a/docs/src/cargo-toml-configuration.md
+++ b/docs/src/cargo-toml-configuration.md
@@ -1,0 +1,29 @@
+# `Cargo.toml` Configuration
+
+`wasm-pack` can be configured via the `package.metadata.wasm-pack` key in
+`Cargo.toml`. Every option has a default, and is not required.
+
+There are three profiles: `dev`, `profiling`, and `release`. These correspond to
+the `--dev`, `--profiling`, and `--release` flags passed to `wasm-pack build`.
+
+The available configuration options and their default values are shown below:
+
+```toml
+[package.metadata.wasm-pack.profile.dev.wasm-bindgen]
+# Should we enable wasm-bindgen's debug assertions in its generated JS glue?
+debug-js-glue = true
+# Should wasm-bindgen demangle the symbols in the "name" custom section?
+demangle-name-section = true
+# Should we emit the DWARF debug info custom sections?
+dwarf-debug-info = false
+
+[package.metadata.wasm-pack.profile.profiling.wasm-bindgen]
+debug-js-glue = false
+demangle-name-section = true
+dwarf-debug-info = false
+
+[package.metadata.wasm-pack.profile.release.wasm-bindgen]
+debug-js-glue = false
+demangle-name-section = true
+dwarf-debug-info = false
+```
--- a/src/bindgen.rs
+++ b/src/bindgen.rs
@@ -2,6 +2,7 @@
 
 use binaries::{Cache, Download};
 use child;
+use command::build::BuildProfile;
 use emoji;
 use failure::{self, ResultExt};
 use manifest::CrateData;
@@ -145,14 +146,17 @@ pub fn wasm_bindgen_build(
     out_dir: &Path,
     disable_dts: bool,
     target: &str,
-    debug: bool,
+    profile: BuildProfile,
     step: &Step,
     log: &Logger,
 ) -> Result<(), failure::Error> {
     let msg = format!("{}Running WASM-bindgen...", emoji::RUNNER);
     PBAR.step(step, &msg);
 
-    let release_or_debug = if debug { "debug" } else { "release" };
+    let release_or_debug = match profile {
+        BuildProfile::Release | BuildProfile::Profiling => "release",
+        BuildProfile::Dev => "debug",
+    };
 
     let out_dir = out_dir.to_str().unwrap();
 
@@ -181,8 +185,15 @@ pub fn wasm_bindgen_build(
         .arg(dts_arg)
         .arg(target_arg);
 
-    if debug {
+    let profile = data.configured_profile(profile);
+    if profile.wasm_bindgen_debug_js_glue() {
         cmd.arg("--debug");
+    }
+    if !profile.wasm_bindgen_demangle_name_section() {
+        cmd.arg("--no-demangle");
+    }
+    if profile.wasm_bindgen_dwarf_debug_info() {
+        cmd.arg("--keep-debug");
     }
 
     child::run(log, cmd, "wasm-bindgen").context("Running the wasm-bindgen CLI")?;
--- a/src/build.rs
+++ b/src/build.rs
@@ -1,6 +1,7 @@
 //! Building a Rust crate into a `.wasm` binary.
 
 use child;
+use command::build::BuildProfile;
 use emoji;
 use failure::{Error, ResultExt};
 use progressbar::Step;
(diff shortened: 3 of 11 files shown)


src/bindgen.rs
```diff
@@ -181,7 +185,7 @@ pub fn wasm_bindgen_build(
         .arg(dts_arg)
         .arg(target_arg);
 
-    if debug {
+    if profile.debug_info_is_enabled() {
```
alexcrichton: This may not be quite the right question to ask here because `--debug` may run the risk of significantly slowing down the bindings. (in that this is enabled for profiling mode I think?)

src/manifest/mod.rs
```diff
@@ -26,6 +26,10 @@ pub struct CrateData {
 #[derive(Deserialize)]
 struct CargoManifest {
     package: CargoPackage,
+
+    #[serde(default)]
```
fitzgen: I think I might need to put this on all the members below too, so that if you just define a profiling config and not a dev and release config, it doesn't break deserializing...

src/build.rs
```diff
@@ -62,13 +63,30 @@ pub fn rustup_add_wasm_target(log: &Logger, step: &Step) -> Result<(), Error> {
...
     PBAR.step(step, &msg);
     let mut cmd = Command::new("cargo");
     cmd.current_dir(path).arg("build").arg("--lib");
-    if !debug {
-        cmd.arg("--release");
+    match profile {
+        BuildProfile::Profiling => {
+            // Force enable debug info, because builds that use the release
+            // cargo profile disables debug info.
+            cmd.env("RUSTFLAGS", "-g");
```
alexcrichton: I think we should not do this for the time being because it doesn't actually do anything useful in the short-term (nothing reads the dwarf info) and otherwise the names section is always present. 

I think eventually we'll want to do this, but for now it may be better to get cache hits for builds in release/profiling mode

src/build.rs
```diff
@@ -62,13 +63,30 @@ pub fn rustup_add_wasm_target(log: &Logger, step: &Step) -> Result<(), Error> {
...
+            cmd.arg("--release");
+        }
+        BuildProfile::Release => {
+            cmd.arg("--release");
+        }
+        BuildProfile::Dev => {
+            // Plain cargo builds use the dev cargo profile, which includes
+            // debug info by default.
```
alexcrichton: This reminds me that we really should add `--dev` to Cargo...

src/manifest/mod.rs
```diff
@@ -36,6 +40,63 @@ struct CargoPackage {
     repository: Option<String>,
 }
 
+#[derive(Default, Deserialize)]
+struct CargoMetadata {
+    #[serde(rename = "wasm-pack")]
```
alexcrichton: I think `serde(default)` may need to be here and all lower fields as well, right? Currently if you have `[package.metadata.foo]` that may fail to deserialize as well as just `[package.metadata.wasm-pack.profile.dev]` (and not other profiles).

`package.metadata.wasm-pack.profile.dev` is also a bit of a mouthful, so perhaps we can drop the `profile` name as well?

src/manifest/mod.rs
```diff
@@ -36,6 +40,63 @@ struct CargoPackage {
...
+            dev,
+            release,
+            profiling,
+        }
+    }
+}
+
+#[derive(Deserialize)]
+struct CargoWasmPackProfile {
+    // This is where wasm-opt, wasm-snip, etc configuration will go.
```
alexcrichton: As a proof of concept, could configuration for `wasm-bindgen` be added? Should allow passing flags like `--no-demangle` and `--keep-debug`

fitzgen:
One of the new tests added depends on this: https://github.com/rustwasm/wasm-bindgen/pull/1020

src/manifest/mod.rs
fitzgen: I guess it comes down to whether we will ever have non profile related things here. It seems probable to me that we will (eg the JS env target) and so I don't think we should get rid of the `profile` path segment.

src/build.rs
fitzgen: Yep, that would be nice.

fitzgen:
This is ready for another round of review.

docs/src/cargo-toml-configuration.md
```diff
@@ -0,0 +1,29 @@
...
+The available configuration options and their default values are shown below:
+
+```toml
+[metadata.wasm-pack.profile.dev.wasm-bindgen]
+# Should we enable wasm-bindgen's debug assertions in its generated JS glue?
+debug = true
+# Should wasm-bindgen demangle the symbols in the "names" custom section?
+demangle = true
+# Should we keep the DWARF debug info custom sections?
+keep-debug = false
```
alexcrichton: I'm not a huge fan of the name added to `wasm-bindgen` here, maybe we could be a bit more descriptive here with something like:

```toml
# Whether or not the DWARF debug info custom sections are preserved
dwarf-debug = true
```

Is the default `false` here because they're somewhat non-operational today?

src/manifest/mod.rs
```diff
@@ -36,6 +42,120 @@ struct CargoPackage {
...
+#[derive(Default, Deserialize)]
+struct CargoWasmPackProfiles {
+    #[serde(default)]
+    dev: CargoWasmPackProfile<Dev>,
+
+    #[serde(default)]
+    release: CargoWasmPackProfile<Release>,
+
+    #[serde(default)]
+    profiling: CargoWasmPackProfile<Profiling>,
```
alexcrichton: The trait-based approach here is a little difficult to follow, and it may run the risk of getting unwieldy as we add more and more configuration.

Could the fields of `CargoWasmPackProfile` perhaps all be `Option<T>` and the defaults are applied later when we have more context?

alexcrichton: I'm basically a little worried of getting to trapped up in trying to shoehorn all the default/deserialization logic into serde. I'm not much of a serde expert by any means, though, and Cargo's way of doing this (the one thing I have experience with) is pretty wordy and verbose as well. 

One option might be do to something like having all of these fields be `CargoWasmPackProfile` but they're all tagged with `#[serde(deserialize_with = "local_function_$default")]`. In `local_function_*` it would deserialize a `CargoWasmPackProfileWithOptions` which deserializes an struct with a bunch of options in it, returning a `CargoWasmPackProfile` which doesn't have any options and uses `unwrap_or` everywhere. 

(that may also be a good way to deal with cascading options that configure others if they're not otherwise configured too)

docs/src/cargo-toml-configuration.md
fitzgen: > Is the default `false` here because they're somewhat non-operational today?

It was true for `--dev` builds in the last iteration of the PR and you requested it be turned off because nothing was consuming it yet :-p

alexcrichton: Oh lol I can't seem to make up my mind...

Sounds like I'm fine with whatever, although we may wish to have a small comment in the documentation indicating that nothing known consumes this and the info itself is probably wrong given what wasm-bindgen does to the binary.

I think I actually like the idea of leaving it off-by-default

fitzgen:
Ok, I am pretty happy with serde and defaults stuff now. Final round of review?

src/manifest/mod.rs
```diff
@@ -34,6 +36,157 @@ struct CargoPackage {
...
+            }
+        }
+        d!(wasm_bindgen.debug_js_glue);
+        d!(wasm_bindgen.demangle_name_section);
+        d!(wasm_bindgen.dwarf_debug_info);
+    }
+
+    /// Get this profile's configured `[wasm-bindgen.debug-js-glue]` value.
+    pub fn wasm_bindgen_debug_js_glue(&self) -> bool {
+        self.wasm_bindgen.debug_js_glue.unwrap()
```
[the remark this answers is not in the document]
fitzgen: I leaned back towards them because making the two-struct approach work with nested structs involves a lot of boilerplate and it isn't totally clear how we would know which defaults to use in the nested struct deserialization where we no longer know which profile we are deserializing for.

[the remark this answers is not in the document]
alexcrichton: Ok! Just wanted to confirm. We can always tweak over time if necessary!

fitzgen:
> Also just to make sure I don't forget, we should file an issue afterwards to handle unknown keys to issue warnings on things like typos. Cargo uses https://crates.io/crates/serde_ignored and wasm-pack can probably use it as well!

Filed https://github.com/rustwasm/wasm-pack/issues/443

PR #442, opened 2018-11-07
Don't require cdylib crate-type for testing

alexcrichton:
The `cdylib` crate type output isn't actually necessary for the
`wasm-pack test` stage because `wasm-bindgen` isn't run over a wasm
file. This commit removes the checks during `wasm-pack test` that the
`cdylib` crate type is configured.


chinedufn:
I just came here to open this issue since it was preventing me from migrating to wasm-pack test... sweet! Will wait for `0.6.0`!

PR #446, opened 2018-11-12
feat: Catch typos in Cargo.toml and warn about them

### LINKED ISSUE #443: Use the serde_ignored crate to catch typos in our toml config
https://crates.io/crates/serde_ignored

If someone specifies `wasm_pack` instead of `wasm-pack` in their toml (for example) then we can use this crate to detect that and emit a warning.

drager:
This intends to fix #443.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,13 +1,13 @@
 [package]
-name = "wasm-pack"
-description = "pack up the wasm and publish it to npm!"
-version = "0.5.1"
 authors = ["Ashley Williams <ashley666ashley@gmail.com>"]
-repository = "https://github.com/ashleygwilliams/wasm-pack.git"
-license = "MIT/Apache-2.0"
-readme = "README.md"
 categories = ["wasm"]
+description = "pack up the wasm and publish it to npm!"
 documentation = "https://rustwasm.github.io/wasm-pack/"
+license = "MIT/Apache-2.0"
+name = "wasm-pack"
+readme = "README.md"
+repository = "https://github.com/ashleygwilliams/wasm-pack.git"
+version = "0.5.1"
 
 [dependencies]
 atty = "0.2.11"
@@ -21,7 +21,6 @@ hex = "0.3"
 human-panic = "1.0.1"
 indicatif = "0.9.0"
 lazy_static = "1.1.0"
-openssl = { version = '0.10.11', optional = true }
 parking_lot = "0.6"
 serde = "1.0.74"
 serde_derive = "1.0.74"
@@ -29,16 +28,21 @@ serde_ignored = "0.0.4"
 serde_json = "1.0.26"
 siphasher = "0.2.3"
 slog = "2.3"
-slog-term = "2.4"
 slog-async = "2.3"
+slog-term = "2.4"
+strsim = "0.8.0"
 structopt = "0.2"
 tar = "0.4.16"
 toml = "0.4"
 which = "2.0.0"
 zip = "0.4.2"
 
+[dependencies.openssl]
+optional = true
+version = "0.10.11"
+
 [dev-dependencies]
 tempfile = "3"
 
 [features]
-vendored-openssl = ['openssl/vendored']
+vendored-openssl = ["openssl/vendored"]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -25,6 +25,7 @@ openssl = { version = '0.10.11', optional = true }
 parking_lot = "0.6"
 serde = "1.0.74"
 serde_derive = "1.0.74"
+serde_ignored = "0.0.4"
 serde_json = "1.0.26"
 siphasher = "0.2.3"
 slog = "2.3"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,13 +1,13 @@
 [package]
-authors = ["Ashley Williams <ashley666ashley@gmail.com>"]
-categories = ["wasm"]
-description = "pack up the wasm and publish it to npm!"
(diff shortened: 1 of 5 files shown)


src/manifest/mod.rs
```diff
@@ -210,8 +211,25 @@ impl CrateData {
...
-        let manifest: CargoManifest = toml::from_str(&manifest)
-            .with_context(|_| format!("failed to parse manifest: {}", manifest_path.display()))?;
+        let manifest = &mut toml::Deserializer::new(&manifest);
+
+        let mut unused_keys = BTreeSet::new();
+
+        let manifest: CargoManifest = serde_ignored::deserialize(manifest, |path| {
+            let path_string = path.to_string();
+
+            if path_string.contains("metadata") {
```
alexcrichton: This may want to be `metadata.wasm-pack` perhaps? That way we won't accidentally warn about other `metadata` sections accidentally

src/manifest/mod.rs
```diff
@@ -210,8 +211,25 @@ impl CrateData {
...
+                unused_keys.insert(path_string);
+            }
+        })
+        .with_context(|_| format!("failed to parse manifest: {}", manifest_path.display()))?;
+
+        unused_keys.iter().for_each(|path| {
+            PBAR.warn(&format!(
+                "\"{}\" is misspelled and will be ignored. Please check your Cargo.toml.",
```
alexcrichton: This may want to be reworded a bit because we don't actually know if it's misspelled, perhaps this could just say that it's an ignored/unknown key?

src/manifest/mod.rs
drager: Oh right! That makes sense. Thanks!

src/manifest/mod.rs
drager: Yes, I was somewhat unsure what to put there and "unknown and will be ignored" sounds good.

src/manifest/mod.rs
drager: @alexcrichton: Hmm, if I do `metadata.wasm-pack` we wouldn't warn about misspelled `wasm-pack`. i.e. if someone writes `[package.metadata.wasmpack.profile.dev.wasm-bindgen]`.

alexcrichton: That's true, but I personally would think that we can't land this while it warns about `[package.metadata.some-unrelated-tool]`. 

We can possibly do a lehvenstein distance thing (typically used for "did you typo this?") to figure out if it looks like `wasm-pack` specifically, but other crates/tools using `[package.metadata]` should not be warned about when using `wasm-pack`

drager: Sounds reasonable! Pushed some code for this.

Cargo.toml
```diff
@@ -1,13 +1,13 @@
...
-license = "MIT/Apache-2.0"
-readme = "README.md"
 categories = ["wasm"]
+description = "pack up the wasm and publish it to npm!"
 documentation = "https://rustwasm.github.io/wasm-pack/"
+license = "MIT/Apache-2.0"
+name = "wasm-pack"
+readme = "README.md"
+repository = "https://github.com/ashleygwilliams/wasm-pack.git"
+version = "0.5.1"
```
alexcrichton: It looks like some unrelated changes snuck in here?

Cargo.toml
```diff
@@ -21,23 +21,28 @@ hex = "0.3"
...
+strsim = "0.8.0"
 structopt = "0.2"
 tar = "0.4.16"
 toml = "0.4"
 which = "2.0.0"
 zip = "0.4.2"
 
+[dependencies.openssl]
+optional = true
+version = "0.10.11"
```
alexcrichton: I think the movement here may be an unrelated change?

src/manifest/mod.rs
```diff
@@ -241,6 +253,59 @@ impl CrateData {
...
+        let manifest = &mut toml::Deserializer::new(&manifest);
+
+        let mut unused_keys = BTreeSet::new();
+        let mut has_possible_typo = false;
+        let levenshtein_threshold = 1;
+
+        let manifest: CargoManifest = serde_ignored::deserialize(manifest, |path| {
+            let path_string = path.to_string();
+
+            if levenshtein(WASM_PACK_METADATA_KEY, &path_string) == levenshtein_threshold {
```
alexcrichton: Could this perhaps have some more guards like testing that the key starts with `package.metadata` and then just running levenshtein over the key afterwards, ideally only matching "wasm-pack" up against the next part of the key?

Also, I think this may perhaps want to be `<= threshold` to ensure that exact matches are warned about

src/manifest/mod.rs
```diff
@@ -241,6 +253,59 @@ impl CrateData {
...
+                has_possible_typo = true;
```
alexcrichton: I think it's still good to warn about all the keys here, so it's fine for this to take the same path as the one below where it just throws a new key in the set to get warned about, that way the exact key that wasn't used is printed as well and the user doesn't have to guess

src/manifest/mod.rs
drager: Sounds good! Ah, ofc it should be `<=`.

src/manifest/mod.rs
drager: Yes, that makes sense. Much better to print the key so the user doesn't have to guess.

alexcrichton:
Code-wise looks great to me, thanks!

drager:
Awesome! Thank you for the guidance and feedback!

PR #450, opened 2018-11-18
update zip to 0.5.0

srijs:
Updating to the latest version of the `zip` crate. This version also brings a bunch of performance improvements which might come in handy.


xmclark:
Something else to note about this upgrade: the deflate backed was swapped out for a pure-rust implementation (no libc). So this brings wasm-pack one step closer to compiling to WebAssembly! Ha!

PR #453, opened 2018-11-29
Generate self-.gitignore as part of pkg folder

RReverser:
As per discussion with @ashleygwilliams in https://twitter.com/ag_dubs/status/1068173944807727110, `pkg` should be commonly ignored, and the simplest way to do that is to just generate nested `.gitignore` inside it ignoring everything (including `.gitignore` itself).

---

```bash
$ rustup override set nightly
```

### DIFF
--- a/src/command/utils.rs
+++ b/src/command/utils.rs
@@ -18,6 +18,7 @@ pub fn create_pkg_dir(out_dir: &Path, step: &Step) -> Result<(), failure::Error>
     let msg = format!("{}Creating a pkg directory...", emoji::FOLDER);
     PBAR.step(step, &msg);
     fs::create_dir_all(&out_dir)?;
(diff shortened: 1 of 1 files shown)


ashleygwilliams:
@RReverser thanks so much for this and so sorry for the delay in reviewing! could you add a small blurb in the documentation and/or README to explain that this directory is now ignored by default?

we may need to wait on landing this til i implement a way to edit package.json and add js or other assets to the pkg, but i imagine that should be for the release directly after the next one (which is happening this week)

RReverser:
Thanks! Sorry - as mentioned in the other thread, I initially sent this from a different laptop and then with holidays and getting sick didn't get around to finishing this.

ashleygwilliams:
it's all good, no worries at all <3 thanks for you efforts!

PR #461, opened 2018-12-16
add an option to pass an arbitrary set of arguments to cargo build

### LINKED ISSUE #455: Allow passing extra command-line flags to cargo
## 💡 Feature description
I'm often building my packages on CI with `--locked` / `--frozen` after the dependencies have been downloaded.

I'd like to do the same with `wasm-pack` but currently there is no way to pass extra args to underlying `cargo` command.

#### 💻 Basic example

I think it would make sense to be able to pass extra arbitrary args after `--` as many other wrappers do, so that one could run:

```bash
$ wasm-pack build wasm-folder -- --frozen
```

torkve:
Fixes #455.
Example: `wasm-pack build -- --features "feature-a feature-b"`

```bash
$ rustup override set nightly
```


ashleygwilliams:
@torkve thanks so much for this! would you mind adding a test? and potentially an example to the documention? thanks so much!

torkve:
I've added a test and a bit of docs, however English is not my native language, so feel free to fix it.

torkve:
Are there any news with this PR? :-)

ashleygwilliams:
hey @torkve - thank you for this and sorry for the lack of and delay of reply. i'm currently a little blocked on getting 0.6.0 out. my only concern with this PR is that we may also want to pass extra unsupported arguments to wasm-bindgen and this interface poses a bit of a conflict with that. if it were gated by a flag (perhaps `--cargo-args`) that might work better? i need to perhaps think on exactly what that interface should be. does that concern make any sense? if you have thoughts i'd be happy to hear them!

torkve:
I suppose you are correct from the POV that options should be similar. `--cargo-args` and `--wasm-bindgen-args` would be more intuitive than my implementation.
The reason I wrote it this very way, is the special additional quotation required otherwise. For example, two ways to specify feature set:
```
$ wasm-pack ... -- --features "feature-a feature-b"
$ wasm-pack ... --cargo-args "--features 'feature-a feature-b'"
```
Second variant is not pretty enough. And if one would need to escape some characters, it'll become even more complicated.

While currently I do not see any possible complex values in wasm-bindgen arguments, I think it would not be the best option to make it look just like `wasm-pack ... --wasm-bindgen-args "..." -- --cargo-args...`

May be every wasm-bindgen option just should be exposed in wasm-pack?

danwilhelm:
Would setting env vars to pass arguments possibly be a solution? That might make it somewhat consistent with cargo, since `RUSTFLAGS` could be set as well to send some flags to rustc.

Perhaps @alexcrichton might have ideas? I believe he contributed to passing arguments to rustc from cargo.

alexcrichton:
I'd personally advocate for a design that:

* Like this PR, accepts arbitrary arguments after `--` to pass to Cargo
* Adds most first-class options from Cargo to `wasm-pack build`, like `--features`, `--test`, etc.
* Also accepts arbitrary arguments to `wasm-bindgen`, but as first-class options to `wasm-pack`. The `wasm-bindgen` tool is basically an internal implementation detail of `wasm-pack`, whereas `cargo` is a first-class feature. I'd much rather prominently support Cargo than `wasm-bindgen`

PR #478, opened 2018-12-29
Refactor: fix clippy warnings

### LINKED ISSUE #477: Fix clippy warnings in the codebase
While working on a different refactor I ran clippy on the codebase and noticed that it gave a handful of warnings about various parts of the code. We should probably refactor the flagged code to be in line with the clippy lints. 

Along this line are there any clippy lints that we want to ignore/diverge from or should we follow all of the linting rules?

mstallmo:
closes #477 

This is a first pass PR at cleaning up the clippy warnings. Most of them were simple enough refactors that it only took minor modifications to the code to get rid of the warnings. There were a few warnings, marked by `#![allow()]`, that would require more significant changes to the code structure so instead of going ahead and implementing those changes I marked them as clippy allow until we can decide if the changes are something we want or not.  

---

```bash
```

### DIFF
--- a/clippy.toml
+++ b/clippy.toml
@@ -1,0 +1,1 @@
+too-many-arguments-threshold = 8
--- a/src/child.rs
+++ b/src/child.rs
@@ -41,7 +41,7 @@ pub fn new_command(program: &str) -> Command {
 /// given sender.
 fn read_and_send<R, F>(
     mut reader: R,
-    sender: mpsc::Sender<OutputFragment>,
+    sender: &mpsc::Sender<OutputFragment>,
     mut map: F,
 ) -> io::Result<()>
 where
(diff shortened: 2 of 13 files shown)


mstallmo:
Awesome! I should be able to fix up the conflicts this evening so everything will be ready to merge.

I'll also go ahead and file a new issue for working on removing the `allows `!

ashleygwilliams:
@mstallmo i might be trying to get this into 0.6.0- would you mind if i took over and fixed the conflicts? i am just feeling guilty and don't want to delay the release any further <3

mstallmo:
@ashleygwilliams Sorry for the delayed response but that sounds great!

mstallmo:
Just pushed the update resolving the merge conflicts with master so this should be good to go as soon as the checks pass.

ashleygwilliams:
i need to do #509 before this because it's more urgent- which *may* introduce another conflict- but i'll handle that for you! thanks so much for your work here!

ashleygwilliams:
since we opted against a point release, we can get this into the next release (0.7.0) thanks!

Issue #480, opened 2019-01-03
Cargo command calls don't output color to the terminal


otaviopace:
## 💡 Feature description
When `wasm-pack` runs `cargo build` for example, under the hood, the output of it has no color.

#### 💻 Basic example
In the image down below, this part should have color just like using `cargo build` directly on the terminal.
![screenshot from 2019-01-03 19-55-23](https://user-images.githubusercontent.com/15306309/50663813-0bda2b80-0f92-11e9-81a7-0616466695b1.png)

Part without color:
```shell
  [4/9] Compiling to WASM...
      Finished release [optimized] target(s) in 0.04s
```

I honestly would like to have color because it eases to understand errors on the terminal :slightly_smiling_face: 

Example of error without color:

![screenshot from 2019-01-03 20-55-41](https://user-images.githubusercontent.com/15306309/50666016-097bcf80-0f9a-11e9-822c-6a3270bb0b6f.png)

Example with color:

![screenshot from 2019-01-03 20-57-06](https://user-images.githubusercontent.com/15306309/50666029-1ac4dc00-0f9a-11e9-9380-4f84f2417d76.png)

ibaryshnikov:
is it still an issue? @otaviopace @ashleygwilliams 
wasm-pack output for current version (0.8.1) is colored enough, it looks like it was fixed in one of the releases

otaviopace:
Honestly I don't know because I don't have that computer anymore 😕 . Maybe we can close the issue, and later if I or anyone suffers from it, we can re open it.

ashleygwilliams:
closing- sorry to hear about your computer @otaviopace ! please file a new issue if this is still a problem.

Issue #298, opened 2018-09-11
Mini RFC: wasm-pack should use progressive enhancement in its output


fitzgen:
## Summary

`wasm-pack` should default to plain text output and use progressive enhancement to add bells and whistles such as colors, bold, and progress bars when the functionality is available.

## Motivation

`wasm-pack` should work seamlessly with all terminals on all platforms, and when its output is piped to another process or a log file. Users should never see garbled characters or color codes.

## Details

By default, `wasm-pack` should emit only plain text output to `stdout`.

We should use [the `atty` crate](https://crates.io/crates/atty) and the `$TERM` environment variable to determine whether `stdout` is a fully featured terminal. If `atty::is(atty::Stream::Stdout)` is true and the `$TERM` environment variable is not set to `"dumb"`, then `wasm-pack` should also emit colors, bold styles, and show progress bars when it makes sense.

To enforce consistency with these rules across the code base, there should be one central place that manages whether colors are printed or progress bars are displayed. All other code should be able to assume that the colors are used and progress bars are displayed, and be none the wiser if that is not actually the case under the covers. Luckily, we already have most of the printing logic isolated into the `progressbar` module. The exception is some `format!`ed string messages used to construct some `Error` instances.

To move forward, we should create a new `terminal` module that encapsulates all terminal output, progressive enhancement, and feature detection.

1. For styling text, we should refactor all existing usage of `style(...).bold().dim()` etc to use our own `terminal` utility function(s) that only apply the `style(...)` when the terminal supports it.

2. For progress bars, we should have a utility in `terminal` that creates the progress bars for the caller. If `stdout` is not a tty or `$TERM` is `"dumb"`, then it should return a progress bar created with `indicatif::ProgressBar::hidden`, which is a no-op progress bar.

## Related Issues

* https://github.com/rustwasm/wasm-pack/issues/284

## References

* https://crates.io/crates/atty
* https://en.wikipedia.org/wiki/Computer_terminal#Dumb_terminals

fitzgen:
There was some discussion of this at todays WG meeting. Consensus was that where possible, we should push this into upstream crates, but also generally do wahtever it takes to get the behavior described here at the end of the day.

killercup:
Started some implementation work at <https://github.com/killercup/output-rs>. Very much WIP and subject to rapid change.

alexcrichton:
[...]

PR #481, opened 2019-01-03
🌈 cargo: make command output with color

otaviopace:
All `cargo` calls were without color because
they were missing the `--color` flag with the
`always` value.

Examples of the terminal with color :rainbow: 
- Success
screenshot from 2019-01-03 19-58-12

- Error
screenshot from 2019-01-03 20-57-06

```bash
```

This pull request closes this issue: https://github.com/rustwasm/wasm-pack/issues/480

### DIFF
--- a/src/bindgen.rs
+++ b/src/bindgen.rs
@@ -130,7 +130,9 @@ pub fn cargo_install_wasm_bindgen(
         .arg("--version")
         .arg(version)
         .arg("--root")
-        .arg(&tmp);
+        .arg(&tmp)
+        .arg("--color")
+        .arg("always");
 
     child::run(logger, cmd, "cargo install").context("Installing wasm-bindgen with cargo")?;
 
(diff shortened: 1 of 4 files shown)


alexcrichton:
Unfortunately this is very caught up in https://github.com/rust-lang/rust/pull/57337 and doesn't actually fix the problem on Windows. I would personally not recommend merging this PR because it only fixes unix platforms (and even there, not completely because there's no way to turn colors off), and it doesn't add colors for Windows.

otaviopace:
@alexcrichton , I can look how to fix the issue on `Windows` too, I think I can do it :slightly_smiling_face:

alexcrichton:
Er sorry I had the wrong link there, the right issue this is caught up in is https://github.com/rustwasm/wasm-pack/issues/298. @otaviopace this is solvable for sure, but I think that a good solution would be pretty architecturally invasive and requires more consensus before pushing it through.

otaviopace:
Ok... So should we close this?

ashleygwilliams:
@otaviopace yeah i think so- thank you so much for your efforts (and amazing emoji usage)- lemme know if there's anything else you'd like to work on! would be happy to help you get started :)

otaviopace:
Great, I will definetly see if I can help with something :smile:

PR #484, opened 2019-01-10
validate target argument

### LINKED ISSUE #483: Terminate on unknown wasm-bindgen target
## 🐛 Bug description
```
$ wasm-pack build --target=node
```
build works but then I found out that it is still a browser target.

To my surprise, the following works and defaults to browser target.
```
$ wasm-pack build --target=random
```

#### 🤔 Expected Behavior

`wasm-pack` fail right away showing the available targets and display unknown target supplied.

#### 👟 Steps to reproduce

Just `wasm-pack build --target=insert_here`.

#### 🌍 Your environment
wasm-pack version: wasm-pack 0.5.1
rustc version: rustc 1.31.1 (b6c32da9b 2018-12-18)

csmoe:
Closes #483

### DIFF
--- a/src/command/build.rs
+++ b/src/command/build.rs
@@ -88,7 +88,7 @@ pub struct BuildOptions {
     pub scope: Option<String>,
 
     #[structopt(long = "mode", short = "m", default_value = "normal")]
-    /// Sets steps to be run. [possible values: no-install, normal]
+    /// Sets steps to be run. [possible values: no-install, normal, force]
     pub mode: BuildMode,
 
     #[structopt(long = "no-typescript")]
@@ -157,6 +157,12 @@ impl Build {
             // functionality yet, so we have to implement it ourselves.
             _ => bail!("Can only supply one of the --dev, --release, or --profiling flags"),
         };
+
+        // `possible_values` in clap isn't supported by `structopt`
+        let possible_targets = ["browser", "nodejs", "no-modules"];
+        if !possible_targets.contains(&build_opts.target.as_str()) {
+            bail!("Supported targets: browser, nodejs, no-modules");
+        }
 
         Ok(Build {
             crate_path,


danwilhelm:
This is a very helpful addition! A few quick things I noticed:

+ `test --mode` duplicates the `build --mode` parameter. So, consider updating it in the same way here -- https://github.com/rustwasm/wasm-pack/blob/master/src/command/test.rs#L73

+ Consider also suggesting valid values for [no-install, normal, force] by amending the existing error message here -- https://github.com/rustwasm/wasm-pack/blob/master/src/command/build.rs#L62

csmoe:
@danwilhelm checking for buildmode was already there: https://github.com/rustwasm/wasm-pack/blob/8af40029a98a76c9033430923bf00e1ce935ef3f/src/command/build.rs#L55-L65

danwilhelm:
Sorry I was unclear! 

In my comment, the first link points to the mode parameter in TestOptions. I believe the default value and help string there should mirror the one changed in this PR (ie remove the default value "normal" and add "force" to the help string). 

In the second link, as you point out, testing for Build Mode is already there. The Issue this PR addresses was that when an invalid argument is passed, the valid options are not displayed. Instead of bailing with 'Unknown build option', it could indicate that valid options are no-install, normal, and force.

ashleygwilliams:
this seems to have a bunch of failures on appveyor-- a large number that seems disproportionate to the change here:

```
failures:
    bindgen::can_download_prebuilt_wasm_bindgen
    build::build_different_profiles
    build::it_format_out_dir_on_windows
    build::it_should_build_crates_in_a_workspace
    build::it_should_build_js_hello_world_example
    build::it_should_build_nested_project_with_transitive_dependencies
    build::renamed_crate_name_works
    test::cdylib_not_required
    test::complains_about_missing_wasm_bindgen_test_dependency
    test::it_can_find_a_webdriver_on_path
    test::it_can_run_browser_tests
    test::it_can_run_failing_tests
    test::it_can_run_node_tests
    test::it_can_run_tests_with_different_wbg_test_and_wbg_versions
    test::it_requires_node_or_a_browser
    test::renamed_crate_name_works
    test::the_headless_flag_requires_a_browser
test result: FAILED. 31 passed; 17 failed; 0 ignored; 0 measured; 0 filtered out
```

@csmoe do you have a windows computer to test this locally? if not perhaps @steveklabnik ? i would be surprised if this was all from this incredibly small change...

csmoe:
@ashleygwilliams I have only one linux machine but i'll recheck the failed test cases.

ashleygwilliams:
i'm gonna restart the build- perhaps it was a weird flake on appveyors part

ashleygwilliams:
looks like it was a fluke! i think this is good to go and we can merge.

Issue #486, opened 2019-01-11
Empty input prompt when 'login'


danwilhelm:
## 🐛 Bug description
`wasm-pack login` gives the appearance that the program is hanging or not working. It does not display stdout/stderr to the user. Instead of asking for a username, the user only sees the cursor.

#### 🤔 Expected Behavior
`wasm-pack login` spawns `npm adduser` and should behave identically. The user should see the following line-by-line interactive prompt:

```
Username:
Password:
Email: (this IS public)
```

Errors should also be displayed. For example, if an email address is entered not containing '@', the user should see "npm WARN Email must be an email address" followed by a repeated Email prompt.

#### 👟 Steps to reproduce
1. In a [non-Windows OS](https://github.com/rustwasm/wasm-pack/issues/277), run: `wasm-pack login`
2. Now, the program will indefinitely wait for input without prompting the user what to do.
3. After Ctrl-C is pressed, buffered output is displayed to the screen.

#### 🌍 Your environment
wasm-pack version: 0.5.1 (also fails on latest master branch)
rustc version: rustc 1.33.0-nightly (8e2063d02 2019-01-07)

#### 🔧 Technical explanation
`wasm-pack login` spawns [`npm adduser`](https://docs.npmjs.com/cli/adduser), which displays an interactive prompt, asking the user for username/password/email. [PR #392](https://github.com/rustwasm/wasm-pack/pull/392) inadvertantly replaced the interactive process spawner with `child::run`, which is hard-coded to buffer stdout/stderr. This caused `login` to become essentially unusable; the user can no longer see interactive input prompts or error messages.

The above can be tested directly:
1. In a [non-Windows OS](https://github.com/rustwasm/wasm-pack/issues/277), run: `wasm-pack login`
2. Enter `asdf` and press Enter. (Username)
3. Enter `asdf` and press Enter. (Password)
4. Enter `asdf@asdf.com` and press Enter. (Email)
5. Because the child process now is complete, all buffered output should be displayed to the screen.

ashleygwilliams:
closed by #487

PR #487, opened 2019-01-11
Login issues

danwilhelm:
1. Fixes Issue #486: Empty input prompt when `login`.
2. Fixes `login` arguments being accidentally merged together when spawning `npm adduser`.

See the text of each commit for more details.

### DIFF
--- a/src/npm.rs
+++ b/src/npm.rs
@@ -62,12 +62,14 @@ pub fn npm_login(
         args.push_str(&format!(" --auth_type={}", auth_type));
     }
 
+    // Interactively ask user for npm login info.
+    //  (child::run does not support interactive input)
     let mut cmd = Command::new("npm");
-    cmd.arg("login")
-        .arg(args)
-        .stdin(Stdio::inherit())
-        .stdout(Stdio::inherit());
-    child::run(log, cmd, "npm login")
-        .with_context(|_| format!("Login to registry {} failed", registry))?;
-    Ok(())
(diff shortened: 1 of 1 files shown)


danwilhelm:
@fitzgen It would be nice to keep all of the process code in child.rs. So alternatively, I could perhaps add a `child::run_interactive` to child.rs? I was hesitant to add a flag to `child::run`, since it is already a bit complex and seems wholly designed to accumulate the outputs.

Note the solution I proposed here does not log `npm adduser` outputs and the user's interactive responses, which was one intention of `child::run`. However, from experience in another language, this might be difficult to do while still keeping it interactive and bug-free.

fitzgen:
> @fitzgen It would be nice to keep all of the process code in child.rs. So alternatively, I could perhaps add a `child::run_interactive` to child.rs? I was hesitant to add a flag to `child::run`, since it is already a bit complex and seems wholly designed to accumulate the outputs.

I'm not sure what `child::run_interactive` would do that the normal `std` doesn't already. `child::run` is all about making sure that we log child output correctly, but running interactively kind of side steps that, so I don't think there is much to be done.

If I'm missing something, please file a follow up issue with suggestions and we can take the discussion there! :)

danwilhelm:
I agree with both of you, which is why I implemented it as-is! Indeed, a new function `child::run_interactive` would do nothing more than call the normal `std` function.

PR #491, opened 2019-01-14
Alexcrichton log

ashleygwilliams:

```bash
```


huangjj27:
Does that mean I need to rewrite the test？

ashleygwilliams:
hey @huangjj27 - i think no, at least for the moment- the way we're doing things with logs now i think it'd be pretty hard to test this. we definitely need a better solution for testing CLI output, but there's some foundational work to do that before we can rewrite this test. i'm also sad about losing it but i do think this logging strategy is the right way forward, and so perhaps for the next release we can sort a way to more robustly test the tool's output. sorry and thank you so much for your diligence!

huangjj27:
@ashleygwilliams Ok. What else can I help?

PR #495, opened 2019-01-15
get binary install ready for publish

ashleygwilliams:


.gitignore
```diff
@@ -6,3 +6,4 @@ wasm-pack.log
 /build-installer
 docs/book
 docs/installer
+binary-install/Cargo.lock
```
fitzgen: This should be part of the wasm-pack workspace, and therefore shouldn't have its own lock file.

Instead, we should do this:

```toml
# wasm-pack/Cargo.toml

[workspace]
members = [
    "./binary-install",
]
```

PR #502, opened 2019-01-16
chore(travis): Run clippy check

drager:
When #478 gets merged I think we should run `clippy` at Travis.


ashleygwilliams:
blocked on merge by #477

ashleygwilliams:
since we opted against a point release we can get this into the next release (0.7.0)

Issue #503, opened 2019-01-16
error don't panic during wasm-bindgen binary installation for `wasm-pack test`


fitzgen:
## `stdout`

<details>

```
~/dodrio
$ wasm-pack test --firefox --headless
  
| [1/6] Checking `rustc` version...
  [1/6] Checking `rustc` version...
| [2/6] Adding WASM target...
  [2/6] Adding WASM target...
| info: component 'rust-std' for target 'wasm32-unknown-unknown' is up to date
  info: component 'rust-std' for target 'wasm32-unknown-unknown' is up to date
| [3/6] Compiling tests to WASM...
  [3/6] Compiling tests to WASM...
|     Finished dev [unoptimized + debuginfo] target(s) in 0.10s
      Finished dev [unoptimized + debuginfo] target(s) in 0.10s
| wasm-bindgen 0.2.31 (41eefa742)
  wasm-bindgen 0.2.31 (41eefa742)
| [4/6] Installing wasm-bindgen...
\ [4/6] Installing wasm-bindgen...
  [4/6] Installing wasm-bindgen...
|     Updating crates.io index
\     Updating crates.io index
-     Updating crates.io index
/     Updating crates.io index
|     Updating crates.io index
\     Updating crates.io index
-     Updating crates.io index
      Updating crates.io index
|   Installing wasm-bindgen-cli v0.2.32
\   Installing wasm-bindgen-cli v0.2.32
    Installing wasm-bindgen-cli v0.2.32
|  Downloading crates ...
\  Downloading crates ...
-  Downloading crates ...
/  Downloading crates ...
   Downloading crates ...
|   Downloaded wasm-bindgen-cli-support v0.2.32
\   Downloaded wasm-bindgen-cli-support v0.2.32
    Downloaded wasm-bindgen-cli-support v0.2.32
|   Downloaded wasm-bindgen-threads-xform v0.2.32
    Downloaded wasm-bindgen-threads-xform v0.2.32
|   Downloaded wasm-bindgen-wasm-interpreter v0.2.32
    Downloaded wasm-bindgen-wasm-interpreter v0.2.32
|   Downloaded wasm-bindgen-gc v0.2.32
    Downloaded wasm-bindgen-gc v0.2.32
|    Compiling semver-parser v0.7.0
     Compiling semver-parser v0.7.0
|    Compiling autocfg v0.1.1
     Compiling autocfg v0.1.1
|    Compiling version_check v0.1.5
[...]

PR #504, opened 2019-01-16
More logging and assertions

fitzgen:
Some logging and assertions to help us figure out what is going on with #503 when a `cargo install` isn't working right.

I can't repro #503 anymore, but this should help us if we ever see it in the wild again :-/


ashleygwilliams:
looks like CI wants the lockfile updated :)

fitzgen:
The assertion was slightly off too -- `Download` represents the tarball/zip, not the binary itself. Fixed in latest force-pushed commit.

fitzgen:
Assuming CI comes back green, this should be ready.

Issue #508, opened 2019-01-17
Cache hits even if `wasm-bindgen` binary is missing


danwilhelm:
## 🐛 Bug description

**This bug affects some v0.6.0 users who will upgrade to v0.6.1, including #497 and #503.**

The `wasm-bindgen` cache falsely reports a hit when the cache directory exists, but the binary does not.

This causes `wasm-pack` to fail on all subsequent builds requiring a `cargo install` (at least until the next `wasm-bindgen` version is released!).

#### 🤔 Expected Behavior
When the `wasm-bindgen` binary is missing, a new cache should be created that overwrites the old one.

#### 👟 Steps to reproduce
Occurs when all four are present:
1. `wasm-bindgen` is not in `$PATH`.
2. A prebuilt binary is not available (e.g. no Internet, [no Windows build](https://github.com/rustwasm/wasm-bindgen/releases/tag/0.2.32)).
3. `wasm-bindgen` has been installed previously to the cache.
4. The `wasm-bindgen` binary in the cache has been removed.

---
#### MAC reproduction (Linux is similar)

In any WASM project:
1. `cargo uninstall wasm-bindgen` (removes `wasm-bindgen` from `$PATH`)
2. `cargo build` (creates the lockfile)
3. `rm -rf ~/Library/Caches/.wasm_pack` (removes existing cache)
4. `mkdir -p ~/Library/Caches/.wasm_pack/wasm-bindgen-cargo-install-0.2.32` (creates a fake cargo install without a binary)
5. Disconnect from the Internet (makes the pre-built binary unavailable)
6. `wasm-pack build` <-- Use master

Should fail with:
> `Error: /Users/<username>/Library/Caches/.wasm_pack/wasm-bindgen-cargo-install-0.2.32/wasm-bindgen binary does not exist`

---
#### WINDOWS reproduction (easier!)

In any WASM project:
1. `cargo uninstall wasm-bindgen` (removes `wasm-bindgen` from `$PATH`)
2. `rm -rf ~\AppData\Local\.wasm_pack` (removes existing cache)
2. `wasm-pack build` <-- Use v0.6.0 (should fail at end)
3. `wasm-pack build` <-- Use master

[...]

PR #509, opened 2019-01-17
Test for cache `wasm-bindgen` binary, not directory

danwilhelm:
Issue #508

1. Tests the cache for the `wasm-bindgen` binary. Previously, only
tested for the directory.

2. Removes the `wasm-bindgen` cache directory if it already exists,
prior to renaming the temporary cache directory.

Verified fix works correctly via the reproduction steps in #508 on both Windows and Mac.

### DIFF
--- a/src/bindgen.rs
+++ b/src/bindgen.rs
@@ -109,13 +109,15 @@ pub fn cargo_install_wasm_bindgen(
 
     let dirname = format!("wasm-bindgen-cargo-install-{}", version);
     let destination = cache.join(dirname.as_ref());
-    if destination.exists() {
+    let destination_dl = Download::at(&destination);
+
+    if let Ok(bindgen_path) = destination_dl.binary("wasm-bindgen") {
         debug!(
             "`cargo install`ed `wasm-bindgen={}` already exists at {}",
             version,
-            destination.display()
+            bindgen_path.display()
         );
-        return Ok(Download::at(&destination));
+        return Ok(destination_dl);
     }
 
     if !install_permitted {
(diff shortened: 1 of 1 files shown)


src/bindgen.rs
```diff
@@ -109,13 +109,15 @@ pub fn cargo_install_wasm_bindgen(
 
     let dirname = format!("wasm-bindgen-cargo-install-{}", version);
     let destination = cache.join(dirname.as_ref());
-    if destination.exists() {
+    let destination_dl = Download::at(&destination);
+
+    if let Ok(bindgen_path) = destination_dl.binary("wasm-bindgen") {
```
fitzgen: Nice.

src/bindgen.rs
```diff
@@ -163,10 +165,21 @@ pub fn cargo_install_wasm_bindgen(
         })?;
     }
 
+    // Remove destination directory if already exists,
+    // e.g. if only the bindgen binary does not exist.
+    if destination.exists() {
+        fs::remove_dir_all(&destination).with_context(|_| {
```
fitzgen: Is this necessary? The rename doesn't overwrite the old directory anyways?

alexcrichton:
I'm not sure I fully understand the bug here, can you help explain it a bit more? The intention was that if this directory exists then that signifies a successful installation, so this may mean that something with installation is awry or setting something up too soon perhaps?

ashleygwilliams:
closing in favor of https://github.com/rustwasm/wasm-pack/pull/510 which is gonna include some refactors and unit tests

Issue #447, opened 2018-11-17
Not reusing cache in test


huangjj27:
## 🐛 Bug description
when I run the "test build::it_should_build_js_hello_world_example", I notice that it reinstall wasm-bindgen for cargo while it had downloaded  prebuilt wasm-bindgen.

In fact, in the test it had installed the `wasm-bindgen` and the `wasm-bindgen-test-runner`  binaries so it should just copy them to the fixture directory, and using them, no need to re install
```
#[test]
fn it_should_build_js_hello_world_example() {
    let fixture = utils::fixture::js_hello_world();
    fixture.install_local_wasm_bindgen();   // we installed first before begin the test!
    let cli = Cli::from_iter_safe(vec![
        "wasm-pack",
        "build",
        &fixture.path.display().to_string(),
    ])
    .unwrap();
    fixture.run(cli.cmd).unwrap();
}

```
#### 🤔 Expected Behavior
It should copy the local wasm-bindgen in test_cache

#### 👟 Steps to reproduce
1. cargo clean
2. click the single test in vs code of `it_should_build_js_hello_world_example`, so that you can see the log
3. wait for compilation. It should show the log as follow:
```
running 1 test
Created fixture at c:\Users\34937\Documents\wasm-pack\target\t\.tmp6BHDQ1\wasm-pack

  [1/9] Checking `rustc` version...
  [2/9] Checking crate configuration...
  [3/9] Adding WASM target...
  info: component 'rust-std' for target 'wasm32-unknown-unknown' is up to date
  [4/9] Compiling to WASM...
      Updating `git://mirrors.ustc.edu.cn/crates.io-index` index
     Compiling proc-macro2 v0.4.20
     Compiling unicode-xid v0.1.0
     Compiling serde v1.0.80
     Compiling ryu v0.2.6
     Compiling wasm-bindgen-shared v0.2.21
     Compiling itoa v0.4.3
     Compiling cfg-if v0.1.6
     Compiling lazy_static v1.2.0
     Compiling wasm-bindgen v0.2.21
     Compiling log v0.4.6
     Compiling quote v0.6.9
|    Compiling syn v0.15.18
test build::it_should_build_js_hello_world_example ... test build::it_should_build_js_hello_world_example has been running for over 60 seco     Compiling syn v0.15.18
     Compiling serde_json v1.0.32
     Compiling serde_derive v1.0.80
     Compiling wasm-bindgen-backend v0.2.21
[...]

PR #510, opened 2019-01-17
[WIP] fix: cargo install fallback cache bug

ashleygwilliams:


alexcrichton:
Out of curiosity, why is this split from https://github.com/rustwasm/wasm-pack/pull/509? (it makes it somewhat confusing knowing where to discuss it)

ashleygwilliams:
it scoops up #509 i just hadn't closed that

alexcrichton:
Ah ok, I'll copy a comment from over there then 

---

I'm not sure I fully understand the bug here, can you help explain it a bit more? The intention was that if this directory exists then that signifies a successful installation, so this may mean that something with installation is awry or setting something up too soon perhaps?

ashleygwilliams:
yeah @alexcrichton so if i'm understanding correctly i think the directory is created before we are sure the installation worked- so you may end up with an empty directory? i'm working through it now on the refactor but that's my sense so far.

alexcrichton:
I believe that's the bug in that case, the design of this cache was that if it exists in the final location it's guaranteed to be valid. If the final location is created before it's valid though, that's the bug to fix

fitzgen:
> I'm not sure I fully understand the bug here, can you help explain it a bit more?

The original bug was that we were creating cache directories with an unexpected layout. This was what I fixed with the renaming the binaries before moving the temp dir into the cache.

There was another bug, where if someone had one of the old, bad layout cache directories from a previous, buggy version of wasm-pack, the new version of wasm-pack didn't detect and fix that.

Orthogonal to those things, at some point in time a "." was added to the cache directory name, when there didn't used to be, and there were some mixups surrounding that that I think resulted in some kind of bug, rather than just a cache miss. I haven't investigated this bit very much, so it is a little fuzzy in my head.

alexcrichton:
Ah ok, makes sense! If we do another release of wasm-bindgen though that'll use new cache directories, and I think it means this PR isn't necessary?

ashleygwilliams:
@alexcrichton this PR still has a fair number of refactors i'd like to keep

also re: the dot, i am certain that was my mistake here: https://github.com/rustwasm/wasm-pack/pull/495/files#diff-4737405cfca0ca1539a2248e9614954cR40 - it should not have the dot?

src/bindgen.rs
```diff
@@ -167,21 +147,46 @@ pub fn cargo_install_wasm_bindgen(
...
-    fs::rename(&tmp, &destination)?;
+    fs::rename(&tmp_dirname, &cargo_install_cache_dirname)?;
 
     Ok(destination_dl)
 }
 
+// Run `cargo install` to a temporary location to handle ctrl-c gracefully
+// and ensure we don't accidentally use stale files in the future
+fn cargo_install_to_tmp_dir(bin_dirname: &str, cache_root_dirname: &PathBuf, version: &str) -> Result<PathBuf, failure::Error> {
+    let tmp = cache_root_dirname.join(format!(".{}", bin_dirname));
```
fitzgen: I don't think we want to use a non-`tempfile::TempDir` temporary directory. This opens the possibility for concurrency bugs (`wasm-pack build` in two different directories stepping on each others' toes). The existing method of `tempfile::TempDir` followed by an atomic rename doesn't suffer from that class of bugs.

We can use a `tempfile::TempDir` and pull this `cargo install` into its own function if the helper function takes a reference to the temp dir.

danwilhelm:
@alexcrichton Aside from the v0.6.0 bug, my reasoning is that if `wasm-pack` only relies on the binary being there, less can go wrong in the future if the binary is specifically tested for.

Similarly, if the cache directory already exists then the [move/rename will definitely fail](https://doc.rust-lang.org/std/fs/fn.rename.html). So, why not try to remove the cache directory first before renaming it? That seems it would only prevent a fatal error at no additional cost.

src/bindgen.rs
danwilhelm: I almost implemented this in my PR. In addition to your reasons, I am also queasy about having such a large (temporary) directory name begin with a '.'. I feel like this might not be supported on some (especially older) OSs.

danwilhelm:
@ashleygwilliams I would personally remove the dot from the cache directory name. On Windows and Mac, none of the other (100+) directories in the Cache folder start with a dot, so it really stands out.

Also, I'm on the verge of misremembering when to write `wasm-pack` versus `wasm_pack`. It's now `RUST_LOG=wasm_pack`, the cache directory is `.wasm_pack`, but the crate name and exe are `wasm-pack`.

ashleygwilliams:
i hear ya @danwilhelm i think we should try to stick with `wasm-pack` - i can fix the cache names here https://github.com/rustwasm/wasm-pack/blob/0acc7b79df4a34ad897bc583d30509a2800b838b/src/command/build.rs#L183 (they were both `wasm_pack` but the comment has a typo) and i can see if the log level can be changed

ashleygwilliams:
looks like test is failing on windows:

```
---- build::build_with_and_without_wasm_bindgen_debug stdout ----
Created fixture at C:\projects\wasm-pack-071k0\target\t\.tmpAQxWmX\wasm-pack
   [INFO]: License key is set in Cargo.toml but no LICENSE file(s) were found; Please add the LICENSE file(s) to your project directory
thread 'build::build_with_and_without_wasm_bindgen_debug' panicked at 'called `Result::unwrap()` on an `Err` value: Custom { kind: NotFound, error: PathError { path: "C:\\projects\\wasm-pack-071k0\\target\\test_cache\\wasm-bindgen-cargo-install-0.2.32\\.wasm-bindgen-cargo-install-0.2.32\\.tmpjvEWCa", err: Os { code: 3, kind: NotFound, message: "The system cannot find the path specified." } } }
```

danwilhelm:
It is crashing on:
```
    let tmp = TempDir::new_in(cache_root_dirname.join(format!(".{}", bin_dirname)))?;
    drop(fs::remove_dir_all(&tmp));
```

^^^ Looks like the temp directory is accidentally being appended to: `{cache_root_dirname}\.{bin_dirname}`.

I believe you can get rid of `bin_dirname` as a parameter to `cargo_install_to_tmp_dir`, since the binaries were moved to the root?

ashleygwilliams:
currently having @steveklabnik confirm that this actually *fails* on windows machines

steveklabnik:
I get this

```text
    Installing C:\Users\steve\AppData\Local\Temp\.tmpffYCSB\bin\wasm-bindgen-test-runner.exe
    Installing C:\Users\steve\AppData\Local\Temp\.tmpffYCSB\bin\wasm2es6js.exe
    Installing C:\Users\steve\AppData\Local\Temp\.tmpffYCSB\bin\wasm-bindgen.exe
| warning: be sure to add `C:\Users\steve\AppData\Local\Temp\.tmpffYCSB\bin` to your PATH to be able to run the installed binaries
Error: The system cannot find the path specified. (os error 3)
error: process didn't exit successfully: `target\debug\wasm-pack.exe build ..\lol\` (exit code: 1)
```

huangjj27:
Is this related to #447?

ashleygwilliams:
@huangjj27 no it shouldn't be- it's just that if you are requesting a binary of wasm-bindgen and that binary doesn't exist, the cargo install fallback will fail. the tests shouldn't be using the fallback (in fact they don't! which is why we didn't catch this bug)

drager:
@ashleygwilliams What's the status of this PR? Is this bug still present?

ashleygwilliams:
i think theres a chance this bug is still here but it's been so long i'm not sure. going to close. this PR is old and if the bug is still there- it'll pop up again!

paytonrules:
Is this the bug? Because shown up for me twice in the last week and I cannot figure out a workaround that isn't - uninstall EVERYTHING and rebuild Rust setup from scratch. 

```
[INFO]: ⬇️  Installing wasm-bindgen...
Error: /Users/eric/Library/Caches/.wasm-pack/wasm-bindgen-1a9c38279f773493/wasm-bindgen binary does not exist
...

[./pkg/index.js] 0 bytes {0} [built]
    + 20 hidden modules
ℹ ｢wdm｣: Compiled successfully.
ℹ ｢wdm｣: Compiling...
ℹ ｢wdm｣: wait until bundle finished: /index.js
✖ ｢wdm｣: Hash: 270856ba0af020d1aca9
Version: webpack 4.46.0
Time: 12ms
Built at: 04/13/2021 7:51:31 AM
 2 assets
Entrypoint index = index.js
[./pkg/index.js] 0 bytes {0} [built]
    + 34 hidden modules

ERROR in Rust compilation.
ℹ ｢wdm｣: Failed to compile.
```

paytonrules:
So further investigation, if this helps. 

wasm-pack build continues to fail for me, and I did a quick search for ~wasm-bindgen~ and noticed that I have both a temp directory without a `.` prefix, and another one with that prefix. The one with the prefix has ~wasm-bindgen~ in it.

```
[INFO]: ⬇️  Installing wasm-bindgen...
Error: /Users/eric/Library/Caches/.wasm-pack/wasm-bindgen-1a9c38279f773493/wasm-bindgen binary does not exist
bash-3.2$ find ~/Library/Caches/.wasm-pack/ -name wasm-bindgen -print
/Users/eric/Library/Caches/.wasm-pack//wasm-bindgen-36e18eb9086b3d15/wasm-bindgen
/Users/eric/Library/Caches/.wasm-pack//.wasm-bindgen-1a9c38279f773493/wasm-bindgen
/Users/eric/Library/Caches/.wasm-pack//wasm-bindgen-7753422a4b4f1c1f/wasm-bindgen
```

*Meanwhile*

```
bash-3.2$ ls ~/Library/Caches/.wasm-pack/wasm-bindgen-1a9c38279f773493/
wasm-bindgen-test-runner
```

PR #517, opened 2019-01-19
test(binaryinstall): Add initial tests

drager:
This PR will add some tests to the binary install crate. Please review even that it's WIP. I appreciate any comments what so ever.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -51,6 +51,10 @@ matrix:
     - rustup component add rustfmt-preview
     - cargo fmt --version
     - cargo fmt --all -- --check
+    - cd binary-install
+    - cargo test
+    - cargo fmt --all -- --check
+    - cd ..
 
   - name: Book
     rust: stable
--- a/.travis.yml
+++ b/.travis.yml
@@ -47,14 +47,10 @@ matrix:
     install:
       - *INSTALL_NODE_VIA_NVM
     script:
-    - cargo test --locked
+    - cargo test --all --locked
     - rustup component add rustfmt-preview
     - cargo fmt --version
     - cargo fmt --all -- --check
-    - cd binary-install
-    - cargo test
-    - cargo fmt --all -- --check
-    - cd ..
 
   - name: Book
     rust: stable
--- a/binary-install/src/lib.rs
+++ b/binary-install/src/lib.rs
@@ -278,64 +278,69 @@ fn hashed_dirname(url: &str, name: &str) -> String {
     format!("{}-{}", name, hex)
 }
 
-#[test]
-fn it_returns_same_hash_for_same_name_and_url() {
-    let name = "wasm-pack";
-    let url = "http://localhost:7878/wasm-pack-v0.6.0.tar.gz";
+#[cfg(test)]
+mod tests {
+    use super::*;
 
-    let first = hashed_dirname(url, name);
-    let second = hashed_dirname(url, name);
+    #[test]
+    fn it_returns_same_hash_for_same_name_and_url() {
+        let name = "wasm-pack";
+        let url = "http://localhost:7878/wasm-pack-v0.6.0.tar.gz";
 
-    assert!(!first.is_empty());
-    assert!(!second.is_empty());
-    assert_eq!(first, second);
-}
+        let first = hashed_dirname(url, name);
+        let second = hashed_dirname(url, name);
 
-#[test]
-fn it_returns_different_hashes_for_different_urls() {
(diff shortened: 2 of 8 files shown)


drager:
Some things that I would especially appreciate comments on, are about spinning up a local server for testing the downloading of tarballs. I think it's a good idea to spin up a local server rather than hitting a server somewhere else but I'm not sure about the ports of the server. Since the tests are run in parallel, each test that want to spin up a server needs its own port to not get into race condition and error with "port already used". This is somewhat error-prone and not ideal. What are your thoughts about this?

Another thing I'm thinking about is the naming of the tests. Should the function under test be mentioned in the test name? As I have done now it doesn't but maybe that's somewhat weird since all the functions are in the same file?

drager:
@danwilhelm Thanks for your comments! Yeah, existing tests in wasm-pack hits external servers, however binary-install will probably be put in it's own repo in the future. I mean, it's not absolutely necessary for binary-install to make the same decisions as wasm-pack but yeah, hitting external servers might not be a large concern, though I like the idea of be able to work fully offline.

Thanks for the suggestion on using lock(), will check that out!

About the tests in lib, my thoughts was that I would put tests in there rather than making functions public that wasn't meant to be public.

drager:
I think it would be great to get your thoughts/review of this @ashleygwilliams, @alexcrichton or @fitzgen. :blush:

binary-install/src/lib.rs
```diff
@@ -270,3 +260,82 @@ fn curl(url: &str) -> Result<Vec<u8>, Error> {
...
+        (result >> 32) as u8,
+        (result >> 40) as u8,
+        (result >> 48) as u8,
+        (result >> 56) as u8,
+    ]);
+    format!("{}-{}", name, hex)
+}
+
+#[test]
+fn it_returns_same_hash_for_same_name_and_url() {
```
fitzgen: Any tests that need access to private functions/types should be inside a `cfg`ed module like this:

```rust
#[cfg(test)]
mod tests {
    user super::*;

    // ...
}
```

This way they aren't included at all in non-test builds.

binary-install/tests/all/utils/mod.rs
```diff
@@ -0,0 +1,73 @@
...
+                None => {}
+            }
+
+            stream.flush().unwrap();
+        }
+        listener
+    });
+
+    // Wait for server to start...
+    thread::sleep(Duration::from_secs(1));
```
fitzgen: Perhaps use a [channel](https://doc.rust-lang.org/nightly/std/sync/mpsc/fn.channel.html) where the server can send a message after it has bound the tcp listener? That would be less racy.

.travis.yml
```diff
@@ -51,6 +51,10 @@ matrix:
     - rustup component add rustfmt-preview
     - cargo fmt --version
     - cargo fmt --all -- --check
+    - cd binary-install
+    - cargo test
+    - cargo fmt --all -- --check
```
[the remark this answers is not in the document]
drager: Oh, didn't know this! Thanks for the info, will fix that.

binary-install/src/lib.rs
drager: Right, missed that. Thanks!

binary-install/tests/all/utils/mod.rs
drager: Sounds like a plan, will try this out. Thanks!

drager:
@fitzgen Thanks for the review! I have now addressed all the comments.

binary-install/tests/all/utils/mod.rs
```diff
@@ -0,0 +1,75 @@
...
+use std::sync::mpsc::{channel, Receiver};
+use std::thread;
+
+pub const TEST_SERVER_HOST: &'static str = "localhost";
+
+pub fn start_server(port: u32, tarball: Option<Vec<u8>>) -> Receiver<()> {
+    let (sender, receiver) = channel();
+
+    thread::spawn(move || {
+        TcpListener::bind(format!("{}:{}", TEST_SERVER_HOST, port))
```
lfairy: If you pass 0 as the port, then the OS will assign a random unused one. You can then send the actual port back through the channel. This can be more robust than hard-coding a different value for each test!

See the third paragraph of https://doc.rust-lang.org/std/net/struct.TcpListener.html#method.bind

drager: Oh, thanks a lot! Didn't know this, will definitely do it!

PR #522, opened 2019-01-24
Port our tests to `assert_cmd`

fitzgen:
Encountering some hanging tests on my machine; not 100% sure what is going on yet.

### DIFF
--- a/src/manifest/mod.rs
+++ b/src/manifest/mod.rs
@@ -294,7 +294,7 @@ impl CrateData {
     pub fn warn_for_unused_keys(manifest_and_keys: &ManifestAndUnsedKeys) {
         manifest_and_keys.unused_keys.iter().for_each(|path| {
             PBAR.warn(&format!(
-                "\"{}\" is a unknown key and will be ignored. Please check your Cargo.toml.",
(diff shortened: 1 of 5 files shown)


danwilhelm:
@fitzgen It seems there are accidentally two `fixture.lock()`s in a row in `test::it_can_find_a_webdriver_on_path`. The second is blocking until the first is released, but it will never be released! This causes all subsequent tests to fail (perhaps since they are run in parallel by default?).

Also, I noticed your tests are using an [old hard-coded `bindgen` version](https://github.com/rustwasm/wasm-pack/blob/master/tests/all/utils/fixture.rs#L171). You may consider updating this version, or get it from the lockfile [as in `build.rs`](https://github.com/rustwasm/wasm-pack/blob/bedbf3de9df7d86a36bbd6bf4795a0b7dbe2f827/src/command/build.rs#L345).

fitzgen:
@danwilhelm d'oh! thanks for taking a look :-p

fitzgen:
Ok this is ready for review.

r? @ashleygwilliams or @alexcrichton

ashleygwilliams:
closing in favor of #534

Issue #469, opened 2018-12-24
Support both Node.js runtime and bundlers by default, out-of-the-box


KSXGitHub:
When I write an NPM package, I rarely think of bundlers as they can comprehend `require()` syntax, so as long as I only use `require()` to load module, I don't worry about bundlers. This is not the case with Node.js WASM modules generated by `wasm-bindgen` as it uses `fs.readFileSync` (which cannot be polyfilled by bundlers) to load `.wasm` file. 

`wasm-pack build` currently supports 2 targets:
  * `--target browser` support bundlers (today's browsers cannot load the output script/module directly) but not Node.js.
  * `--target nodejs` support Node.js runtime but not bundlers/browsers.

This is inconvenient, if I want to support both targets, I have to write my own wrapper that calls `wasm-bindgen` twice.

## Suggestion

* Without options, generate 4 files for 4 targets:
  * `index.bundler.js` for JavaScript bundlers, set `"browser"` field in `package.json` to point to this file.
  * `index.js` for Node.js runtime, set `"main"` field in `package.json` to point to this file.
  * `index.mjs` to work with `node --experimental-module`, `"module"` field of `package.json` may or may not point to this file.
  * `index.script.js` to work with `<script>` tag (I want to be able to test this directly in browser console).

* Deprecate the use of `--target <target>` in favor of `--nodejs index.js`, `--module index.mjs`, `--browser index.bundler.js`, and `--script index.script.js`.

## Additional Suggestion

Please don't use package name in `Cargo.toml` for JavaScript entry files, use `index.*` like most NPM packages.

ashleygwilliams:
hi @KSXGitHub ! this is a pretty common request that we get and something we fully intend to implement starting after the most recent release comes out!

KSXGitHub:
@ashleygwilliams Since you tagged this as "duplicate", can you also give me the link to the original issue?

BTW, this issue isn't a mere request, but also a small suggestion of how this could be implemented.

ashleygwilliams:
https://github.com/rustwasm/wasm-pack/issues/157

and i understand that it's a suggestion for implementation- we already have a pretty good idea of how we'd like to do this (very similar to what you have suggested- except we are not interested in supporting .mjs at this time) - just need to do it!

KSXGitHub:
@ashleygwilliams Thanks.

> except we are not interested in supporting .mjs at this time

What prevents you guy from doing this?

ashleygwilliams:
[...]

Issue #332, opened 2018-09-21
refactor target to use an enum


ashleygwilliams:
it's stringly typed right now, let's not, and use the code for access i wrote as a pattern

ashleygwilliams:
@alexcrichton did this in #567

Issue #157, opened 2018-06-06
Possible to publish both CommonJS and ES6 modules in same package?


mbalex99:
It looks like wasm_pack is sort of one or the other when it comes to publishing something for webpack browser consumption and one for nodejs.

Is it possible that wasm_pack can publish a package for both at the same time? 

Ideally something like this:

__Node__
```js
const { greet } = require('myModule')
```

__Browser in Webpack__
```js
const { greet } = await import("myModule");
```

ashleygwilliams:
hey @mbalex99 this is a great question! currently, `wasm-pack` can't do this, but i don't see why not! i can imagine a few ways to do this, so i'm gonna wait to hear other folks weigh in, but i think this could be a really neat feature! thanks for filing 🎉

Pauan:
This syntax is wrong:

```js
const { greet } = await import("myModule");
```

It should be like this:

```js
import { greet } from "myModule";
```

I'm not sure if it makes sense to publish CommonJS packages: if Node supports importing to/from wasm, then it should also support ES6 modules.

ashleygwilliams:
hey @Pauan - let's try to be a little more positive and constructive in our comments! yours was a bit blunt and critical (perhaps not your intention!) in a way that i don't think was necessary and doesn't fit the tone i'd like to set for this project. i'd love if you'd edit your comment to be friendlier. simply because we believe something "should" be the case doesn't always mean that it is the case, and we should endeavor to help folks deal with the current restrictions, even if they aren't ideal.

responding to your point: Node does *not* currently support ES6 modules out of the box (they are behind a flag), and many folks definitely still use CommonJS (in fact it's quite common)- i added support for commonJS to `wasm-bindgen` specifically to support this use case. it's a use case that ideally goes away in the future- but I don't think that future is very soon.

Pauan:
@ashleygwilliams My intention was certainly not to be rude! How should I point out factually incorrect things in a friendlier way?

Ah, I did not realize that wasm-bindgen has support for CommonJS, in that case you're right that it does make sense to support it.

bigfish24:
@ashleygwilliams could you share some of your thoughts on how to do this?

ashleygwilliams:
[...]

PR #523, opened 2019-01-24
[WIP] feat(build): allow specifying multiple targets

hobofan:
Addresses #157 / #469

There are still a few things that should be done:
- [ ] Deduplicate typescript definitions
- [ ] Refactor https://github.com/hobofan/wasm-pack/blob/d641e017e3c5869c8749b03b7b9c617bc749cb1b/src/manifest/mod.rs#L373-L390 into own function
- [ ] Tests tests tests ❗️ 

A few things that would be great to get your feedback on @ashleygwilliams :
- I removed the `EsModulesPackage`, etc. and flattened that to the `NpmPackage`, as I there was quite some duplication there, and I didn't see a good way to achieve the feature without that. I hope that's okay 😬 .
- Should I also tackle #313 and #332 in this PR, or better in a different PR?
- I'm not sure if I should add a special case handling for the case of a single target, in which case the old file naming convention (`{crate_name}.js`) should be used instead of the new names (`{crate_name}_{target}.js`)? I guess that might be required for better compatibility?

------ 

```bash
```


ashleygwilliams:
hey @hobofan ! thanks for this- it's late here for me but i'll take a look and leave some comments tomorrow morning. for a feature like this, one thing that's gonna be important is the expected final structure of the "isomorphic" package. could you share here in a comment the structure you anticipate?

thanks again!

ashleygwilliams:
@hobofan did you have a chance to read my previous comment? looks like we have some merge conflicts now... let me know where you are at with this PR and if there is anything i can do to help!

ashleygwilliams:
closing as out of date. @hobofan if you'd like to tackle this still, feel free to reopen this PR or make a new one!

PR #527, opened 2019-01-26
Support license-file field in Cargo.toml

rhysd:

```bash
```
  - I checked issues list. No issue was found.

In my package, no `license` field is provided in `Cargo.toml`, but `license-file` is provided. I noticed `wasm-pack` does not copy license file specified in `license-file` field.

https://doc.rust-lang.org/cargo/reference/manifest.html

This patch supports `license-file` field by following two additions:

- `wasm-pack` copies file specified in `license-file` in `Cargo.toml` into `pkg` directory
- Set `"SEE LICENSE IN <filename>"` to `license` field of `package.json`. This is standard way documented in official npm document: https://docs.npmjs.com/files/package.json#license

I added one new test case to check my implementation works fine.

> ✨✨ 😄 Thanks so much for contributing to wasm-pack! 😄 ✨✨

:+1:

### DIFF
--- a/src/license.rs
+++ b/src/license.rs
@@ -57,8 +57,8 @@ pub fn copy_from_crate(
         "crate's pkg directory should exist"
     );
 
-    match crate_data.crate_license() {
-        Some(_) => {
+    match (crate_data.crate_license(), crate_data.crate_license_file()) {
+        (Some(_), _) => {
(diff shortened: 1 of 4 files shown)


ashleygwilliams:
actually- this can go into 0.6.1, merging now

Issue #423, opened 2018-10-29
Compiler output doesn't have colors


alexcrichton:
## 🐛 Bug description

I just now realized that there wasn't an issue filed for this! But compiler diagnostics (from rustc and Cargo) both don't have color when executed through `wasm-pack`, due to the piping that `wasm-pack` currently does which disables the auto-detection of colors and such.

fitzgen:
Is there a way to force colors? If we let the child process inherit our own stdout/stderr, then we can't also log the output, nor can we return stdout/stderr or stuff them into the `Error` if the child exits nonzero.

alexcrichton:
I think the same trick cargo uses for rustc may be able to work with wasm-pack? Cargo basically always passes `--color always` to rustc and always captures the output. This causes rustc to use ansi escape codes on all platforms, and then cargo uses the [fwdansi crate](https://crates.io/crates/fwdansi) to translate the ansi codes to native APIs on Windows

alexcrichton:
er actually this is a dupe of https://github.com/rustwasm/wasm-pack/issues/288

PR #529, opened 2019-01-27
Colorize output from cargo on non-Windows OSes

rhysd:
related: #298

```bash
```

I found previous attempts at #293 #423. I understood that enabling colors causes a problem on Windows, but I still want to colorize output from cargo because it's much easier to read compilation errors from `cargo build`.

By this patch, I want to suggest to enable colors on non-Windows OSes. I read #298 and it seems to solve the Windows issue, but there is no update on the issue from Nov. If it would still take time to complete, I want to enable colors at least on non-Windows OSes.

### DIFF
--- a/src/bindgen.rs
+++ b/src/bindgen.rs
@@ -134,6 +134,9 @@ pub fn cargo_install_wasm_bindgen(
         .context("failed to create temp dir for `cargo install wasm-bindgen`")?;
 
     let mut cmd = Command::new("cargo");
+    if cfg!(not(windows)) {
+        cmd.arg("--color").arg("always");
+    }
     cmd.arg("install")
         .arg("--force")
         .arg("wasm-bindgen-cli")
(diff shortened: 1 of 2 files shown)


ashleygwilliams:
@alexcrichton does this have any negative implications that you are aware of? i certainly agree that we need to address #298 but would this be an acceptable mitigation for non windows users in the meantime?

alexcrichton:
I [feel the same as before](https://github.com/rustwasm/wasm-pack/pull/293#issuecomment-420311639), I wouldn't want to accept a fix which doesn't actually fix one of the major platforms.

rhysd:
@alexcrichton I understood your concern. I feel you're right. This change may give some impression that Windows support is worse than other platforms.

Since I agree with @alexcrichton 's thought, I'm closing this. I'll apply this patch personally.

@ashleygwilliams @alexcrichton Thank you for your quick reviews.

PR #530, opened 2019-01-27
Pass arguments through to cargo test

### LINKED ISSUE #525: Pass arbritrary arguments from wasm-pack test to wasm-bindgen-test
I started using wasm-pack at 0.6.0 and it's been great - thanks!

---

My biggest point of friction so far is trying to TDD with `wasm-pack` test. Needing to run an entire test suite when I'm focused one one function is no fun.

In #524 I had an idea for how to fix that, but I'm realizing that it adds a one off flag that might not really need to exist.

Instead, what if you could just pass anything through to `cargo test`?

```sh
# Anything after `--` gets passed to the `cargo test`
wasm-pack test --firefox --headless -- --package my-workspace-crate my_test_name --color=always
```

```rust
[wasm_bindgen_test]
fn my_test_name() {
  assert_eq!(2, 2);
}
```

In this way `wasm-pack` wouldn't need to support every little request that someone might have. We just expose `cargo test` and call it a day.

---

Could this work in theory? Any holes in my thinking? How difficult would this be (I'm unfamiliar with the codebase)?

Cheers!

\cc @fitzgen since you know a good bit about `wasm-pack test`

chinedufn:
Closes #525

### DIFF
--- a/docs/src/commands/test.md
+++ b/docs/src/commands/test.md
@@ -1,0 +1,54 @@
+# wasm-pack test
+
+The `wasm-pack test` command wraps the [wasm-bindgen-test-runner](https://rustwasm.github.io/wasm-bindgen/wasm-bindgen-test/index.html)
+CLI allowing you to run wasm tests in different browsers without needing to install the different
+webdrivers yourself.
+
+```
+wasm-pack test --help
+```
+
+## Path
+
+The `wasm-pack test` command can be given an optional path argument.
+
+This path should point to a directory that contains a `Cargo.toml` file. If no
+path is given, the `test` command will run in the current directory.
+
+```
+# Run tests for the current directory's crate
+wasm-pack test
+
+# Run tests for a specified crate
+wasm-pack test crates/crate-in-my-workspace
+```
+
+## Profile
+
+The `test` command accepts an optional profile argument: `--release`.
+
+If none is supplied, then a debug test build will be used.
+
+## Test environment
+
+Choose where to run your tests by passing in any combination of testing environment flags.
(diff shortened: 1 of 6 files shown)


src/command/test.rs
```diff
@@ -240,18 +246,6 @@ impl Test {
         Ok(())
     }
 
-    fn step_build_tests(&mut self, step: &Step) -> Result<(), Error> {
```
chinedufn: Don't need to build the tests before running them since `cargo test` will build them so I removed it.

LMK if that is an issue

chinedufn:
Mmmm I see the lockfile is used to know what version of wasm-bindgen to download... Cool adding that stuff back

src/test/mod.rs
```diff
@@ -10,7 +10,12 @@ use std::process::Command;
 
 /// Run `cargo test` with the `nightly` toolchain and targeting
 /// `wasm32-unknown-unknown`.
-pub fn cargo_test_wasm<I, K, V>(path: &Path, release: bool, envs: I) -> Result<(), failure::Error>
+pub fn cargo_test_wasm<I, K, V>(
+    path: &Path,
+    release: bool,
+    envs: I,
+    extra_options: &Vec<String>,
```
csmoe:
```suggestion
    extra_options: &[String],
```

docs/src/commands/test.md
```diff
@@ -0,0 +1,56 @@
+# wasm-pack test
```
fitzgen: Thank you!!!

PR #532, opened 2019-01-29
Add examples of running individual tests to the testing docs

chinedufn:


docs/src/commands/test.md
```diff
@@ -47,10 +47,39 @@ supported in wasm-pack.
 To use them you should add standalone `--` argument at the very
 end of your command, and all the arguments you want to pass to cargo should go after.
 
-Here's an example of running only tests that contain the world "apple".
+`cargo test -h` for a list of all options that you can pass through.
+
+## Running only some tests
+
+Are your test suite grows it may become unrealistic to need to run all of your tests in a browser when you're
```
ashleygwilliams: I would probably change this sentence to something like this, simply to keep it more direct. "When debugging a specific issue, you may find yourself wanting to run a subset of tests, instead of your entire suite of tests."

Don't feel the need to take my words, but I think I'd be interested in seeing your version made slightly more direct :)

chinedufn: Your version is was better than mine - took your advice verbatim - thanks!

Issue #535, opened 2019-01-31
Wrong error message for missing chromedriver binaries and missing fallback


jscheffner:
## 🐛 Bug description
When running `wasm-pack test --chrome` on a 64 bit Windows machine I get the error message `geckodriver binaries are unavailable for this target`.

#### 🤔 Expected Behavior
I would expect a fallback to win32 if a win64 binary is not available. In case there are some arguments against this, I'd still expect the error message to be `chromedriver binaries are unavailable for this target`. This is possibly a copy/paste error.

#### 👟 Steps to reproduce
Run `wasm-pack test --chrome` on a 64 bit Windows machine.

#### 🌍 Your environment
wasm-pack version: 0.6.0
rustc version: 1.32.0

**I'd like to submit a PR for this issue, but first I want to know if it should only fix the error message or actually implement the fallback to the win32 binary, so I'd really appreciate some feedback.**

ashleygwilliams:
hey @jscheffner thanks for filing this! i am running on a 64-bit mac machine and am not seeing this behavior- so this may be a windows only bug- which is very interesting!

i'd like to get someone to reproduce this, just to confirm, before we move forward. i think a PR that does the fallback seems very reasonable and would be something we want, but i do want to make sure we have a clear understanding of the scope of the problem before we act on a solution!

i have a 64-bit windows machine that i will try this out on now!

ashleygwilliams:
can confirm that i reproduced this on windows!

ashleygwilliams:
so- i get:

```
| [5/6] Getting chromedriver...
Error: geckodriver binaries are unavailable for this target
PS C:\Users\ashley\Projects\wasm-pack-template>
```

https://github.com/mozilla/geckodriver/releases -> this has a win64 binary...
my understanding is that chromedriver doesn't release a separate 32 vs 64 bit binary... https://sites.google.com/a/chromium.org/chromedriver/downloads

it's doubly weird that i asked it to test chrome and the error message says it's looking for the geckodriver :/

@fitzgen any thoughts on this? otherwise i can dig into the code and figure out why this is failing.

jscheffner:
[...]

PR #537, opened 2019-02-04
Use win32 chromedriver binary on win64, update chromedriver version and fix error message

jscheffner:
This solves #535.

Changes:
* use win32 chromedriver binary on win64 systems
* update chromedriver version to 2.45
* fix error message


jscheffner:
@ashleygwilliams: I updated the chromedriver version to 2.45 because you said so, but there is already a version 2.46. Is there any reason not to use this? Also, should I also update the geckodriver version to the latest?

ashleygwilliams:
@jscheffner sorry about that! the most up to date is all i meant!

PR #552, opened 2019-02-22
Fixing wasm-pack on non-rustup setups 

### LINKED ISSUE #545: wasm32-unknown-unkown target without rustup?
I am trying to follow the [Rust and WebAssembly tutorial](https://rustwasm.github.io/book/introduction.html), but I am stuck at ["Build the Project"](https://rustwasm.github.io/book/game-of-life/hello-world.html#build-the-project).

## What I have done

I am completely new to both rust and WebAssembly, so here is literally everything I have done, in order.

1. `git clone git@github.com:rust-lang/rust.git`.
2. Compiled rust from source.
3. Installed rust at _~/.local/bin/rustc_.
4. `git clone git@github.com:rust-lang/cargo.git`.
5. Compiled cargo from source (using the stage0 `cargo` binary from step 2).
6. `cp target/release/cargo ~/.local/bin/cargo`.
7. Setup _~/.cargo/config_ with `install.root` as _~/.local_.
8. `cargo install cargo-generate`.
9. `cargo generate --git https://github.com/rustwasm/wasm-pack-template`.
10. `cargo install wasm-pack` - and yes, I really did this *after* generating my project. The documentation is lacking here.
11. I followed [reddit directions](https://www.reddit.com/r/rust/comments/9t95fd/howto_setting_up_webassembly_on_stable_rust/) to install "wasm32-unknown-unknown" and verified the result is in _~/.local/lib/rustlib/wasm32-unknown-unknown_.

## ERRORS

It is at this point I am now stuck.

#### wasm-pack error

`wasm-pack build` is complaining that I don't have rustup installed. Which is true, but I don't really understand why that's a problem.

#### cargo error

`cargo build --target=wasm32-unknown-unknown` is throwing the following error.

```
error: failed to run `rustc` to learn about target-specific information

Caused by:
  process didn't exit successfully: `sccache rustc - --crate-name ___ --print=file-names --target wasm32-unknown-unknown --crate-type bin --crate-type rlib --crate-type dylib --crate-type cdylib --crate-type staticlib --crate-type proc-macro` (exit code: 1)
--- stderr
error: Could not create LLVM TargetMachine for triple: wasm32-unknown-unknown: No available targets are compatible with triple "wasm32-unknown-unknown"
```

#### rustup installer error

`curl https://sh.rustup.rs -sSf | sh` is complaining that I have an existing install of rust.

```
info: downloading installer
error: it looks like you have an existing installation of Rust at:
error: /home/code_m/.local/bin
error: rustup cannot be installed alongside Rust. Please uninstall first
error: if this is what you want, restart the installation with `-y'
error: cannot install while Rust is installed
```

This error also doesn't make any sense to me since it appears that rustup does more than _just install rustc_.

## Report Elsewhere

I would really rather not completely start over just to install the rustup tool. Please help me by telling me where else I should be reporting some of these errors.

drager:
This is a continuation on PR #451 which had some issues that this PR addresses.

Tested on NixOS without Rustup and on another Linux machine with Rustup.

Fixes #545


src/build.rs
```diff
@@ -49,11 +50,39 @@ fn rustc_minor_version() -> Option<u32> {
...
 }
 
+fn check_wasm32_target() -> Result<process::ExitStatus, ()> {
+    let echo_cmd = Command::new("echo")
+        .arg("fn main(){}")
+        .stdout(Stdio::piped())
+        .spawn();
+
+    echo_cmd.map_err(|_| ()).and_then(|echo_cmd| {
+        echo_cmd.stdout.ok_or_else(|| ()).and_then(|stdout| {
```
fitzgen: If there isn't a `stdout` then `std` is pretty broken and it probably doesn't even make sense to keep working with these child processes anymore. So we should just `stdout.unwrap()` here.

Personally, I also think this would read a bit better using normal `?` error propagation and a return type of `Result<bool, failure::Error>` where the `bool` is whether we have a working `wasm32-unknown-unknown` or not:

```rust
let echo_cmd = Command::new("echo")
    .arg("fn main() {}")
    .stdout(Stdio::piped())
    .spawn()
    .context("failed to spawn `echo` when checking wasm32-unknown-unknown target")?;

let status = Command::new("rustc")
    ...
    .stdin(echo_cmd.stdout.unwrap())
    .status();
match status {
    Ok(s) if s.success() => Ok(true),
    _ => Ok(false),
}
```

drager: Thank you for this great review! All of your comments makes sense!

drager:
@fitzgen Thanks, I have addressed all the comments :)

src/build.rs
```diff
@@ -49,11 +49,40 @@ fn rustc_minor_version() -> Option<u32> {
     otry!(pieces.next()).parse().ok()
 }
 
+fn check_wasm32_target() -> Result<bool, Error> {
+    let echo_cmd = Command::new("echo")
+        .arg("fn main(){}")
+        .stdout(Stdio::piped())
+        .spawn()
+        .context("failed to spawn `echo` when checking wasm32-unknown-unknown target")?;
```
kevincox: Why do you need to spawn a command here? Shouldn't you be able to just open a pipe and write into it?

RReverser: Yeah sorry @drager, I wasn't very clear in the comment on the original PR - by "equivalent of..." I didn't mean to literally invoke `echo`, but rather to use piping to `rustc` to write string directly.

You can do that with `.stdin(Stdio::piped())` and then writing a string.

fitzgen: It's about as verbose either way, but yeah this is probably more portable.

RReverser: Oh, sure, it's likely even more verbose, but definitely more portable.

alexcrichton:
I've personally found that running rustc (especially doing a full compliation) even for small files can be somewhat slow (500+ ms). In light of that I think we may want to take a different strategy for probing for the wasm target. Perhaps something like:

* First, run `rustc --print sysroot`
* Next, look at where `rustc` is when looked up via `PATH`
* If `rustc` is in the sysroot, then this is pretty unlikely to be rustup, so this strategy can be used of compiling a file.
* If `rustc` is not in the sysroot, then this is highly likely to be rustup. In that case if `$sysroot/lib/rustlib/wasm32-unknown-unknown` exists, the wasm target is installed

That means that for rustup setups we should get much faster lookups of the wasm target (only a handful of milliseconds as it's just a few filesystem lookups) but for other setups where we can't do any better than compiling a sample program we can do that.

src/build.rs
RReverser: Btw, one more thing: by default this will create a file called `rust_out` in the current dir. We might want to set `-o` to a file under OUT_DIR or something.

drager: My bad, I totally missed that a file was created... Thanks for letting me know. 

Also, I appreciate all the other comments from you all! What's your preference? Piping the string directly or invoke `echo`?

drager:
@alexcrichton Your suggestion sounds interesting! I will experiment with it a bit and see if I can get something to work out nicely.

Would this approach work cross platform or do you think it's something that needs to be handled differently depending on the platform?

Pauan:
@alexcrichton As an alternative, what if it ran `rustup` (just like we do now), but if the command fails (for any reason), then do the slower compile-and-check?

That seems more robust than making various assumptions about file system paths.

src/build.rs
RReverser: Piping is more reliable than `echo` command, so I'd change it.

drager:
> @alexcrichton As an alternative, what if it ran `rustup` (just like we do now), but if the command fails (for any reason), then do the slower compile-and-check?
> 
> That seems more robust than making various assumptions about file system paths.

I was thinking the same but this would result in a "unncessary check for the non-rustup case. In @alexcrichton's suggestion I interpret that once

drager:
> @alexcrichton As an alternative, what if it ran `rustup` (just like we do now), but if the command fails (for any reason), then do the slower compile-and-check?
> 
> That seems more robust than making various assumptions about file system paths.

I was thinking the same but this would result in a "unnecessary" check for the non-rustup case? In @alexcrichton's suggestion I interpret that we can choose a path early on to determine if Rustup or not which wouldn't result in doing "unnecessary" work. However, this might be fine?

About the closing of the issue, I missclicked, so that can be ignored...

alexcrichton:
@drager for rustup it's cross platform, and then the fallback to compiling a program should be cross platform as well. But in any case the sysroot layout for rustup is the same across all platforms.

@Pauan the downside of that is that we ideally want nothing printed if the target is already installed, but if it needs to be installed we want rustup to have control over the console so it can print its own progress bars/colors/such. In that sense we need to know when or not to run rustup.

src/build.rs
```diff
@@ -49,15 +51,88 @@ fn rustc_minor_version() -> Option<u32> {
...
-    let msg = format!("{}Adding the Wasm target...", emoji::TARGET);
-    PBAR.step(step, &msg);
+/// Get rustc's sysroot as a String
+fn get_rustc_sysroot() -> Result<String, FromUtf8Error> {
+    String::from_utf8(
+        Command::new("rustc")
+            .args(&["--print", "sysroot"])
+            .stdout(Stdio::piped())
+            .output()
+            .unwrap()
```
alexcrichton: This'll want to be sure to use `?` instead of `.unwrap()` here, and I think we'll also want to check `.status` to ensure it was successful before returning `Ok` as well

src/build.rs
```diff
@@ -49,15 +51,88 @@ fn rustc_minor_version() -> Option<u32> {
...
+    )
+}
+
+/// Checks if the wasm32-unknown-unknown is present in rustc's sysroot.
+fn is_wasm32_target_in_sysroot(sysroot: &str) -> Result<bool, Error> {
+    let wasm32_target = "wasm32-unknown-unknown";
+
+    info!("Looking for {} in {}", wasm32_target, sysroot);
+
+    let rustlib_path = &format!("{}/lib/rustlib", sysroot.replace("\n", ""));
```
alexcrichton: Instead of `replace` here, could the `get_rustc_sysroot` function above internally `trim()`?

src/build.rs
```diff
@@ -49,15 +51,88 @@ fn rustc_minor_version() -> Option<u32> {
     otry!(pieces.next()).parse().ok()
 }
 
-/// Ensure that `rustup` has the `wasm32-unknown-unknown` target installed for
-/// current toolchain
-pub fn rustup_add_wasm_target(step: &Step) -> Result<(), Error> {
-    let msg = format!("{}Adding the Wasm target...", emoji::TARGET);
-    PBAR.step(step, &msg);
+/// Get rustc's sysroot as a String
+fn get_rustc_sysroot() -> Result<String, FromUtf8Error> {
```
alexcrichton: This'll probably want to return a `PathBuf` instead of a `String`

src/build.rs
```diff
@@ -49,15 +51,88 @@ fn rustc_minor_version() -> Option<u32> {
...
+
+/// Checks if the wasm32-unknown-unknown is present in rustc's sysroot.
+fn is_wasm32_target_in_sysroot(sysroot: &str) -> Result<bool, Error> {
+    let wasm32_target = "wasm32-unknown-unknown";
+
+    info!("Looking for {} in {}", wasm32_target, sysroot);
+
+    let rustlib_path = &format!("{}/lib/rustlib", sysroot.replace("\n", ""));
+
+    let ls_command = Command::new("ls")
```
alexcrichton: Instead of using the `ls` and `grep` commands, this can probably use the native platform APIs, and I think all it needs to do is to test for `path.exists()` perhaps?

src/build.rs
```diff
@@ -49,15 +51,88 @@ fn rustc_minor_version() -> Option<u32> {
...
+}
+
+/// Add wasm32-unknown-unknown using `rustup`.
+fn rustup_add_wasm_target() -> Result<bool, Error> {
     let mut cmd = Command::new("rustup");
     cmd.arg("target").arg("add").arg("wasm32-unknown-unknown");
     child::run(cmd, "rustup").context("Adding the wasm32-unknown-unknown target with rustup")?;
-    Ok(())
+
+    Ok(true)
```
alexcrichton: Since the `bool` doesn't really mean much from this function, could this continue to return `Ok(())` and then above it's `.map(|()| true)`'d?

src/build.rs
```diff
@@ -49,15 +51,88 @@ fn rustc_minor_version() -> Option<u32> {
...
+}
+
+/// Ensure that `rustup` has the `wasm32-unknown-unknown` target installed for
+/// current toolchain
+pub fn check_for_wasm32_target(step: &Step) -> Result<(), Error> {
+    let msg = format!("{}Checking for the Wasm target...", emoji::TARGET);
+    PBAR.step(step, &msg);
+
+    // Check if wasm32 target is present, otherwise bail.
+    match check_wasm32_target() {
```
alexcrichton: I think this'll want to be:

```rust
if !check_wasm32_target()? {
    bail!(...)
}
```

src/build.rs
drager: Yes, good point! Thanks!

src/build.rs
drager: Sounds good!

src/build.rs
drager: Yes, that seems like a good idea indeed.

src/build.rs
drager: Yes, that true. I think that's a much better solution, thanks!

src/build.rs
drager: Agree!

src/build.rs
drager: Hmm, shouldn't this function return a `Result<(), Error>` anymore?

alexcrichton: Oh sure yeah, but if an error happens while handling the wasm target we want to be sure to return that error rather than accidentally swallow it up

src/build.rs
```diff
@@ -49,17 +50,80 @@ fn rustc_minor_version() -> Option<u32> {
...
-/// Ensure that `rustup` has the `wasm32-unknown-unknown` target installed for
-/// current toolchain
-pub fn rustup_add_wasm_target(step: &Step) -> Result<(), Error> {
-    let msg = format!("{}Adding the Wasm target...", emoji::TARGET);
-    PBAR.step(step, &msg);
+/// Get rustc's sysroot as a PathBuf
+fn get_rustc_sysroot() -> Result<PathBuf, failure::Error> {
+    let command = Command::new("rustc")
+        .args(&["--print", "sysroot"])
+        .stdout(Stdio::piped())
```
alexcrichton: This can actually be removed since it's the default when calling `.output()`

src/build.rs
drager: Right! I misunderstood, now I only `bail!` for falses, otherwise return `Err(err)`.

src/build.rs
drager: Oh right, thanks!

ashleygwilliams:
restarted travis- if it's green it looks like we can merge this! (and sneak it into 0.7.0!!)

ashleygwilliams:
(this is gonna have some merge conflicts with the 1.0 strategy PR but i feel confident i can resolve them pretty quick (just the removal of `step` and maybe a few other lil logging things))

drager:
Great! Let me know if I should handle the merge conflicts 😊

ashleygwilliams:
ok just tried to merge the conflicts, think i did ok but it isssss the end of my day - we'll see what travis says

ashleygwilliams:
welp missed some instances of step- will fix once i get back to a computer

ashleygwilliams:
closing in favor of https://github.com/rustwasm/wasm-pack/pull/577

PR #555, opened 2019-02-26
"Wasm" is a contraction, not an acronym

fitzgen:
And therefore should not be all caps.

https://webassembly.github.io/spec/core/intro/introduction.html#wasm

v important

### DIFF
--- a/src/build.rs
+++ b/src/build.rs
@@ -52,7 +52,7 @@ fn rustc_minor_version() -> Option<u32> {
 /// Ensure that `rustup` has the `wasm32-unknown-unknown` target installed for
 /// current toolchain
 pub fn rustup_add_wasm_target(step: &Step) -> Result<(), Error> {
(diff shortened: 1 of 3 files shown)


src/build.rs
```diff
@@ -67,7 +67,7 @@ pub fn cargo_build_wasm(
     step: &Step,
     extra_options: &Vec<String>,
 ) -> Result<(), Error> {
-    let msg = format!("{}Compiling to WASM...", emoji::CYCLONE);
+    let msg = format!("{}Compiling to Wasm...", emoji::CYCLONE);
```
ashleygwilliams: this feels like the weirdest of WASM, wasm, Wasm- any objection to making it all lower case?

fitzgen: It's what the spec and official website use:

> WebAssembly (abbreviated Wasm) ...

> The Wasm stack machine is designed to be ...

Makes sense to me since it is still a proper noun

PR #556, opened 2019-02-27
fix(manifest): LICENSE and README don't need to be whitelisted

### LINKED ISSUE #551: revert pull/528
https://docs.npmjs.com/files/package.json#files

README and LICENSE are whitelisted so for the sake of maintenance cost we should remove the code that adds these.

ashleygwilliams:
fixes #551

### DIFF
--- a/src/manifest/mod.rs
+++ b/src/manifest/mod.rs
@@ -430,17 +430,13 @@ impl CrateData {
             None
         };
 
-        let readme_file = out_dir.join("README.md");
-        if readme_file.is_file() {
-            files.push("README.md".to_string());
-        }
-
         if let Ok(entries) = fs::read_dir(out_dir) {
             let file_names = entries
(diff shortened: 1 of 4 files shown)


ashleygwilliams:
both LICENSE and README (with any/all/no extension) are always included in the package, regardless of presence in the `files` key or a `.npmignore` or `.gitignore` file. i mistakenly merged a PR that added any file starting with `LICENSE` and `README` to the `files` key in `package.json`.

this PR reverts a bit of that work, leaving the work that adds non-standard license files (e.g. `license = "MY_LICENSE" or "LICENSE-MIT") to the `files` key. 

r? @alexcrichton or @fitzgen

ashleygwilliams:
i think that using the `files` key is still something i'd like to do, but we can talk about exactly why in an issue- do you want to file one @fitzgen ?

PR #565, opened 2019-03-07
Flag gh-pages docs as unpublished

alexcrichton:
This commit adds a header dialog to the gh-pages mdbook documentation of
this repository to indicate that it represents unpublished documentation
that may not reflect the currently published version of `wasm-pack`. The
header redirects users to https://rustwasm.github.io/docs/wasm-pack/
which has the documentation for the latest published version of
`wasm-pack`.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -62,7 +62,7 @@ matrix:
     - (test -x $HOME/.cargo/bin/mdbook || cargo install --vers "^0.2" mdbook)
     - cargo install-update -a
     script:
-    - (cd docs && mdbook build)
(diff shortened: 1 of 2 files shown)


ashleygwilliams:
@alexcrichton could you add a screencap for posterity? thanks for this!

alexcrichton:
When rendered this will look like what [wasm-bindgen](https://rustwasm.github.io/wasm-bindgen/) and  the [book](https://rustwasm.github.io/book/) currently look like, only with the current text here for wasm-pack

PR #567, opened 2019-03-07
Add support for `--target web`

alexcrichton:
This commit adds support for the new `--web` flag in `wasm-bindgen`
under the flag name `--target web`. To ensure that it was plubmed around
the stringly-typed `target` type was switched to an `enum Target` to
ensure that all cases it's looked at are handled appropriately for the
new `web` target.

### DIFF
--- a/docs/src/commands/build.md
+++ b/docs/src/commands/build.md
@@ -47,20 +47,26 @@ The exact meaning of the profile flags may evolve as the platform matures.
 
 ## Target
 
-The `build` command accepts a `--target` argument. This will customize the output files
-to align with a particular type of JS module. This allows wasm-pack to generate either
-ES6 modules or CommonJS modules for use in browser and in NodeJS. Defaults to `browser`.
-The options are:
+The `build` command accepts a `--target` argument. This will customize the JS
+that is emitted and how the WebAssembly files are instantiated and loaded. For
+more documentation on the various strategies here, see the [documentation on
+using the compiled output][deploy].
 
 ```
 wasm-pack build --target nodejs
 ```
 
-| Option    | Description                                                                                                     |
-|-----------|-----------------------------------------------------------------------------------------------------------------|
-| `nodejs`  | Outputs JS that uses CommonJS modules, for use with a `require` statement. `main` key in `package.json`. |
-| `no-modules`  | Outputs JS that use no modules. `browser` key in `package.json`. |
-| `browser` | Outputs JS that uses ES6 modules, primarily for use with `import` statements and/or bundlers such as `webpack`. `module` key in `package.json`. `sideEffects: false` by default. |
+| Option    | Usage | Description                                                                                                     |
+|-----------|------------|-----------------------------------------------------------------------------------------------------|
+| *not specified* or `bundler` | [Bundler][bundlers] | Outputs JS that is suitable for interoperation with a Bundler like Webpack. You'll `import` the JS and the `module` key is specified in `package.json`. `sideEffects: false` is by default. |
+| `nodejs`  | [Node.js][deploy-nodejs] | Outputs JS that uses CommonJS modules, for use with a `require` statement. `main` key in `package.json`. |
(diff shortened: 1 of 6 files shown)


src/bindgen.rs
```diff
@@ -203,9 +203,10 @@ pub fn wasm_bindgen_build(
         "--typescript"
     };
     let target_arg = match target {
```
ashleygwilliams: thanks for making this an enum it's a clear improvement!

src/bindgen.rs
```diff
@@ -203,9 +203,10 @@ pub fn wasm_bindgen_build(
...
-        "nodejs" => "--nodejs",
-        "no-modules" => "--no-modules",
-        _ => "--browser",
+        Target::Nodejs => "--nodejs",
+        Target::NoModules => "--no-modules",
+        Target::Web => "--web",
+        Target::Bundler => "--browser",
```
ashleygwilliams: should we perhaps deprecate `--browser` and add a `--bundler` alias in wasm-bindgen? (you might already be on this, in which case linking to a PR/issue would be great)

ashleygwilliams:
rebase to include the panic fix for the tests should make this green!

alexcrichton:
Sure, I've updated the documentation

src/bindgen.rs
alexcrichton: I think that long term we may want to but for now it serves a purpose we haven't displaced, so it can't be removed just yet

docs/src/commands/build.md
```diff
@@ -47,20 +47,26 @@ The exact meaning of the profile flags may evolve as the platform matures.
...
 ## Target
 
-The `build` command accepts a `--target` argument. This will customize the output files
-to align with a particular type of JS module. This allows wasm-pack to generate either
-ES6 modules or CommonJS modules for use in browser and in NodeJS. Defaults to `browser`.
-The options are:
+The `build` command accepts a `--target` argument. This will customize the JS
+that is emitted and how the WebAssembly files are instantiated and loaded. For
+more documentation on the various strategies here, see the [documentation on
+deployment][deploy].
```
ashleygwilliams: i totally see how "deployment" can work as a word to describe what we mean here but it *may* be a bit overloaded for a general programmer reading the docs, which might cause some confusion (i.e. deployment = pushing to a prod server of some sort). maybe we can say usage instead? i think that may be less intimidating.

src/bindgen.rs
ashleygwilliams: do we want to open a tracking issue? let me know what you think.

docs/src/commands/build.md
alexcrichton: Sure, I've tweaked the wording

src/bindgen.rs
alexcrichton: I've opened up https://github.com/rustwasm/wasm-bindgen/issues/1332 for this

docs/src/commands/build.md
```diff
@@ -47,20 +47,26 @@ The exact meaning of the profile flags may evolve as the platform matures.
...
 ```
 wasm-pack build --target nodejs
 ```
 
-| Option    | Description                                                                                                     |
-|-----------|-----------------------------------------------------------------------------------------------------------------|
-| `nodejs`  | Outputs JS that uses CommonJS modules, for use with a `require` statement. `main` key in `package.json`. |
-| `no-modules`  | Outputs JS that use no modules. `browser` key in `package.json`. |
-| `browser` | Outputs JS that uses ES6 modules, primarily for use with `import` statements and/or bundlers such as `webpack`. `module` key in `package.json`. `sideEffects: false` by default. |
+| Option    | Deployment | Description                                                                                                     |
```
ashleygwilliams: hey @alexcrichton sorry for the nit but the word "Deployment" is still used in this chart- could we change it to Usage? thanks!

alexcrichton: Sure, updated

PR #599, opened 2019-03-22
Added support for --out-name flag

### LINKED ISSUE #596: Support --out-name from wasm-bindgen
## 💡 Feature description
Support passing `--out-name` to **wasm-bindgen**

There is an [issue](https://github.com/rustwasm/wasm-bindgen/issues/1076) from **wasm-bindgen**, and a [PR](https://github.com/rustwasm/wasm-bindgen/pull/1084)
#### 💻 Basic example
`wasm-pack --out-name index`
This will set output file name to `index.js`. Defaults to package name.

ibaryshnikov:
Closes #596 

Should I also add tests for this feature?

### DIFF
--- a/docs/src/commands/build.md
+++ b/docs/src/commands/build.md
@@ -32,7 +32,7 @@ wasm-pack build --out-dir out
 The above command will put your build artifacts in a directory called `out`, instead
 of the default `pkg`.
 
-## Output file names
+## Generated file names
 
 Flag `--out-name` sets the prefix for output file names. If not provided, package name is used instead.
 
--- a/docs/src/commands/build.md
+++ b/docs/src/commands/build.md
@@ -32,6 +32,16 @@ wasm-pack build --out-dir out
 The above command will put your build artifacts in a directory called `out`, instead
 of the default `pkg`.
 
+## Output file names
+
+Set output file names. Defaults to package name.
+
+```
+wasm-pack build --out-name index
+```
+
+This command will build the package with file names like `index.js`, `index.d.ts`, and so on.
+
 ## Profile
 
 The `build` command accepts an optional profile argument: one of `--dev`,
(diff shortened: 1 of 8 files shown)


ashleygwilliams:
@ibaryshnikov tests and documentation would be great! thank you so much for taking the initiative on this!

ibaryshnikov:
@ashleygwilliams I've added tests

docs/src/commands/build.md
```diff
@@ -32,6 +32,16 @@ wasm-pack build --out-dir out
 The above command will put your build artifacts in a directory called `out`, instead
 of the default `pkg`.
 
+## Output file names
```
ashleygwilliams: We might add some more info on exactly what files are outputted and what the names by default are and what parts of the name change when you use this command.

docs/src/commands/build.md
```diff
@@ -32,6 +32,16 @@ wasm-pack build --out-dir out
...
+
+Set output file names. Defaults to package name.
+
+```
+wasm-pack build --out-name index
+```
+
+This command will build the package with file names like `index.js`, `index.d.ts`, and so on.
```
ashleygwilliams: i'd also included what the wasm file is called here (this covers almost all the files and since the wasm one is pretty important, i'd include that here.)

docs/src/commands/build.md
ibaryshnikov: Have no idea how to choose the title here ((

docs/src/commands/build.md
ibaryshnikov: Updated with more detailed information

docs/src/commands/build.md
ashleygwilliams: @ibaryshnikov maybe "Generated file names"?

ibaryshnikov: @ashleygwilliams thank you!

PR #601, opened 2019-03-23
feat: Print unexpected panics to standard error

### LINKED ISSUE #562: Unexpected panics should still print the full panic message
Unexpected panics are unfortunate but they're currently covered up and written out to an auxiliary file. This makes [panics in CI](https://dev.azure.com/rustwasm/wasm-weight-tracker/_build/results?buildId=33) difficult to debug, especially at a glance, as CI builders are likely not uploading those files.

`wasm-pack` should still print something to the console indicating what sort of error just happened.

drager:
This will print to standard error for unexpected panics and then let `human_panic` handle panics, just like before.

This will fix #562.

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -2,13 +2,13 @@ extern crate atty;
 extern crate env_logger;
 #[macro_use]
 extern crate failure;
-#[macro_use]
 extern crate human_panic;
 extern crate structopt;
 extern crate wasm_pack;
 extern crate which;
 
 use std::env;
+use std::panic;
 use structopt::StructOpt;
 use wasm_pack::{command::run_wasm_pack, Cli};
 
@@ -16,7 +16,9 @@ mod installer;
 
 fn main() {
     env_logger::init();
-    setup_panic!();
+
+    setup_panic_hooks();
+
     if let Err(e) = run() {
         eprintln!("Error: {}", e);
         for cause in e.iter_causes() {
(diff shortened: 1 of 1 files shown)


src/main.rs
```diff
@@ -49,3 +51,29 @@ fn run() -> Result<(), failure::Error> {
...
+    match env::var("RUST_BACKTRACE") {
+        Err(_) => {
+            panic::set_hook(Box::new(move |info: &panic::PanicInfo| {
+                // First call the default hook that prints to standard error.
+                default_hook(info);
+
+                // Then call human_panic.
+                let file_path = human_panic::handle_dump(&meta, info);
+                human_panic::print_msg(file_path, &meta)
+                    .expect("human-panic: printing error message to console failed");
```
ashleygwilliams: if you'd like to use ? here instead of expect we can add returning a Result from main to this PR in another commit! if not, that's OK :)

drager:
@ashleygwilliams Yeah, you could do this by just panicing somewhere in the wasm-pack codebase. Here's what it looks like:

image

ashleygwilliams:
per convo in discord, @drager is gonna do the refactor of main to return a result by monday so we can land this in the 0.8.0 release i intend to do on that day. if not, we can land this as is for the release and refactor later :)

src/main.rs
drager: I'm not sure we can do that since [set_hook](https://doc.rust-lang.org/std/panic/fn.set_hook.html) wont accept a `Result` as a return type for the hook. It only accepts `()`.

drager:
@ashleygwilliams I'm not sure we can refactor that `expect` do that since `set_hook` wont accept a `Result` as a return type for the hook. It only accepts ().

ashleygwilliams:
@drager let's just land this as is then!

Issue #589, opened 2019-03-16
wasm-pack can't install wasm32-unknown-unkonwn target on DragonFly BSD


strangelittlemonkey:
## 🐛 Bug description
Following the hello world tutorial example, I seem to hit 32bit requirements on a 64bit only operating system.

The following was observed, and annotated in another issue here:

I tried building off commit badbe1c, as well as off the 0.6.0 tag and the 0.7.0 tag, all failed. This was as well with the hello world example.

In order:     0.6.0, 0.7.0, master 

```
zach@dev03 ~/wasm-pack-template> wasm-pack build

  [1/10] Checking `rustc` version...
  [2/10] Checking crate configuration...
| [3/10] Adding WASM target...
Error: Adding the wasm32-unknown-unknown target with rustup
Caused by: No such file or directory (os error 2)
zach@dev03 ~/wasm-pack-template> ../wasm-pack/target/release/wasm-pack build
[INFO]: Checking for the Wasm target...
Error: wasm32-unknown-unknown target not found!
zach@dev03 ~/wasm-pack-template> ../wasm-pack/target/release/wasm-pack build
[INFO]: Checking for the Wasm target...
Error: wasm32-unknown-unknown target not found!
```

#### 🤔 Expected Behavior
I expected that the hello world example would work out of the box, but it's possible that the documentation is not explicit that only 32bit compilation is supported and/or that 32bit libraries are required and/or that rustup support is required with 32bit binaries and libraries for this.

#### 👟 Steps to reproduce
I did a git clone of the wasm-pack-template (my version of cargo does not support generate), so I then edited the Cargo.toml for an arbitrary author name and project name.

I installed the latest release of wasm-pack by cargo (0.6.0 as of this morning) and tried to do wasm-pack build, after not working and finding an issue exhibiting what I thought was the issue, checked out the source and did a git checkout of 0.7.0, with the a similar error. I then checked out master and tried on master with the same error as on 0.7.0.

#### 🌍 Your environment
Include the relevant details of your environment.
wasm-pack version: 0.6.0,0.7.0, master (badbe1c37a752963243da526b6fb01b66e7889bd)
rustc version: 1.31.1

Pauan:
> I seem to hit 32bit requirements on a 64bit only operating system.

What OS are you using that is 64-bit only and doesn't support rustup?

[...]

PR #602, opened 2019-03-23
Improve error message when wasm32-unknown-unknown is missing

### LINKED ISSUE #579: Error for missing wasm32-unknown-unknown on non-rustup environments can be improved
## 🐛 Bug description
As a newbie I am following the tutorial at https://rustwasm.github.io/book/game-of-life/hello-world.html and get errors

#### 🤔 Expected Behavior
When building the project with `wasm-pack build` I expect no errors.

#### 👟 Steps to reproduce
```
$ cargo generate --git https://github.com/rustwasm/wasm-pack-template
```
then type `wasm-game-of-life`

```
$ cd wasm-game-of-life
$ wasm-pack build
[INFO]: 🎯  Checking for the Wasm target...
Error: wasm32-unknown-unknown target not found!
```

#### 🌍 Your environment
wasm-pack version: 0.7.0
rustc version: 1.31.0
MacOS Sierra

drager:
A work in progress to resolve #579.


drager:
Would appreciate review from you @ashleygwilliams and @alexcrichton. I put the docs under "Prerequisites" but since it became pretty long I think it should maybe be in it's own section. What do you guys think?

docs/src/prerequisites/index.md
```diff
@@ -17,3 +17,39 @@ you're excited about that work- you should reach out to the maintainers and get
 involved!
 
 [npm]: prerequisites/npm.html
+
+## Target wasm32-unknown-unknown
```
ashleygwilliams: i agree with your assessment that this might be best as it's own page- i think the title might better be phrased as "Non-rustup setups" this way, new folks who see this and are unsure, or just normal users with a rustup setup don't think they need to read this bit. 

I would make this its own page, replacing this section with a line that links to the page with this content saysing "Non-rustup setup? Learn how to configure for wasm-pack"- or something like this.

src/build.rs
```diff
@@ -80,20 +81,77 @@ fn is_wasm32_target_in_sysroot(sysroot: &PathBuf) -> bool {
     }
 }
 
-fn check_wasm32_target() -> Result<bool, Error> {
+struct Wasm32Check {
```
ashleygwilliams: this logic is getting pretty large- we may consider making build a directory, having the primary build work in a mod.rs and putting the wasm32 target logic in it's own file called wasm-target or something similar

docs/src/prerequisites/index.md
alexcrichton: Agreed with a separate page, and I think we'll want a disclaimer as well saying that it may not actually be possible to add the wasm32 target to all setups. The instructions below are correct for setups which match rustc nightlies/releases exactly (down to the precise commit being used to build the compiler). If anything is different though (aka if a distro has patches, or if it's not built the same way as Rust releases) then the downloaded wasm32 target will still be incompatible.

I think it's good to have the instructions below, but we probably still want a warning saying that it's not guaranteed to work.

ashleygwilliams: do we know of toolchains where it definitely will or wont work? it's always nice to list some examples.

alexcrichton: I'm not personally aware of either myself (I've only ever used rustup). It looks like the [DragonFly BSD toolchain](https://github.com/rustwasm/wasm-pack/issues/589#issuecomment-473934474) for sure isn't compatible due to it carrying local patches, but that's all I'd know of

drager: Yeah, I will move it to a separate page and add a disclaimer as well

src/build.rs
drager: Sure, I can move it

docs/src/prerequisites/index.md
drager: I have moved it to a separate page and I link to that from prerequisites. I also added a disclaimer. Feel free to comment on wording in the section as well. Thanks!

gotcha:
@drager @ashleygwilliams I am impressed by the level of attention to details and usability ! Thanks

PR #610, opened 2019-03-29
fix(doc): put non rustup conetnt in the right spot

ashleygwilliams:
i realize that this was originally a relative paths issue, but when i noted that the link in the prereq index was broken i immediately figured that the doc was in the wrong place (root of docs, instead of prereqs). this PR adds it to the prereq section (which also happens to unbreak the link heh)


ashleygwilliams:
@drager actually i ran it locally and that link works! i think it's the relative paths stuff that makes it hard to realize

drager:
@ashleygwilliams Oh ok, I thought it didn't work since I looked at how the link to npm looked like: `[npm]: prerequisites/npm.html`, also I can't get it to work without `prerequisites` locally either.

ashleygwilliams:
@drager huh! i will try again locally

ashleygwilliams:
the npm link is also broken on this repo and the primary rustwasm bookshelf! i'm going to keep this as is and land it and see how it works.

PR #612, opened 2019-04-01
fix the docs

ashleygwilliams:


drager:
Found another broken link here in the bottom: https://raw.githubusercontent.com/rustwasm/wasm-pack/master/docs/src/tutorials/npm-browser-packages/getting-started/manual-setup.md Should also be `tutorials`.

docs/src/tutorials/npm-browser-packages/template-deep-dive/cargo-toml.md
```diff
@@ -100,12 +98,10 @@ console_error_panic_hook = { version = "0.1.1", optional = true }
...
 ```
 
-[`cfg-if`] allows us to check if certain features are enabled on a Rust crate. We'll
-use this crate later to optionally enable [`console_error_panic_hook` or
-`wee_alloc`.
-
-By default, only `console_error_panic_hook` is enabled. To disable either
-feature, we can remove its name from the `default` vector.
+In our code, we'll mark certain parts of code as running only if certain `[features]`
+are enabled, specifically, `console_error_panic_hook` and `weealloc`. By default,
```
fitzgen:
```suggestion
are enabled, specifically, `console_error_panic_hook` and `wee_alloc`. By default,
```

docs/src/tutorials/npm-browser-packages/template-deep-dive/src-lib-rs.md
```diff
@@ -63,43 +82,19 @@ Either way, the contents of `utils.rs` define a single public function `set_pani
 
 
 ```rust
-use cfg_if::cfg_if;
+// When the `wee_alloc` feature is enabled, use `wee_alloc` as the global
```
fitzgen:
```suggestion
    // When the `wee_alloc` feature is enabled, use `wee_alloc` as the global
```

Since the code below is also indented by 4 spaces.

ashleygwilliams:
ok so:

the weird "relative link "bug"" was actually because mdbook shows the prereq page as the "landing page" at "/" (root route of the docs) and when you click on it "prerequisites/index.html", this makes our usage of ".." in the paths problematic.

to fix this, i opted to solve a different problem. instead of landing folks on the prereqs page, we now land them on this. it has no relative links and so i have removed the problem.
Screen Shot 2019-04-01 at 3 49 13 PM

ashleygwilliams:
added some more stuff from the readme per @drager 's excellent suggestion.

Screen Shot 2019-04-01 at 3 51 19 PM

ashleygwilliams:
merging before CI because the relevant tests passed and i would like to get a release out.

PR #616, opened 2019-04-02
fix: Check for "rustup" rather than ".rustup" when checking for wasm32

### LINKED ISSUE #613: Doesn't automatically add wasm32-unknown-uknown target in Docker.
## 🐛 Bug description
The official Docker image has rustup installed but wasm-pack doesn't automatically add the wasm32-unknown-unknown target when building for the first time.

#### 🤔 Expected Behavior
It should have added the wasm32-unknown-unknown target automatically.

#### 👟 Steps to reproduce
I came across this when going through the rust-wasm book. Here's the steps to reproduce but it involves a different repository from this one, so I hope the steps are acceptable.

```
$ docker run -it --rm --env USER rust:1.33.0-stretch bash -c '
cargo install cargo-generate;
curl https://rustwasm.github.io/wasm-pack/installer/init.sh -sSf | sh;
cargo generate --git https://github.com/rustwasm/wasm-pack-template --name wasm-game-of-life;
cd wasm-game-of-life;
wasm-pack build;
'
...redacted a bunch of compiling...
   Compiling cargo-generate v0.2.2
    Finished release [optimized] target(s) in 3m 00s
  Installing /usr/local/cargo/bin/cargo-generate
info: downloading wasm-pack
info: successfully installed wasm-pack to `/usr/local/cargo/bin/wasm-pack`
 Creating project called `wasm-game-of-life`...
 Done! New project created /wasm-game-of-life
Error: wasm32-unknown-unknown target not found!
$
```

#### 🌍 Your environment
wasm-pack version: latest?
rustc version: 1.33.0

I develop things out of Docker. I narrowed down the bug to [this line in the wasm-pack source where it checks for a `syspath`/.rustup file](https://github.com/rustwasm/wasm-pack/blob/294741411c6ec30e6cdf063a0955d476ef351b0e/src/build/wasm_target.rs#L116). The Docker image has rustup but doesn't have this file.

```
$ docker run -it --rm rust:1.33.0-stretch bash -c 'echo Has rustup: $(which rustup); ls $(rustc --print sysroot)/.rustup;'
Has rustup: /usr/local/cargo/bin/rustup
ls: cannot access '/usr/local/rustup/toolchains/1.33.0-x86_64-unknown-linux-gnu/.rustup': No such file or directory
```

drager:
When checking for wasm32 target we did check if the sysroot contained ".rustup". While this covered the most common cases it didn't work when using Docker. So checking for "rustup" instead covers both
cases.

I also noticed another bug where if the target was found we printed the helpful information and not the way around so I fixed that as well. Also, the code linked to the wrong docs section since the last docs update so I corrected that.

This will fix #613

### DIFF
--- a/src/build/wasm_target.rs
+++ b/src/build/wasm_target.rs
@@ -20,7 +20,7 @@ impl fmt::Display for Wasm32Check {
     fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
         let target = "wasm32-unknown-unknown";
 
-        if self.found {
+        if !self.found {
             let rustup_string = if self.is_rustup {
                 "It looks like Rustup is being used.".to_owned()
             } else {
--- a/src/build/wasm_target.rs
+++ b/src/build/wasm_target.rs
@@ -24,7 +24,7 @@ impl fmt::Display for Wasm32Check {
             let rustup_string = if self.is_rustup {
                 "It looks like Rustup is being used.".to_owned()
             } else {
-                format!("It looks like Rustup is not being used. For non-Rustup setups, the {} target needs to be installed manually. See https://rustwasm.github.io/wasm-pack/book/prerequisites/index.html#target-wasm32-unknown-unknown on how to do this.", target)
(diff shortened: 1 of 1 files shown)


drager:
@ashleygwilliams Yes, I think that would be a good idea

PR #623, opened 2019-04-11
generate

### LINKED ISSUE #373: Bundle `cargo generate` temporarily for ease of installation
## 💡 Feature description

One of the [first steps](https://rustwasm.github.io/wasm-pack/book/project-setup/using-a-template.html) in the getting started guide is to `cargo install cargo-generate`, but this can take quite some time and also be difficult to execute in all circumstances (OpenSSL issues, libgit2 issues, etc). 

It'd be awesome if instead `wasm-pack` bundled the code for `cargo-generate` by default (by using a dependeny on the actual `cargo-generate`). That way we could have something like `wasm-pack generate` which is immediately available after downloading wasm-pack. Eventually this wouldn't be necessary if `cargo generate` becomes more first class, but we could have a note saying that it's just a temporary measure for now!

ashleygwilliams:
fixes #373 

- [x] todo: respect `noinstall`
- [x] todo: lookup version of latest
- [x] docs
- [x] tests


ashleygwilliams:
this is still a WIP since i need to sort the testing but i'd love some feedback on the approach. additionally @alexcrichton if you rebase your #625 i think you can take advantage of some of the refactoring i've done here

README.md
```diff
@@ -66,38 +67,10 @@ This project was started by [ashleygwilliams] and is co-maintained by [ashleygwi
...
-  ```
-
-5. Install this tool: `cargo install wasm-pack`
-6. Run `wasm-pack build`, optionally, pass a path to a dir or a scope (see above for details)
-7. This tool generates files in a `pkg` dir
-8. To publish to npm, run `wasm-pack publish`. You may need to login to the
+1. Install this tool: `cargo install wasm-pack`
+1. Run `wasm-pack generate`.
+1. `cd hello-wasm`
+1. Run `wasm-pack build`, optionally, pass a path to a dir or a scope (see above for details)
```
alexcrichton: Since this is the "quick start" section, perhaps the `, optionally, ...` could be removed and/or just link to the reference?

docs/src/commands/generate.md
```diff
@@ -0,0 +1,32 @@
...
+
+The template can be an address to a git repo that contains a [`cargo-generate`]
+template.
+
+## Name
+
+The `wasm-pack generate` command can be given and optional name argument, e.g.:
+
+```
+wasm-pack generate --name myproject
```
alexcrichton: I wonder if it might be better to `wasm-pack generate myproject`? That jives with `cargo new` and I think commands like `git init` as well.

src/generate.rs
```diff
@@ -0,0 +1,28 @@
...
+use emoji;
+use failure::{self, ResultExt};
+use std::process::Command;
+
+/// Run `cargo generate` in the current directory to create a new
+/// project from a template
+pub fn generate(template: &str, name: &str, download: &Download) -> Result<(), failure::Error> {
+    let bin_path = download.binary("cargo-generate")?;
+    let cargo_generate = format!(
+        "{} generate --git {} --name {}",
```
alexcrichton: I think this'll want to be:

```rust
Command::new(&bin_path)
    .arg("generate")
    .arg("--git").arg(&template)
    .arg("--name").arg(&name)
```

src/install/krate.rs
```diff
@@ -0,0 +1,24 @@
+use install::Tool;
+use serde::Deserialize;
+
+#[derive(Debug, Deserialize)]
+pub struct Krate {
```
alexcrichton: FWIW I typically only use "krate" for the literal name "crate", but other variations like "Crate" I tend to use instead of "Krate" b/c it's valid Rust both ways (only `crate` is invalid)

src/install/mod.rs
```diff
@@ -0,0 +1,245 @@
...
+pub use self::mode::InstallMode;
+pub use self::tool::Tool;
+
+/// Install a cargo CLI tool
+///
+/// Prefers an existing local install, if any exists. Then checks if there is a
+/// global install on `$PATH` that fits the bill. Then attempts to download a
+/// tarball from the GitHub releases page, if this target has prebuilt
+/// binaries. Finally, falls back to `cargo install`.
+pub fn install(
```
alexcrichton: Perhaps `install` => `cargo_install_or_download_prebuilt`?

(to emphasize that this only works for cargo-based tools due to the fallback)

Cargo.toml
```diff
@@ -23,6 +23,7 @@ glob = "0.2"
 log = "0.4.6"
 openssl = { version = '0.10.11', optional = true }
 parking_lot = "0.6"
+reqwest = "0.9.14"
```
alexcrichton: Could this perhaps use `curl` to use the same downloading backend as `binary-install`?

README.md
```diff
@@ -66,38 +67,10 @@ This project was started by [ashleygwilliams] and is co-maintained by [ashleygwi
...
-      alert(&format!("Hello, {}!", name));
-  }
-  ```
-
-5. Install this tool: `cargo install wasm-pack`
-6. Run `wasm-pack build`, optionally, pass a path to a dir or a scope (see above for details)
-7. This tool generates files in a `pkg` dir
-8. To publish to npm, run `wasm-pack publish`. You may need to login to the
+1. Install this tool: `cargo install wasm-pack`
+1. Run `wasm-pack generate`.
```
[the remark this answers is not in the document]
ashleygwilliams: yeah, tbh i think this being in the README is a relic of the very very old days of this project. perhaps i'll make a new doc for quickstart and link it here?

docs/src/commands/generate.md
ashleygwilliams: so the reason i did a flag was that you can also pass --template.. we could have name be positional tho and still have the other flags. i think conforming to similar tools makes more sense.

src/generate.rs
ashleygwilliams: ack! yes indeed.

src/install/mod.rs
ashleygwilliams: ah yeah very good call. when we merge #625 with this that'll be important.

docs/src/commands/generate.md
alexcrichton: Ah yeah I meant just having `--name` be a positional argument instead (but keeping `--template`). We could perhaps take a leaf out of `git clone`'s book though and do something like:

```
# generates a new project named "foo"
$ wasm-pack generate https://github.com/alexcrichton/foo

# generates a new project named "bar"
$ wasm-pack generate https://github.com/alexcrichton/foo bar
```

I don't really feel too strongly any direction, but I do think that having the name be a positional argument rather than a flag might be nice.

docs/src/commands/generate.md
```diff
@@ -0,0 +1,32 @@
+# wasm-pack build
```
ashleygwilliams: fix this

alexcrichton:
To bikeshed a little bit, if we ignore backwards compatibility and that `wasm-pack init` was historically a subcommand, would be perhaps best named as `wasm-pack init` or `wasm-pack new`? This seems akin to `cargo new` and `cargo init` to me (which arguably both should take template parameters)

ashleygwilliams:
@alexcrichton i honestly had the same feeling when i was making this PR, `new` and `init` seem better and are a lot easier to type.

i think init has been deprecated for a lot of cycles at this point. i'd be open to replacing it, or we can go with `new`.

docs/src/commands/generate.md
ashleygwilliams: i really like this.

alexcrichton:
I think I'd vote for `new` over `init` since it's creating directories rather than initializing pre-existing ones (sorta like cargo `new`/`init`), but I'm not sure how much precedent there is for that distinction honestly

src/generate.rs
```diff
@@ -0,0 +1,25 @@
...
+/// project from a template
+pub fn generate(template: &str, name: &str, download: &Download) -> Result<(), failure::Error> {
+    let bin_path = download.binary("cargo-generate")?;
+    let mut cmd = Command::new(&bin_path);
+    cmd.arg("generate");
+    cmd.arg("--git").arg(&template);
+    cmd.arg("--name").arg(&name);
+
+    println!(
+        "{} Generating a new rustwasm worker project with name '{}'...",
```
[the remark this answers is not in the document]
ashleygwilliams: ah this is a copy pasta from when i was first implementing this in wrangler, good catch! will fox

Cargo.toml
ashleygwilliams: would you mind if we did this in another PR- it's... a little tricky with the interface and i'm worried this PR is getting pretty big cc @drager

docs/src/commands/generate.md
ashleygwilliams: i think this would be a neat feature but let's put it off til another PR as i'm worried this one is getting a lil big

Cargo.toml
drager: Personally I think that would be fine, as long as we have a tracking issue for it and planning to do it in the near future.

alexcrichton: Ok, deferring it to an issue seems fine

alexcrichton:
One recommendation I might have for testing comes from Cargo's test suite which is to use local file paths to git repositories. The general idea would be to make a builder-style API which creates a bunch of files in a directory, and then has a `build()` method which writes it all out and then turns it into a git repository with one commit as well. The result of `build()` could then be a `Url` which is actually just a file path, and then `wasm-pack generate` is invoked with that URL to ensure that it basically calls `cargo generate` and the right results show up (pass in a few different names, etc)

The various git helpers for Cargo are located in https://github.com/rust-lang/cargo/blob/master/tests/testsuite/support/git.rs and largely rely on `git2`, but using the CLI here would probably also be fine.

drager:
@ashleygwilliams About the tests I agree with you that some more test would be nice to have. However, I think we can add those afterwards. Let's get this PR merged first.

Issue #620, opened 2019-04-05
Walk up parent directories to find the current crate


fitzgen:
When there isn't any crate directory specified, we should walk up directories until we see a Cargo.toml instead of giving up if the current directory has none.

I.e. we should make the first `wasm-pack test` attempt below Just Work:

```
~/src/dodrio/examples $ wasm-pack test --firefox --headless
Error: crate directory is missing a `Cargo.toml` file; is `.` the wrong directory?

~/src/dodrio/examples $ cd ..

~/src/dodrio $ wasm-pack test --firefox --headless
[INFO]: 🎯  Checking for the Wasm target...
   Compiling futures v0.1.26
   Compiling cfg-if v0.1.7
...
```

ashleygwilliams:
i agree and i wish cargo fmt did this every day <3

ashleygwilliams:
closed by #624

PR #624, opened 2019-04-12
Recurse up the path looking for the Cargo.toml

gameldar:
PR for #620 Walk up parent directories to find the current crate

When setting the crate path, if the default path of the current working directory is defined pass back a flag indicating the default was used
If the default is used, and a path is not specified, then walk up the path to find the Cargo.toml file

I'm relatively new to Rust and figured this was a good first step to getting to know the project and wet my feet, but I'm not sure I'm doing things the idiomatic way - so I'm happy to have all and any feedback!

### DIFF
--- a/src/command/utils.rs
+++ b/src/command/utils.rs
@@ -8,8 +8,30 @@ use walkdir::WalkDir;
 
 /// If an explicit path is given, then use it, otherwise assume the current
 /// directory is the crate path.
-pub fn set_crate_path(path: Option<PathBuf>) -> Result<PathBuf, failure::Error> {
-    Ok(path.unwrap_or_else(|| PathBuf::from(".")))
+pub fn get_crate_path(path: Option<PathBuf>) -> Result<PathBuf, failure::Error> {
+    match path {
+        Some(p) => Ok(p),
+        None => find_manifest_from_cwd(),
+    }
+}
+
+/// Search up the path for the manifest file from the current working directory
+/// If we don't find the manifest file then return back the current working directory
+/// to provide the appropriate error
+fn find_manifest_from_cwd() -> Result<PathBuf, failure::Error> {
+    let mut parent_path = std::env::current_dir()?;
+    let mut manifest_path = parent_path.join("Cargo.toml");
+    loop {
+        if !manifest_path.is_file() {
+            if parent_path.pop() {
+                manifest_path = parent_path.join("Cargo.toml");
+            } else {
+                return Ok(PathBuf::from("."));
+            }
+        } else {
+            return Ok(parent_path.to_owned());
+        }
+    }
 }
 
 /// Construct our `pkg` directory in the crate.
--- a/tests/all/manifest.rs
+++ b/tests/all/manifest.rs
@@ -4,6 +4,7 @@ use std::fs;
 use std::path::PathBuf;
 use utils::{self, fixture};
 use wasm_pack::command::build::Target;
+use wasm_pack::command::utils::get_crate_path;
 use wasm_pack::{self, license, manifest};
 
 #[test]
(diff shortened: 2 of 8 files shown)


src/manifest/mod.rs
```diff
@@ -214,15 +214,34 @@ pub struct ManifestAndUnsedKeys {
...
+    /// a manifest file is found, or return an error if one cannot be found.
+    pub fn new(
+        crate_path: &Path,
+        recurse: bool,
+        out_name: Option<String>,
+    ) -> Result<CrateData, Error> {
+        let full_path = crate_path.canonicalize()?;
+        let full_path = full_path.as_path();
+        let mut manifest_path = crate_path.join("Cargo.toml");
+        if recurse {
```
fitzgen: I think this logic should be in `set_crate_path` (which should probably be renamed *get* instead of *set*) so that we don't have to deal with a `bool` part in a returned tuple nor a `bool` parameter. It would be cleaner without that, IMO.

gameldar: That was the part that I thought was quite clunky but I was worried about the scope of the refactoring. As it turns of it was the opposite of what I thought - it is the smaller refactoring and is a lot cleaner and obvious. Thanks!

src/command/utils.rs
```diff
@@ -8,8 +8,34 @@ use walkdir::WalkDir;
...
+        None => find_manifest_from_cwd(),
+    }
+}
+
+/// Search up the path for the manifest file from the current working directory
+/// If we don't find the manifest file then return back the current working directory
+/// to provide the appropriate error
+fn find_manifest_from_cwd() -> Result<PathBuf, failure::Error> {
+    let cwd = PathBuf::from(".");
+    let parent_path = cwd.canonicalize()?;
```
fitzgen: I'm a little hesitant to add `canonicalize` calls here, since we have historically had issues with canonicalized paths on windows (and in non-english locales?) that I never fully understood the root causes for.

If this isn't required for correctness (I don't think any APIs we use require canonical paths, and work just fine with relative paths), then I'd like to remove the `canonicalize` call. Or maybe @ashleygwilliams or @alexcrichton understood the bug we hit better than I do.

Sorry I missed this in the first round of review!

src/command/utils.rs
```diff
@@ -8,8 +8,34 @@ use walkdir::WalkDir;
...
+    let mut parent_path = parent_path.as_path();
```
fitzgen: This call should be unnecessary, and I think we can wholesale remove this line. Since `PathBuf` derefs to `Path` the `.join` call below should Just Work(tm).

src/command/utils.rs
```diff
@@ -8,8 +8,34 @@ use walkdir::WalkDir;
...
+        if !manifest_path.is_file() {
+            manifest_path = match parent_path.parent() {
+                Some(p) => {
+                    parent_path = p;
+                    p.join("Cargo.toml")
+                }
+                None => break Ok(cwd),
+            };
+        } else {
+            break Ok(parent_path.to_path_buf());
```
fitzgen: These `break`s should be `return`.

When it is `break`, I double check if there is a `let foo = loop { ... };` but there isn't, and then I realize it is the last expression in the function, so it is equivalent to `return`... and we can avoid readers doing that same little bit of confusion+understanding in the future by using `return`.

fitzgen:
AppVeyor failure:

```
error: package collision in the lockfile: packages project_b v0.1.0 (C:\projects\wasm-pack-071k0\target\t\.tmpFEj1Z4\wasm-pack\project_b) and project_b v0.1.0 (C:\projects\wasm-pack-071k0\target\t\.tmpFEj1Z4\wasm-pack\main\../project_b) are different, but only one can be written to lockfile unambiguously
```

Maybe related to `canonicalize` or are we manually add "/" path separators somewhere? Double checking the PR...

tests/all/manifest.rs
```diff
@@ -481,3 +482,46 @@ fn it_lists_license_files_in_files_field_of_package_json() {
...
+            authors = ["The wasm-pack developers"]
+            description = "so awesome rust+wasm package"
+            license = "WTFPL"
+            name = "recurse-for-manifest-test"
+            repository = "https://github.com/rustwasm/wasm-pack.git"
+            version = "0.1.0"
+            homepage = "https://rustwasm.github.io/wasm-pack/"
+        "#,
+    );
+    let path = get_crate_path(Some(PathBuf::from("src/"))).unwrap();
```
fitzgen:
```suggestion
    let path = get_crate_path(Some(PathBuf::from("src"))).unwrap();
```

Let's avoid hard coding path separators.

fitzgen:
So the `it_should_build_nested_transitive_dependencies` test has hard coded file paths:

```
            [dependencies]
            wasm-bindgen = "0.2"
            project_a = { path = "../project_a" }
            project_b = { path = "../project_b" }
```

and I think this is fine and cargo understands how to translate unix separators into windows separators *so long as the path is not canonical(/absolute/has windows separators?).* So I think this test failure will be fixed by removing the call to `canonicalize` which will in turn make our underlying `cargo` invocation do the Right Thing Automatically For Us(tm).

src/command/utils.rs
alexcrichton: I think this shoudl be fixable by basically just using `env::current_dir()` rather than `canonicalize` which should return an appropriate absolute path

gameldar: Before refactoring I did need to use canonicalize - the problem was that a relative path coming into the function (e.g. PathBuff::from(".") that was the default) - isn't absolute so parent_path.parent() returns None even if you are not in the root directory. I needed to canonicalize it to ensure I had an absolute path.

However the logic changed after migrating the code into the the get_crate_path as now we take the current working directory and then work back from there. So thanks @alexcrichton using env::current_dir() gives the absolute path  needed. Looking back at the output on Windows of the cwd after it was canonicalized it had a whole bunch of escaping included - compared to the absolute path.

src/command/utils.rs
gameldar: Ah yes I see now - after more digging due to conflicting type errors while trying to address this I actually found the pop method on PathBuf... which is exactly what I was doing but better!

src/command/utils.rs
gameldar: I agree - this was my attempt to do things the 'idiomatic' way - it seems from what I've been reading rustwise the preferred thing is to not have explicit return calls. I personally prefer explicit use of return because it makes it clear that is the intention (although some of that is that half the time I automatically add the semi colon and wonder why the compiler is saying I forgot to return a value from the function)

gameldar:
I asked this on discord but it disappeared up into the ether:
What is the policy is for cleaning up the PR? E.g. should I squash and do a force push to my fork to bring it back down to one commit?

fitzgen:
>  What is the policy is for cleaning up the PR? E.g. should I squash and do a force push to my fork to bring it back down to one commit?

I don't think we have an official policy for wasm-pack, but my rule of thumb is that if the follow up commits stand on their own (make sense outside the context of the PR, pass CI without relying on commits that come after, etc) then I leave them be, otherwise I squash them.

ashleygwilliams:
this is great! sorry for the delay and thanks for your patience. this will go out in 0.9.0!

PR #625, opened 2019-04-12
Add support for automatically executing `wasm-opt`

### LINKED ISSUE #159: Add support for running wasm-opt

alexcrichton:
This commit adds support for automatically executing the `wasm-opt`
binary from the [Binaryen project][binaryen]. By default `wasm-pack`
will now, in release and profiling modes, execute `wasm-opt -O` over the
final binary. The goal here is to enable optimizations that further
reduce binary size or improve runtime. In the long run it's expected
that `wasm-opt`'s optimizations may mostly make their way into LLVM, but
it's empirically true today that `wasm-opt` plus LLVM is the best
combination for size and speed today.

A configuration section for `wasm-opt` has been added as [previously
proposed][fitzgen], namely:

```toml
[package.metadata.wasm-pack.profile.release]
wasm-opt = ['-Os']
```

The `wasm-opt` binary is downloaded from Binaryen's [own
releases](https://github.com/webassembly/binaryen/releases). They're
available for the same platforms that we download predownloaded binaries
for `wasm-bindgen` on. We'll also opportunistically use `wasm-opt` in
`PATH` if it's available. If we're untable to run `wasm-opt`, though, a
warning diagnostic is printed informing such.

Closes #159

[binaryen]: https://github.com/webassembly/binaryen
[fitzgen]: https://github.com/rustwasm/wasm-pack/issues/159#issuecomment-454888890


ashleygwilliams:
i'd like to merge this with the refactor in #623, so gonna wait til that lands (which should be early next week, if not earlier)

docs/src/cargo-toml-configuration.md
```diff
@@ -9,6 +9,17 @@ the `--dev`, `--profiling`, and `--release` flags passed to `wasm-pack build`.
...
 ```toml
+[package.metadata.wasm-pack.profile.dev]
+# Should `wasm-opt` be used to further optimize the wasm binary generated after
+# the Rust compiler has finished? Using `wasm-opt` can often further decrease
+# binary size or do clever tricks that haven't made their way into LLVM yet.
+#
+# Configuration can be set to `false` if you want to disable it (as is the
+# default for the dev profile), or it can be an array of strings which are
+# explicit arguments to pass to `wasm-opt`. For example `['-Os']` would optimize
+# for size while `['-O4']` would execute very expensive optimizations passes
```
fitzgen: This is missing "size" and "speed" meta options [that we planned in the original design](https://github.com/rustwasm/wasm-pack/issues/159#issuecomment-454888890):

> ```toml
> # This is the dev profile, but could also be the profiling or release profiles.
> [package.metadata.wasm-pack.profile.dev]
> # The `wasm-opt` key may be absent, in which case we choose a default
> #
> # or we can explicitly configure that we *don't* want to run it
> wasm-opt = false
> # or use our default alias to optimize for size
> wasm-opt = "size"
> # or use our default alias to optimize for speed
> wasm-opt = "speed"
> # or give your own custom set of CLI flags
> wasm-opt = ["--dce", "--duplicate-function-elimination", "--instrument-memory"]
> ```

Is the intention to add support for these in follow up PRs?

src/lib.rs
```diff
@@ -33,6 +33,7 @@ pub mod license;
 pub mod lockfile;
 pub mod manifest;
 pub mod npm;
+pub mod opt;
```
fitzgen: Maybe name this module `wasm_opt`? I know that we name the `wasm-bindgen`-related module just `bindgen` but I think in that case it is obvious what `bindgen` we are talking about, where as `opt` is not obviously `optimize` (as opposed to `options`) to me, and even if it was obviously `optimize` it isn't also obviously the-specific-tool-`wasm-opt`(imize)-and-not-generic-optimization-passes to me.

docs/src/cargo-toml-configuration.md
alexcrichton: True! I forgot to mention that here, but I actually first implemented it as a string being a space-separated list of arguments (like `wasm-opt = "-Os -some-other-pass"`), and then I read the original issue and liked your idea better.

I figured though it may be best to do that as a separate PR as it should be pretty easy to add on top of this, but rather we'd just want to get the bare bones in here

src/lib.rs
alexcrichton: 👍

alexcrichton:
Ping for review on this? I'm not sure if there's anything still blocking this?

ashleygwilliams:
i was blocking this on #623, but i think we can land this and i'll rebase #623 and add the updates that let this use the new more generic logic for binstall.

ashleygwilliams:
taking on the merge for this- gonna rewrite a bit to use the general logic in #623

ashleygwilliams:
gonna merge this and then refactor

ashleygwilliams:
this is failing on a cargo fmt, since i'm about to do #685 i think it's fine and gonna merge

torch2424:
@alexcrichton @ashleygwilliams 

So i was working on a PR, and I think this is breaking functionality for the current master? I get the following error:

Screenshot from 2019-07-29 07-36-12

Is there a way to skip the step? Or will this be fixed by #685 ?

Thanks! :smile:

alexcrichton:
@torch2424 you should be able to cofnigure:

```toml
[package.metadata.wasm-pack.profile.dev]
wasm-opt = false
[package.metadata.wasm-pack.profile.release]
wasm-opt = false
```

to turn off `wasm-opt` support. Mind also opening an issue for that bug you're running into? It's an upstream issue with `wasm-opt`, but newer releases of `wasm-opt` may have fixed the issue.

torch2424:
@alexcrichton Awesome, will try that later tonight, and update here 😄 

I'll open the issue then too, since a newer version may have fixed this like you said (and I assume we will consume this newer version 👍 )

torch2424:
Oh after looking at: `src/wasm_opt.rs` I now realize the version we are using is hardcoded? Opening an issue now :smile:

jotyy:
@alexcrichton  `-O3` is not supported?

It works wrong with this configuration:
```
[profile.release]
opt-level = 3

[package.metadata.wasm-pack.profile.release]
wasm-opt = ["-O3"]
```

PR #630, opened 2019-04-16
Only run specific tests during `wasm-pack test`

alexcrichton:
This builds on rustwasm/wasm-bindgen#1465 by ensuring that `wasm-pack
test --node` will now only test node.js tests


alexcrichton:
In a sense this is actually a bugfix to update to what the CLI flags actually mean (when you pass `--node` only running Node tests), so I think the documentation largely already made sense but I tweaked a bit to make sure this is explicitly mentioned.

Issue #627, opened 2019-04-12
check wasm-bindgen version if target = web is passed


ashleygwilliams:
we've seen some people get confused by the error message about an incorrect argument when wasm-pack build --target web is run on a project that depends on a wasm-bindgen that didn't support that target. we should give an error message that tells folks to update their wasm-bindgen dep in their Cargo.toml.

gameldar:
After looking through the wasm-bindgen changelog to find the appropriate minimum version of wasm-bindgen to depend upon I was going to create a new issue because --web has been deprecated then found #597 - should I list the minimum as 0.2.40 rather than 0.2.39 so we can potentially move ahead with #597 sooner?

EverlastingBugstopper:
was this fixed by https://github.com/rustwasm/wasm-pack/commit/97fdc94d535b0f1c3be6aad6424da12267295367? Looks like it was committed directly to master - if so we should close this

EverlastingBugstopper:
@ashleygwilliams this was closed by #684 - we should close this issue

PR #633, opened 2019-04-21
custom error message when wasm-bindgen fails because --web isn't available

gameldar:
PR for #627 - check wasm-bindgen version if target = web is passed

create a custom error message when wasm-bindgen fails because --web isn't available

Change the return type of the child run commands so it can return a CommandError that also captures the stderr if the process fails.
See if the returned output contains the 'Unknown flag: --web' error message and if so provide an alternative error that suggests upgrading


ashleygwilliams:
gonna take this over and do the rebase and add a test

ashleygwilliams:
closing in favor of https://github.com/rustwasm/wasm-pack/pull/684

PR #640, opened 2019-05-04
Updated help message for `build` and `publish` subcommands

### LINKED ISSUE #636: Help message for wasm-pack build is not updated
## 🐛 Bug description
In help message `target` still defaults to `browser`

More details:
`wasm-pack build --help` prints this:
```
-t, --target <target>        Sets the target environment. [possible values: browser, nodejs, web, no-modules]
                                 [default: browser]
```

In the [book](https://rustwasm.github.io/docs/wasm-pack/commands/build.html#target) the default is `bundler`

Actually, both are supported, but `--target browser` silently became deprecated
https://github.com/rustwasm/wasm-pack/blob/master/src/command/build.rs#L97

#### 🤔 Expected Behavior
`wasm-pack build --help` should print that `bundler` is the default target, and maybe include some information about deprecation of `browser` target

#### 🌍 Your environment
wasm-pack version: 0.8.1

I'd like to submit a pr if this update makes sense

ibaryshnikov:
Closes #636 

Should we also add a deprecation warning for those who still use `browser`?

There's also an interesting line here:
https://github.com/rustwasm/wasm-pack/blob/master/src/bindgen.rs#L204
Update it to `--bundler`?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,10 @@
 # Changelog
 
+## Unreleased
+- ###  Fixes
+
+  - Updated help message for `build` and `publish` subcommands. Replaced `browser` with the `bundler`
+
 ## 🛠️ 0.8.1
 
 - ### 🤕 Fixes
--- a/src/command/build.rs
+++ b/src/command/build.rs
@@ -135,8 +135,8 @@ pub struct BuildOptions {
     /// this flag will disable generating this TypeScript file.
     pub disable_dts: bool,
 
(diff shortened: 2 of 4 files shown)


CHANGELOG.md
```diff
@@ -1,5 +1,10 @@
 # Changelog
 
+## Unreleased
```
ashleygwilliams: appreciate you doing this- i've considered asking for changelog entries in PRs but... it turns out that the merge conflicts get really gnarly. gonna remove this and then merge this in

PR #649, opened 2019-05-19
Fix sideEffects value in generated package.json should be boolean

rhysd:

```bash
```
- [x] You reference which issue is being closed in the PR text: There is no issue related to this fix

I'm not heavy user of webpack but `sideEffects` field of `package.json` seems boolean or array value.

https://webpack.js.org/guides/tree-shaking/#mark-the-file-as-side-effect-free

But current wasm-pack sets `"false"` string value to the field. This patch fixes the point.

### DIFF
--- a/tests/all/build.rs
+++ b/tests/all/build.rs
@@ -28,17 +28,17 @@ fn it_should_build_crates_in_a_workspace() {
             "Cargo.toml",
             r#"
                 [workspace]
-                members = ["blah"]
+                members = ["it_should_build_crates_in_a_workspace"]
             "#,
         )
         .file(
-            Path::new("blah").join("Cargo.toml"),
+            Path::new("it_should_build_crates_in_a_workspace").join("Cargo.toml"),
             r#"
                 [package]
                 authors = ["The wasm-pack developers"]
                 description = "so awesome rust+wasm package"
                 license = "WTFPL"
-                name = "blah"
+                name = "it_should_build_crates_in_a_workspace"
                 repository = "https://github.com/rustwasm/wasm-pack.git"
                 version = "0.1.0"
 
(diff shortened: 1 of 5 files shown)


rhysd:
CI failed due to the reason not related to changes of this PR.

https://travis-ci.com/rustwasm/wasm-pack/jobs/201309430

`wasm-pack build` failed due to

> Error: Directory not empty (os error 39)

I guess this is because each test case use the same directory name for test. As you know, Rust runs tests parallelly. So when using the same directory for multiple tests, the directory may be (or may not be) used by other test cases. It would cause some flakiness on testing. I'll try to resolve it and check if my guess is correct or not

rhysd:
Travis CI job is now back to normal. Failure on AppVeyor also occurs on master branch so it is not related to changes on this PR.

tests/all/build.rs
```diff
@@ -28,17 +28,17 @@ fn it_should_build_crates_in_a_workspace() {
             "Cargo.toml",
             r#"
                 [workspace]
-                members = ["blah"]
```
ashleygwilliams: i'm not sure i understand this change- can you explain? otherwise this looks good!

ashleygwilliams:
gonna snag this PR and try to work through the test failures. thanks for noticing this! it's a BIG thing to catch!

ashleygwilliams:
cannot replicate this locally. can you perchance @steveklabnik ? this is failing on the workspaces test: https://ci.appveyor.com/project/ashleygwilliams/wasm-pack-071k0/builds/24651407#L394

steveklabnik:
I get eight test failures on both master and with this PR, so I don't think this PR is causing it directly.

ashleygwilliams:
i'm gonna remove the commit on this that attempted to address the failures and then merge since i dont think it is related.. we'll see what CI says

ashleygwilliams:
closing in favor of https://github.com/rustwasm/wasm-pack/pull/680

PR #658, opened 2019-05-25
fix: Add user agent for wasm-pack version check

### LINKED ISSUE #651: wasm-pack's version checking calls violate crates.io's crawler policy
## 🐛 Bug description

When wasm-pack checks for a new version, it does not provide any user-agent header. crates.io recently rolled out a change which blocks all traffic which does this, as it violates our policies:

> We also require all crawlers to provide a user-agent header that allows us to uniquely identify your bot. This allows us to more accurately monitor any impact your bot may have on our service. Providing a user agent that only identifies your HTTP client library (such as "request/0.9.1") increases the likelihood that we will block your traffic. It is recommended, but not required, to include contact information in your user agent. This allows us to contact you if we would like a change in your bot's behavior without having to block your traffic.

> Bad: "User-Agent: reqwest/0.9.1"
> Better: "User-Agent: my_bot"
> Best: "User-Agent: my_bot (my_bot.com/info)" or "User-Agent: my_bot (help@my_bot.com)"

#### 🤔 Expected Behavior
A user agent header, such as "wasm-pack version check" should be provided. Alternatively, this check should be done against the index, which contains all the information required to check if a more recent version is available without hitting the API

#### 👟 Steps to reproduce
- Publish a new version of wasm-pack
- Run wasm-pack
- See that no new version warning is emitted

#### 🌍 Your environment
wasm-pack version: n/a
rustc version: n/a

drager:
This PR adds a user agent header for the wasm-pack version check. It will look like this: `User-Agent: wasm-pack/0.8.1 (https://github.com/rustwasm/wasm-pack)`.

Fixes #651.

### DIFF
--- a/src/manifest/mod.rs
+++ b/src/manifest/mod.rs
@@ -26,6 +26,8 @@ use toml;
 use PBAR;
 
 const WASM_PACK_METADATA_KEY: &str = "package.metadata.wasm-pack";
+const WASM_PACK_VERSION: Option<&'static str> = option_env!("CARGO_PKG_VERSION");
+const WASM_PACK_REPO_URL: &str = "https://github.com/rustwasm/wasm-pack";
 
 /// Store for metadata learned about a crate
 pub struct CrateData {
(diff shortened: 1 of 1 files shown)


fitzgen:
It seems like CI has been failing for all the PRs recently and seems sort of related to `binary-install`. I saw some github emails in that repo, but haven't dug in yet. Do you happen to know what's up?

drager:
@fitzgen Hmm, I haven't really dug in the latest changes for binary-install just yet. Maybe @ashleygwilliams have an idea?

This is the test that fails though: https://github.com/rustwasm/wasm-pack/blob/b07ec958d210a2dddf9e7ce348253d7cc22b7a19/tests/all/build.rs#L24-L69

ashleygwilliams:
merging as this change is not related to the CI failure

Issue #653, opened 2019-05-21
wasm-pack does not rate limit its version check if the request fails


sgrif:
## 🐛 Bug description
When wasm-pack looks for a more recent version, it limits this check to once every 24 hours normally. When the request fails, however, it will try again on the next run. Based on our logs, there are folks who are running wasm-pack in a loop on their home network. Since all of these version checks are currently being blocked (#651), this has resulted in a substantial enough spike in 4xx traffic to be noticeable on our graphs.

 If the reason this failed is because crates.io has gone down due to being overloaded, this can amplify the problem. There should be some form of exponential backoff in the case of an error.

#### 🤔 Expected Behavior
wasm-pack should not immediately retry its version check if the last one failed.

fitzgen:
It seems fine to me to apply the 24 hours rate limiting regardless if the check succeeded or failed.

fitzgen:
+cc @drager @daubaris

drager:
Thanks for bringing this up @sgrif! I will try to get a PR that fixes this.

drager:
Fixed in #660.

PR #660, opened 2019-05-26
fix: Rate limit version check if the request fails and output warning if version check fails

### LINKED ISSUE #652: wasm-pack provides no output when its version check fails
## 🐛 Bug description
When wasm-pack checks to see if a more recent version is released, the error is silently swallowed. Since this is an optional behavior, it makes sense to not bother the user if they're not online or something, but (as documented in #651) -- this check *always* fails, which has been the case for a few days, and gone unnoticed.

It's unlikely that a request to crates.io will complete with a non-200 status code, it seems reasonable to at least inform the user something happened in this case.

drager:
This PR will fix #652 and #653.

### DIFF
--- a/src/build/mod.rs
+++ b/src/build/mod.rs
@@ -62,12 +62,7 @@ fn rustc_minor_version() -> Option<u32> {
 /// Checks and returns local and latest versions of wasm-pack
 pub fn check_wasm_pack_versions() -> Result<WasmPackVersion, Error> {
     match wasm_pack_local_version() {
-        Some(local) => {
-            match Crate::return_wasm_pack_latest_version() {
-                Some(latest) => Ok(WasmPackVersion {local, latest}),
-                None => Ok(WasmPackVersion {local, latest: "".to_string()})
-            }
-        },
+        Some(local) => Ok(WasmPackVersion {local, latest: Crate::return_wasm_pack_latest_version()?.unwrap_or_else(|| "".to_string())}),
         None => bail!("We can't figure out what your wasm-pack version is, make sure the installation path is correct.")
     }
 }
--- a/src/manifest/mod.rs
+++ b/src/manifest/mod.rs
@@ -149,39 +149,49 @@ struct CrateInformation {
 
 impl Crate {
     /// Returns latest wasm-pack version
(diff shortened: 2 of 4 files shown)


src/build/mod.rs
```diff
@@ -64,8 +64,13 @@ pub fn check_wasm_pack_versions() -> Result<WasmPackVersion, Error> {
...
             match Crate::return_wasm_pack_latest_version() {
-                Some(latest) => Ok(WasmPackVersion {local, latest}),
-                None => Ok(WasmPackVersion {local, latest: "".to_string()})
+                Ok(latest) => {
+                    match latest {
+                        Some(latest) => Ok(WasmPackVersion {local, latest}),
+                        None => Ok(WasmPackVersion {local, latest: "".to_string()})
+                    }
+                },
+                Err(err) => Err(err)
```
fitzgen: This error propagation could be done slightly easier and more idiomatically with `?` and then providing a default for `None` is a bit nicer with `unwrap_or_else`. Put it together and we get:

```rust
Ok(Crate::return_wasm_pack_latest_version()?.unwrap_or_else(|| WasmPackVersion { ... }))
```

src/manifest/mod.rs
```diff
@@ -141,39 +141,45 @@ struct CrateInformation {
...
-            Some(v)
-        })
+    fn return_api_call_result(
+        current_time: DateTime<offset::Local>,
+    ) -> Result<String, failure::Error> {
+        let version = Self::return_latest_wasm_pack_version();
+
+        match version {
+            Ok(ref version) => Self::override_stamp_file(current_time, Some(&version)).ok(),
+            Err(_) => Self::override_stamp_file(current_time, None).ok(),
```
fitzgen: Probably worth a comment here explaining why we override the time stamp file even when getting the version fails.

src/build/mod.rs
drager: Good suggestion, will do that. Thanks!

src/manifest/mod.rs
drager: Absolutely!

drager:
@fitzgen @ashleygwilliams I think this is good to go now!

PR #671, opened 2019-06-29
fix installer to allow for custom install locations

### LINKED ISSUE #470: installer: do not attempt to install to /usr/bin
On Arch Linux the `rustup` package is installed globally, but that doesn't mean that wasm-pack should also do so. /usr/bin/ is basically for packages only, and manually installed binaries should at least go /usr/local/bin.

Instead, we should always try to install to the Cargo binary directory when rustup is detected.

dhpollack:

- [ X ] You have the latest version of `rustfmt` installed
```bash
```
- [ X ] You ran `cargo fmt` on the code base before submitting
- [ X ] You reference which issue is being closed in the PR text

This fixes #470 and install wasm-pack to the `CARGO_HOME` environmental variable if it is set.  I guess one minor note is that I've made it so that the environmental variable is checked before check for `rustup`.  This has the advantage that users who installed rust without `rustup` can still install `wasm-pack` as `rustup` is not an actual requirement.

### DIFF
--- a/docs/_installer/index.html
+++ b/docs/_installer/index.html
@@ -51,7 +51,7 @@
           <code>curl https://rustwasm.github.io/wasm-pack/installer/init.sh -sSf | sh</code>
           <p>
             You appear to be running a *nix system (Unix, Linux, MacOS). If not,
-            <a class="default-platform-button" href="#">display all supported installers</a>.
+            <a class="default-platform-button" href="#">display all supported installers</a>. The default location will be $CARGO_HOME/bin or the location of rustup if CARGO_HOME is not set.
           </p>
         </div>
       </div>
--- a/src/installer.rs
+++ b/src/installer.rs
@@ -56,6 +56,7 @@ fn do_install() -> Result<(), failure::Error> {
     match cargo_home {
         Ok(path) => {
             installation_dir.push(path);
+            installation_dir.push("bin");
         }
         Err(_) => {
             println!("$CARGO_HOME not set, default to rustup location for install");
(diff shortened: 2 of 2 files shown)


dhpollack:
Another solution to this problem would be to keep the current instructions, but in the README give instructions for building from source.  Something like...

```sh
cargo install --root $CARGO_HOME
```

Personally, I found it a bit esoteric what the installer script is actually doing.  At first, I was editing the installer script to manually set a dir there, which was a bit of a disaster, because I deleted my `.local/bin` dir...

Having said all this, once I got 'wasm-pack` installed, I found it super cool.

dhpollack:
@ashleygwilliams any comment on this?

ashleygwilliams:
hey @dhpollack ! i think we'd be happy to take this PR. thanks for the effort :) would you be willing to add some documentation so folks know that this is a feature?

ashleygwilliams:
@dhpollack i think the best place would be the installation page (https://github.com/rustwasm/wasm-pack/blob/master/docs/_installer/index.html) !

dhpollack:
I added a line to the documentation page.  I also changed the location of the install to `$CARGO_HOME/bin` since the `CARGO_HOME` variable defaults to something like `~/.cargo` per [the rust docs](https://doc.rust-lang.org/cargo/reference/environment-variables.html).  

I think this is good to go tho.

dhpollack:
@ashleygwilliams this is ready to go

dhpollack:
@ashleygwilliams any news on this?

dhpollack:
stale

PR #678, opened 2019-07-16
#640 minus changelog

ashleygwilliams:
#640 minus the changelog


ibaryshnikov:
@ashleygwilliams what do you think about updating this line?
https://github.com/rustwasm/wasm-pack/blob/master/src/bindgen.rs#L45
which is currently `Target::Bundler => "--browser",`

ashleygwilliams:
CI failure on debug assertions is not related to this PR

ashleygwilliams:
whoops @ibaryshnikov totally missed your comment. i'll get it in!

PR #680, opened 2019-07-16
#649 without the test update commit

ashleygwilliams:
#649 without the test update commit because the test failure seems spurious


ashleygwilliams:
need to make sideEffects optional and then this can be good to go
the error on this is hilarious right now if anyone wants to check it out https://travis-ci.com/rustwasm/wasm-pack/jobs/216625948#L454

PR #684, opened 2019-07-16
check wasm-bindgen version if running with target web

ashleygwilliams:
#633 with merge, edits, and test


ashleygwilliams:
CI should fail on branch but not PR on debug assertions test bug that has already been fixed on master

alexcrichton:
👍

PR #694, opened 2019-07-24
Adding in `--quiet` and `--log-level` flags

Pauan:


Pauan:
> I wonder however if --quiet should mean that we don't emit any output at all and just use the exit code.

I did consider that, however I *do* actually want the `cargo build` output to show, since building a crate can take a long time and the user wants to see progress being made.

(It's possible to achieve full silence by using `wasm-pack --quiet build -- --quiet` though that's very non-obvious)

So maybe this should actually be a different flag, something like `--message-format none` (analogous to `cargo --message-format`)?

Pauan:
Okay, I changed `--quiet` so it also silences output from `cargo`.

And I added in a new `--log-level` flag which controls wasm-pack's log level. So the previous behavior of `--quiet` can be achieved by using `--log-level error`.

Pauan:
> can we add a test and some docs?

I don't mind, but the existing CLI flags (e.g. `verbose`) don't have that. So where would I put it?

Pauan:
(Just a heads up that I added in `-q` as a short-hand for `--quiet`, which is consistent with `cargo`)

ashleygwilliams:
@Pauan i'd probably create a new doc under commands about log level! https://rustwasm.github.io/wasm-pack/book/commands/index.html

Pauan:
@ashleygwilliams I added in docs and unit tests.

Pauan:
@ashleygwilliams Ping.

Pauan:
@ashleygwilliams Ping.

niedzielski:
This is a subtle but important change as verbose informational output is useful for debugging but may conceal important issues. ([Silence is golden!](https://en.wikipedia.org/wiki/Unix_philosophy#Eric_Raymond's_17_Unix_Rules))

Pauan:
@ashleygwilliams This is ready for review.

PR #695, opened 2019-07-25
[WIP] feat(build): --no-pack flag

### LINKED ISSUE #691: Mini-RFC: --no-pack flag
**Feature Description**
when calling `wasm-pack build` a user can optionally pass `--no-pack` and wasm-pack will build your wasm, generate js, and not build a package.json

**Example**
```
wasm-pack build --target no-modules --no-pack
```

**Outstanding Questions**

- Mode? should we stick this in as a mode? or should it be toplevel? docs for mode: 
   https://rustwasm.github.io/docs/wasm-pack/commands/build.html#mode
- the default outdir is called `pkg`.. i'd... probably keep this the default for `--no-pack` but it will certainly be Weird

ashleygwilliams:
fixes #691


ashleygwilliams:
assuming folks approve this i'll write docs for it, and then should be ready to merge

alexcrichton:
The implementation here all looks good to me, although I might bikeshed a bit on the name. Perhaps something like `--no-package-json` or `--no-npm-manifest` or something like that? I feel like this is largely centered around the "don't do the extra steps necessary to create a package.json" part of things.

Also as a future-compatible note, this will probably want to generate a hard error if a Rust crate depends on an NPM package (e.g. when the npm dependencies implementation lands here if that is combined with `--no-pack` we'd want to give a first-class "this project isn't compatible with that" error)

ashleygwilliams:
@alexcrichton yeah i thought about a manifest related type name, i'm happy to do that

if i am understanding the ask, i actually imagine the end user would *want* a package.json... it'd be necessary. we could still skip the cargo.toml serialization and the readme and license steps.. this convo might be best on the npm feature tracking issue tho

ashleygwilliams:
tho.. if we would eventually with npm feature support make a packagejson for this no pack flag... that makes the name more confusing.

ashleygwilliams:
i am waiting to sync up with @Pauan but i think we should probably just close this. when considered with the npm feature, this feature does not really make sense. 

- copying readme + licenses: doesn't block and doesn't happen if they aren't there. --quiet can silence the warnings
- serializing the cargo.toml: should not be expensive, if we are already making a partial package.json because the package depends on an npm package it feels weird to make a feature that just copies the wasm-bindgen one over and not just do the whole thing. if it is expensive, the feature request is perf related.. not a feature request.

fitzgen:
+1 for `--no-package-json`

src/command/build.rs
```diff
@@ -239,7 +246,7 @@ impl Build {
         Ok(())
     }
 
-    fn get_process_steps(mode: InstallMode) -> Vec<(&'static str, BuildStep)> {
+    fn get_process_steps(mode: InstallMode, no_pack: bool) -> Vec<(&'static str, BuildStep)> {
```
daxpedda:
```suggestion
    fn get_process_steps(&self) -> Vec<(&'static str, BuildStep)> {
```
Would give it access to both without having to pass any parameters at all.

BlackGlory:
Any progress?
I wouldn't even consider wasm-pack without this option.
In fact, it should be the default, the current `wasm-pack build` command makes it seem like the developers have bad taste.

Issue #697, opened 2019-08-04
build command with --out-name that contains dots, `.`, generates incorrect files


torch2424:
## 🐛 Bug description
When creating a PR for #313 I noticed that the `--out-name` flag does not work properly if it contains a `.` character. I assume this is also an issue with wasm-bindgen, but it would also affect what we do here.

I think this is better done by showing in examples. I'll use the `node` target since it requires the `_bg.js` shim, which also doesn't generate correctly.

**Examples:**

**Happy Case:**

Expected output

![Screenshot from 2019-08-04 02-26-18](https://user-images.githubusercontent.com/1448289/62421957-42c70b80-b65f-11e9-98ab-7254d314270c.png)

**Case where name contains `*.js`:**

* Correctly ignores the .js extension when generating the main files.
* Missing shim
* wasm file is incorrectly names
* package json references non-existant files

![Screenshot from 2019-08-04 02-27-06](https://user-images.githubusercontent.com/1448289/62421961-5ffbda00-b65f-11e9-9b35-7ba76519552a.png)

**Case where name contains `*.somerandomextension`:**

* Generated files ignore the extension
* Package.json references the correctly names files, but they were generated incorrectly

![Screenshot from 2019-08-04 02-32-16](https://user-images.githubusercontent.com/1448289/62422003-27a8cb80-b660-11e9-98ac-fab668683aba.png)

**Case where name contains `*.*.js`:**

* Correctly ignores the .js extension when generating the main files.
* Missing shim
* wasm file is incorrectly names
* package json references non-existant files

![Screenshot from 2019-08-04 02-30-35](https://user-images.githubusercontent.com/1448289/62421994-f92af080-b65f-11e9-8b49-0abb131a07d7.png)

#### 🤔 Expected Behavior

Whenever there is a `.` in the `--out-name`, it should respect the additional extensions, and then build off of them. There probably is some discussion points of:

* How should the `_bg` be handled?
* Should `.js` or `.wasm` be ignored?

#### 👟 Steps to reproduce

Please see the examples screenshots :smile: 

#### 🌍 Your environment
Include the relevant details of your environment.
wasm-pack version: current master
rustc version: 1.34.1

![Screenshot from 2019-07-31 01-11-56](https://user-images.githubusercontent.com/1448289/62195149-39246780-b330-11e9-9ca3-5fcf6a8af836.png)

pop_OS! is a derivative of Ubuntu :smile: 

Thank you! :smile: 

[...]

PR #705, opened 2019-08-21
Introduce an `All` target for outputting all supported module types by wasm-pack

### LINKED ISSUE #313: give option target=all which generates pkg supporting commonjs and esm

torch2424:

- [ x ] You have the latest version of `rustfmt` installed
```bash
```
- [ x ] You ran `cargo fmt` on the code base before submitting
- [ x ] You reference which issue is being closed in the PR text

relates to #697 
closes #313
 
This introduces a `--target all` flag that will run wasm-bindgen for all of our supported outputs, prefixing them with a `_[MODULE TYPE HERE]`. And then adding all of the files to the resulting package.json.

# Things still to do

* Tests (Wanted to get some approval before writing them)
* Comments made in #313 regarding `.mjs` :smile: cc @stephanemagnenat

cc@alexcrichton @ashleygwilliams

P.S I wrote the last half of this at 2AM. So the code may be terrible, but I figured we could get through this together :joy:

### DIFF
--- a/src/command/build.rs
+++ b/src/command/build.rs
@@ -226,6 +226,8 @@ impl Build {
             process_step(self)?;
         }
 
+        info!("TODO Testing!");
+
         let duration = crate::command::utils::elapsed(started.elapsed());
         info!("Done in {}.", &duration);
         info!(
@@ -373,7 +375,7 @@ impl Build {
 
         match self.target {
             Target::All => {
-            
+                info!("TODO Target all!"); 
             },
             _ => {
                 bindgen::wasm_bindgen_build(
--- a/src/command/build.rs
+++ b/src/command/build.rs
@@ -226,8 +226,6 @@ impl Build {
             process_step(self)?;
         }
 
-        PBAR.info("TODO Testing!");
-
         let duration = crate::command::utils::elapsed(started.elapsed());
         info!("Done in {}.", &duration);
         info!(
@@ -370,25 +368,15 @@ impl Build {
 
     fn step_run_wasm_bindgen(&mut self) -> Result<(), Error> {
         info!("Building the wasm bindings...");
-
-        info!("target {}", self.target);
-
-        match self.target {
-            Target::All => {
-                info!("TODO Target all!"); 
-            },
-            _ => {
-                bindgen::wasm_bindgen_build(
-                    &self.crate_data,
-                    self.bindgen.as_ref().unwrap(),
-                    &self.out_dir,
-                    &self.out_name,
-                    self.disable_dts,
-                    self.target,
-                    self.profile,
-                    )?;
-            }
-        }
+        bindgen::wasm_bindgen_build(
+            &self.crate_data,
+            self.bindgen.as_ref().unwrap(),
+            &self.out_dir,
+            &self.out_name,
+            self.disable_dts,
+            self.target,
+            self.profile,
+        )?;
         info!("wasm bindings were built at {:#?}.", &self.out_dir);
         Ok(())
     }
--- a/src/command/build.rs
+++ b/src/command/build.rs
@@ -54,6 +54,8 @@ pub enum Target {
     /// in a browser but pollutes the global namespace and must be manually
     /// instantiated.
     NoModules,
+    /// Correspond to `--target all` where the output is all other targets
+    All,
 }
 
 impl Default for Target {
@@ -69,6 +71,7 @@ impl fmt::Display for Target {
             Target::Web => "web",
             Target::Nodejs => "nodejs",
             Target::NoModules => "no-modules",
+            Target::All => "all",
         };
         write!(f, "{}", s)
     }
@@ -82,6 +85,7 @@ impl FromStr for Target {
             "web" => Ok(Target::Web),
             "nodejs" => Ok(Target::Nodejs),
             "no-modules" => Ok(Target::NoModules),
+            "all" => Ok(Target::All),
             _ => bail!("Unknown target: {}", s),
         }
     }
@@ -120,7 +124,7 @@ pub struct BuildOptions {
     pub disable_dts: bool,
 
     #[structopt(long = "target", short = "t", default_value = "bundler")]
-    /// Sets the target environment. [possible values: bundler, nodejs, web, no-modules]
+    /// Sets the target environment. [possible values: bundler, nodejs, web, no-modules, all]
     pub target: Target,
 
     #[structopt(long = "debug")]
@@ -364,15 +368,25 @@ impl Build {
 
     fn step_run_wasm_bindgen(&mut self) -> Result<(), Error> {
         info!("Building the wasm bindings...");
-        bindgen::wasm_bindgen_build(
-            &self.crate_data,
-            self.bindgen.as_ref().unwrap(),
-            &self.out_dir,
-            &self.out_name,
-            self.disable_dts,
-            self.target,
-            self.profile,
-        )?;
+
+        info!("target {}", self.target);
+
+        match self.target {
+            Target::All => {
+            
+            },
+            _ => {
+                bindgen::wasm_bindgen_build(
+                    &self.crate_data,
+                    self.bindgen.as_ref().unwrap(),
+                    &self.out_dir,
+                    &self.out_name,
+                    self.disable_dts,
+                    self.target,
+                    self.profile,
+                    )?;
+            }
+        }
         info!("wasm bindings were built at {:#?}.", &self.out_dir);
         Ok(())
     }
--- a/src/command/build.rs
+++ b/src/command/build.rs
@@ -226,7 +226,7 @@ impl Build {
             process_step(self)?;
         }
 
-        info!("TODO Testing!");
+        PBAR.info("TODO Testing!");
 
         let duration = crate::command::utils::elapsed(started.elapsed());
         info!("Done in {}.", &duration);
--- a/src/bindgen.rs
+++ b/src/bindgen.rs
@@ -55,7 +55,7 @@ pub fn wasm_bindgen_build(
                 Target::Web,
                 profile
                 )?;
-        
+
             // Nodejs
             let mut out_name_for_nodejs = Some(format!("{}.cjs.js", data.crate_name()));
             if let Some(value) = out_name {
--- a/src/bindgen.rs
+++ b/src/bindgen.rs
@@ -115,6 +115,7 @@ fn build_target_arg_legacy(target: Target, cli_path: &PathBuf) -> Result<String,
             }
         }
         Target::Bundler => "--browser",
+        Target::All => "--all",
     };
     Ok(target_arg.to_string())
 }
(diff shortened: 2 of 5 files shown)


Pauan:
I haven't reviewed this deeply, but I wonder if `--target all` is the right approach. I can't think of any use case where you would actually want to output *all* of the targets.

Instead, I think supporting multiple `--target` would be better, so you could do `wasm-pack build --target bundler --target nodejs` for example. You could still output all targets, you'd just have to list all of them. This gives a lot more flexibility.

ashleygwilliams:
@Pauan multiple people have requested an all or "isomorphic" target for publishing cross compatible packages to npm.  i do like your suggestion that perhaps the targets are configurable- strikes me as something we could have folks do in a Cargo.toml!

src/command/build.rs
```diff
@@ -54,6 +54,8 @@ pub enum Target {
     /// in a browser but pollutes the global namespace and must be manually
     /// instantiated.
     NoModules,
+    /// Correspond to `--target all` where the output is all other targets
+    All,
```
ashleygwilliams: given @Pauan's comment and the likelihood that folks will want to customize the types of targets they want... i could imagine that instead of an `All` type, you have a new type that is a `Targets: Vec<Target>`- and for just Node, the Targets would by `Vec<Nodejs>`, and all would be `Vec<Web, Nodejs`, Modules>

Pauan:
@ashleygwilliams I know, but my point is that `--target all` is the same as `--target nodejs --target bundler`, so if wasm-pack supports multiple targets then there's no need for `--target all` anymore.

ashleygwilliams:
@Pauan i think you and i are agreeing with each other! the one difference is that i think aliasing `--target nodejs --target bundler` to `--target all` is probably OK (honestly i think configuring target in `Cargo.toml` is my preferred)

this task is so common that i dont think it's reasonable to make it burdensome with an overly verbose command, and target feels like the property of a project so putting it in the project config makes a lot of sense i think

torch2424:
Sorry for the late response, was having a fun time at RustConf! :smile: 

---

@Pauan Thanks for the feedback :smile: 

--- 

@ashleygwilliams @Pauan 

So what's the plan here? I see 3 things here that I think would be worth getting an agreement on :smile: 

1. Do we want an `all` target? (Either as an alias, or a target itself).
2. Do we want this configurable in the `Cargo.toml`?
3. How do we specify which targets to build in either in the `Cargo.toml`, or as CLI flags?

In my opinion, I think my answer to each question depends on each of the answers to the above. But reading what we currently have, I think maybe it's worth proceeding like this:

**As CLI Flags**

1. Restart this PR.
2. I personally think aliasing is confusing, so I would not want an `all` target unless it had specific behavior (like this PR shows).
3. I'd prefer target to become a comma seperated of space seperated list that you pass in. E.g `--target web nodejs bundler`
4. We do the `Vec<>` implementation that @ashleygwilliams mentioned, and migrate relevant code from the PR to a new PR.

**As `Cargo.toml`**

1. Restart this PR.
2. Create a new field like `[wasm-pack targets]`?
3. Read the field when building, and a `--target` flag overrides the targets in the `Cargo.toml`.
4. We do the `Vec<>` implementation that @ashleygwilliams mentioned, and migrate relevant code from the PR to a new PR.

Thanks everyone! :smile: :+1: Let me know what you think!

rivertam:
User waiting on this feature weighing in:

- I currently build using CLI arguments that I like to change without having to change the Cargo.toml. The main reason for this is that I need to output multiple targets. However, `--target all` feels very natural to me in addition to the fact that it's already part of my workflow. I think maybe a `Cargo.toml` default which can be overridden by CLI flags? I'd start with just CLI flags because I don't think that should hold up this PR because there's bikeshedding that could be done when the feature could just be completed later
- I think `--target all` makes perfect sense for aliasing. `--target web,nodejs,bundler` is fine too.

Basically, in my ideal world, I don't think any of these ideas are at odds. I'm also pretty impatient waiting for this to land as I've been struggling with this kind of issue all week.

torch2424:
Thanks for the feedback @rivertam ! :+1: 

Going to bump this to @ashleygwilliams and @Pauan if they don't mind :smile: 

As I want to get a plan going before I write more code :joy:

Pauan:
@torch2424 I can't speak for Ashley, but this is my personal opinion:

* CLI flags are mandatory, so we will want multiple `--target` support no matter what.

* I don't like the `all` target, because it implies that it builds all the targets, but it actually only builds 2.

   And there's some complex questions like, "what targets should `all` build? what if we want to change the `all` targets later?"

   So I think the `all` target should be debated/handled in a new PR.

* I think configuration with `Cargo.toml` is okay, but that can be handled in a separate PR, I think this PR should be narrowly scoped to just multiple `--target`.

* `--target` cannot be space separated. Comma separated is okay, though passing multiple `--target` I think is a bit "cleaner".

   I think it might be reasonable to support both multiple `--target` *and* comma separated as well, so you can do either `--target foo --target bar` or `--target foo,bar`

cormacrelf:
> * `--target` cannot be space separated. Comma separated is okay, though passing multiple `--target` I think is a bit "cleaner".

`cargo` uses space-separators for features. They just need to be surrounded by quotes. Is this anomalous or the standard for cargo?

```
cargo build --features "one two three"
```

Pauan:
@cormacrelf Being surrounded by quotes is okay too, but @torch2424 mentioned space separated without any quotes.

The reason spaces don't work is because then it can't tell the difference between a `--target` argument and a regular argument. So it needs to have *something* to distinguish it (whether it be quotes, commas, etc.)

torch2424:
@cormacrelf @Pauan 

Totally okay with not being space separated. If cargo is currently using quotes, than I think using quotes and being space separated within the quotes makes sense 😄 

That being said, to just push things along for users like @rivertam , let's proceed as follows:

**CLI Flags**
1. Restart this PR.
2. Multiple targets are passed in as quotes. E.g `--target "web nodejs bundler"`
3. We do the `Vec<>` implementation that @ashleygwilliams mentioned, and migrate relevant code from the PR to a new PR.

I'll wait for a final thumbs up from @ashleygwilliams (I'll ping them on discord as well) and then start on the new PR 😄 🎉

rivertam:
Thanks @torch2424. I appreciate you.

It's worth noting I used your branch to create a similar as your branch makes:

package.json:
```json
{
  "name": "my-package",
  "collaborators": ["Ben Berman <ben@standardbots.com>"],
  "version": "0.0.0",
  "files": [
    "my-package.bundle/my-package_bg.wasm",
    "my-package.bundle/my-package.js",
    "my-package.bundle/my-package.d.ts",
    "my-package.node/my-package_bg.wasm",
    "my-package.node/my-package.js",
    "my-package.node/my-package_bg.js",
    "my-package.node/my-package.d.ts"
  ],
  "main": "my-package.node/my-package.js",
  "module": "my-package.bundler/my-package.js",
  "types": "my-package.bundler/my-package.d.ts",
  "sideEffects": "false"
}
```

then I bundle the package with

```
wasm-pack build --target bundler --out-dir my-package/my-package.bundler 
wasm-pack build --target nodejs --out-dir my-package/my-package.node
```

and I think this results in basically the same output as I want, so for *me* and presumably for anyone who reads this comment, the issue isn't too urgent anymore. This workflow is actually totally fine for me, and I'm no longer awaiting this PR because of this. That said, obviously this is a useful feature.

tlaukkan:
Please, implement this as it is really valuable for any libraries which target both platforms.

holygits:
+1 for this feature. Any help needed to progress the PR?

aminya:
Looking forward to this. 👍 It simplifies many things.

curiousdannii:
Is there any way to manually accomplish this at present? I would like to be able to make an isomorphic package which uses a wasm-pack sub-package, ideally without needing a bundling step (for Node, something will still be needed for browser). The idea would be for my isomorphic package to be able to `import` the sub-package, with the sub-package loading the appropriate version, either Node or browser.

Problem is that I don't think there's any current target that will make it output for Node using ES modules.

Any ideas? What are people currently doing to make isomorphic packages?

rivertam:
@curiousdannii Yeah, check out my comment earlier. That works for me.

holygits:
In our project we generate a nodejs and browser compatible package with wasm-pack and have a top-level package.json which is published like so: https://github.com/cennznet/doughnut-rs/blob/2af67a162f650e993779642c6879a110fad7a498/js/package.json#L5-L6

Poking around the project may give you some ideas: https://github.com/cennznet/doughnut-rs/tree/master/js

The main issue we're seeing with this approach is it generates 2 wasm blobs which blows out the package size

curiousdannii:
@rivertam The `bundler` output is not useable in Node. It tried to import the wasm file directly. (I also thought that using `require` would be a problem, but it doesn't seem to complain?)

I'm thinking that I'll have to run the `node` output through Rollup or the like.

rivertam:
@curiousdannii I'm not sure what you mean. The method I describe above includes a package for both bundler and node. I'm not sure whether the bundler works for Rollup.

curiousdannii:
@rivertam  I want package A to be able to `import` package W, loading either the Node or the Browser version as appropriate. This means for Node it would load the `module` property of package W. But if you generate that with the `bundler` you can't load it in Node because it tries to `import` the .wasm directly.

rivertam:
@curiousdannii it's hard for me to understand what the problem is because that sounds exactly like what I'm describing to me. A is sometimes used in node and sometimes with webpack, and they both work.

curiousdannii:
@rivertam Could it be that you're not running into issues because either Jest or TS is rewriting your import statements into CommonJS require calls, so that you're getting the main version instead of the module version.

rivertam:
@curiousdannii Err... I'm pretty sure Jest/TS *are* rewriting to `require` calls, which I think is what it's supposed to do in a node context. It's only once you add a bundler (which I'm not doing except in the React application) that you should want it to use the "module" version. That's the point of the setup I describe -- it has a different "main" and "module" key so different environments load them differently.

curiousdannii:
@rivertam Node has supported ES Modules natively for years (behind flags originally, though they're not required anymore), so you're not supposed to rewrite them to CommonJS. And even if you use TS, it can output ESM too. It's perfectly reasonable to ask for a tool such as wasm-pack to output ES Modules for Node, and if you want to make an isomorphic package without further build steps, it's necessary.

rivertam:
Ah, I see. I wasn't aware that node natively supported modules at this point (and it appears it's still experimental as of node 15) but I see the problem now. Does it work in node if you export with target `web`?

curiousdannii:
`web` doesn't work in Node either, though I can't remember exactly why.

Until something like this PR is merged, I'm expecting I'll just have to run the `node` output through Rollup to make an ESM version. Then at least consumers of my package won't need to use a transpiler/build step until they already were.

PR #706, opened 2019-08-22
Select correct webdriver version

### LINKED ISSUE #611: Headless Chrome test fails
## 🐛 Bug description
I'm currently following this tutorial: https://rustwasm.github.io/docs/book/game-of-life/testing.html. The default tests/web.rs file looks like this:

```
//! Test suite for the Web and headless browsers.

#![cfg(target_arch = "wasm32")]

extern crate wasm_bindgen_test;
use wasm_bindgen_test::*;

wasm_bindgen_test_configure!(run_in_browser);

#[wasm_bindgen_test]
fn pass() {
    assert_eq!(1 + 1, 2);
}
```

The tutorial suggests that I run `wasm-pack test --chrome --headless` to "confirm that the Rust-generated WebAssembly test is working". Instead it fails with the following output:

```
C:\dev\Projekte\Rust\Sandbox\wasm-game-of-life>wasm-pack test --chrome --headless
    Finished dev [unoptimized + debuginfo] target(s) in 0.42s
[INFO]: Installing wasm-bindgen...
    Finished dev [unoptimized + debuginfo] target(s) in 0.20s
     Running target\wasm32-unknown-unknown\debug\deps\wasm_game_of_life-18a0250534315977.wasm
no tests to run!
     Running target\wasm32-unknown-unknown\debug\deps\web-1bc24e2bd87521b3.wasm
Running headless tests in Chrome with `C:\Users\Tobias\AppData\Local\.wasm-pack\chromedriver-87c19a493f3a2690\chromedriver.exe`
driver status: exit code: 1
driver stdout:
    Starting ChromeDriver 2.46.628402 (536cd7adbad73a3783fdc2cab92ab2ba7ec361e1) on port 62230
    Only local connections are allowed.
    Please protect ports used by ChromeDriver and related test frameworks to prevent access by malicious code.

error: failed to find element reference in response
error: test failed, to rerun pass '--test web'
Error: Running Wasm tests with wasm-bindgen-test failed
Caused by: failed to execute `cargo test`: exited with exit code: 1
```

Running the test with `wasm-pack test --chrome` (and opening the page in chrome) is successful, as is running `wasm-pack test --firefox --headless`.

#### 🤔 Expected Behavior
The test should work and succeed for chrome headless mode.

#### 👟 Steps to reproduce
1. `cargo generate --git https://github.com/rustwasm/wasm-pack-template`
2. Project Name: wasm-game-of-life
3. `cd wasm-game-of-life`
4. `wasm-pack test --chrome --headless`

#### 🌍 Your environment
wasm-pack version: 0.7.0
rustc version: 1.33.0
System: Windows 7 64bit
Chrome: 73.0.3683.86 (official build) (64-Bit)

MartinKavik:
Closes #611

**Changes**
- `webdriver.rs` split into `chromedriver.rs`, `geckodriver.rs` and `safaridriver.rs` for better code organization.
- Functions `install_geckodriver` and `install_chromedriver` fetches info about the latest `*driver` version so they can find out if we are using the latest `*driver`.

**Testing**
- I've tested it locally on Windows 10 - it downloads both drivers without problems.
- There are new / updated `curl` calls - should I test it in unit/integration tests somehow? (Is there an example for inspiration?)

**Possible future improvements**
- We can fetch info about a new `[chrome/gecko]driver` only when testing fails because of old driver and then restart testing.
  - **+** A little bit faster testing.
  - **+** We probably don't need internet connection during testing.
  - **+** Less requests to fetched endpoints.
  - **-** It's not possible to determine exact fail reason.
  - **-** We have to remember to restart it only once to prevent infinite loop.
  - **-** More complex code.
  - **-** Probably premature optimization because those fetches aren't noticeable during testing.
- Downloaded/cached drivers are never deleted/invalidated - we can delete old driver(s) when the newer one has been downloaded.
   - Probably out of scope this issue. Maybe general problem/feature of `wasm-pack` cache / crate `binary-install`.
- We can find out versions of the installed browsers and then download compatible drivers.
  - It's not super easy and we would need platform-specific code which can be error prone.
  - See comments for functions `get_chromedriver_url` and `get_geckodriver_url` for more info.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1506,6 +1506,23 @@ dependencies = [
 ]
 
 [[package]]
+name = "serial_test"
+version = "0.2.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+dependencies = [
+ "lazy_static 1.3.0 (registry+https://github.com/rust-lang/crates.io-index)",
+]
+
+[[package]]
+name = "serial_test_derive"
+version = "0.2.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+dependencies = [
+ "quote 0.6.11 (registry+https://github.com/rust-lang/crates.io-index)",
+ "syn 0.15.29 (registry+https://github.com/rust-lang/crates.io-index)",
+]
+
+[[package]]
 name = "siphasher"
 version = "0.2.3"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -1992,6 +2009,8 @@ dependencies = [
  "serde_derive 1.0.89 (registry+https://github.com/rust-lang/crates.io-index)",
  "serde_ignored 0.0.4 (registry+https://github.com/rust-lang/crates.io-index)",
  "serde_json 1.0.39 (registry+https://github.com/rust-lang/crates.io-index)",
+ "serial_test 0.2.0 (registry+https://github.com/rust-lang/crates.io-index)",
+ "serial_test_derive 0.2.0 (registry+https://github.com/rust-lang/crates.io-index)",
  "siphasher 0.2.3 (registry+https://github.com/rust-lang/crates.io-index)",
  "strsim 0.8.0 (registry+https://github.com/rust-lang/crates.io-index)",
  "structopt 0.2.15 (registry+https://github.com/rust-lang/crates.io-index)",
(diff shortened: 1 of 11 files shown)


fitzgen:
Thanks for sending in this PR! Will take a look at this next week, after rust conf.

MartinKavik:
Changes in the latest commit (I can squash them before merging if you want):
- new file `src/stamps.rs` 
   - super simple key-value file store for saving versions + timestamps
- new file `tests/all/stamps.rs`
   - tests for `stamps.rs`
   - I've tried to write it as simple as possible but I'm waiting for review comments about best practices and file testing
   - added library `serial_test`
- updated `RELEASE_CHECKLIST` so we don't forget to update driver versions
- updated `webdrivers*`
  - see comment above function `get_chromedriver_url` or  `get_geckodriver_url` - there is a new algorithm for getting driver version (it should be similar to your comment - check once per day, "offline" mode, etc.)
- added `context`s - I'm waiting for review comments about wording and `failure` / error handling best practices

I've tested it with a local project and it seems working even without internet connection (if drivers are already downloaded).

Your comments from the previous review should be resolved.
Thanks for reviews!

PR #712, opened 2019-09-07
Correct out-dir argument behavior

### LINKED ISSUE #704: --out-dir should be relative to current directory
## 🐛 Bug description
`wasm-pack build subdir/ --out-dir pkg` dumps the output in `$PWD/subdir/pkg/`.

#### 🤔 Expected Behavior
Should output to `$PWD/pkg`, not the `subdir`. The help says: 

> -d, --out-dir <out_dir>      Sets the output directory with a relative path. [default: pkg]
> --out-name <out_name>    Sets the output file names. Defaults to package name.

My understanding is that "relative" here means relative to the current working directory.

#### 👟 Steps to reproduce
`wasm-pack build subdir/ --out-dir pkg` 

#### 🌍 Your environment
```
$ wasm-pack --version
wasm-pack 0.8.1
```

My use case here is working in a cargo workspace. I like to work in the top-level directory, and keep all of my build artifacts in the top-level `target/` directory. 

Thanks!

nalshihabi:
Fixes #704 
Correct the out-dir argument's behavior so that it behaves as documentation says it does.

```bash
```

### DIFF
--- a/src/command/build.rs
+++ b/src/command/build.rs
@@ -179,7 +179,7 @@ impl Build {
     pub fn try_from_opts(build_opts: BuildOptions) -> Result<Self, Error> {
         let crate_path = get_crate_path(build_opts.path)?;
         let crate_data = manifest::CrateData::new(&crate_path, build_opts.out_name.clone())?;
-        let out_dir = crate_path.join(PathBuf::from(build_opts.out_dir));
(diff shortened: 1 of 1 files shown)


ashleygwilliams:
if i'm understanding correctly, this would produce a breaking change. we should discuss if this is breaking and what the expected impact would be.

Issue #716, opened 2019-09-24
Warn and info log levels have different outputs


petosorus:
## 🐛 Bug description
Warn level and info levels have different behaviors. Most notably, the warn level has an emoji first, and the log level second.

![image](https://user-images.githubusercontent.com/4728156/65515813-28334500-dee0-11e9-86dc-d8b934a8b43c.png)

I took a look at the code, warn and error are similar, but info is different.
```/// Add an informational message.
    pub fn info(&self, message: &str) {
        let info = format!("{}: {}", style("[INFO]").bold().dim(), message,);
        self.message(&info);
    }

    /// Add a warning message.
    pub fn warn(&self, message: &str) {
        let warn = format!(
            "{} {}: {}",
            emoji::WARN,
            style("[WARN]").bold().dim(),
            message
        );
        self.message(&warn);
    }

    /// Add an error message.
    pub fn error(&self, message: &str) {
        let err = format!(
            "{} {}: {}",
            emoji::ERROR,
            style("[ERR]").bold().dim(),
            message
        );
        self.message(&err);
    }
```

#### 🤔 Expected Behavior
I would expect the log level to always be the first element of the log. I don't mind the emoji, I just find it in the wrong place.

#### 👟 Steps to reproduce
I ran the example from this page https://developer.mozilla.org/en-US/docs/WebAssembly/Rust_to_wasm

When running `wasm-pack build`, I see the incriminated logs.

#### 🌍 Your environment
Include the relevant details of your environment.
wasm-pack version: 0.8.1
rustc version: 1.37.0

I would be happy to take a stab at doing a PR for this :)

drager:
Makes sense! Thanks for reporting!

ashleygwilliams:
@drager you marked this as PR attached- did this get solved? let me know!

drager:
@ashleygwilliams There is a PR that hasn't been merged yet: https://github.com/rustwasm/wasm-pack/pull/723

drager:
Fixed in #723

PR #723, opened 2019-10-01
Fix logs for uniformity (#716)

petosorus:

### DIFF
--- a/src/progressbar.rs
+++ b/src/progressbar.rs
@@ -82,9 +82,9 @@ impl ProgressOutput {
     pub fn warn(&self, message: &str) {
         if !self.quiet() && self.is_log_enabled(LogLevel::Warn) {
             let warn = format!(
-                "{} {}: {}",
-                emoji::WARN,
+                "{}: {} {}",
                 style("[WARN]").bold().dim(),
+                emoji::WARN,
                 message
             );
             self.message(&warn);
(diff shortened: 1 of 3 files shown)


petosorus:
Yeah, I'm a bit dumb and made this change a bit too quickly. Thanks for the feedback.

petosorus:
Regarding the broken test, should I change `"[WARN]: \"package.metadata.wasm-...` to `"[WARN]: :-) \"package.metadata.wasm-...` as the reference string we test against, or do you think there's a way to ignore the `:-)` part as it's not that important, and maybe subject to change?

This is in `manifest.rs`, line 451. I don't know if `predicates::str::contains` would allow something like a regex similar to this `"[WARN]: (:-\))? \"package.metadata.wasm-`

drager:
> Regarding the broken test, should I change `"[WARN]: \"package.metadata.wasm-...` to `"[WARN]: :-) \"package.metadata.wasm-...` as the reference string we test against, or do you think there's a way to ignore the `:-)` part as it's not that important, and maybe subject to change?
> 
> This is in `manifest.rs`, line 451. I don't know if `predicates::str::contains` would allow something like a regex similar to this `"[WARN]: (:-\))? \"package.metadata.wasm-`

Yeah, I think you can check for `[WARN]: :)`

petosorus:
Hi! I'm still not too sure about rebase, should have done what's needed :)

drager:
@petosorus Thanks! It looks good! There's something weird going on with the checks. Once those passes we can get this merged :blush:

ashleygwilliams (approved):

PR #736, opened 2019-11-09
feat(build): add first-class support for binary crates

### LINKED ISSUE #734: support building binary crates
## 💡 Feature description
Using `wasm-pack` to build libraries is nice, but binaries are also good to have, both in the npm ecosystem and for browser usage (see also https://github.com/rustwasm/wasm-bindgen/issues/1630 and https://github.com/rust-gamedev/wg/issues/51). However, `wasm-pack` asserts in a few places that all crates being built are libraries.

#### 💻 Basic example
A crate with the following code in `src/main.rs` should build and run as intended with `wasm-pack build --target web`:
```rust
#[cfg(target_arch = "wasm32")]
use wasm_bindgen::prelude::*;

#[cfg(target_arch = "wasm32")]
fn display(text: &str) {
    // Use `web_sys`'s global `window` function to get a handle on the global
    // window object.
    let window = web_sys::window().expect("no global `window` exists");
    let document = window.document().expect("should have a document on window");
    let body = document.body().expect("document should have a body");

    // Manufacture the element we're gonna append
    let val = document.create_element("p").expect("failed to create a <p>");
    val.set_inner_html(text);

    body.append_child(&val).expect("failed to append child to body");
}

#[cfg(not(target_arch = "wasm32"))]
fn display(text: &str) {
    println!("{}", text);
}

fn main() {
    display("Hello from Rust!");
}

// i'm also working on wasm-bindgen changes to make this less painful
#[cfg(target_arch = "wasm32")]
#[wasm_bindgen(start)]
pub fn wasm_main() {
    main();
}
```

(I've already implemented this, and was about to submit a pull request when I actually got around to reading the contributing guidelines. I'll send in the pull request once someone's indicated that this is worth implementing.)

boringcactus:

- [X] You have the latest version of `rustfmt` installed
```bash
```
- [X] You ran `cargo fmt` on the code base before submitting
- [X] You reference which issue is being closed in the PR text

---

fixes #734 

Now that https://github.com/rustwasm/wasm-bindgen/issues/1630 has been implemented, I think it's good to have the same support in `wasm-pack`.

This implementation currently doesn't add `bin` entries in `package.json`, partially because [the docs on that field](https://docs.npmjs.com/files/package.json.html#bin) say that targets must be prefixed with `#!/usr/bin/env node` and I'm not sure how or where to ensure that all and only binaries get that prefix. I don't think that feature is vital, but if other people think it's a must-have then I can take a look at adding it.

### DIFF
--- a/src/bindgen.rs
+++ b/src/bindgen.rs
@@ -28,49 +28,52 @@ pub fn wasm_bindgen_build(
 
     let out_dir = out_dir.to_str().unwrap();
 
-    let wasm_path = data
-        .target_directory()
-        .join("wasm32-unknown-unknown")
-        .join(release_or_debug)
-        .join(data.crate_name())
-        .with_extension("wasm");
+    for target_name in data.targets() {
+        let wasm_path = data
+            .target_directory()
+            .join("wasm32-unknown-unknown")
+            .join(release_or_debug)
+            .join(target_name)
+            .with_extension("wasm");
 
-    let dts_arg = if disable_dts {
-        "--no-typescript"
-    } else {
-        "--typescript"
-    };
-    let bindgen_path = bindgen.binary("wasm-bindgen")?;
+        let dts_arg = if disable_dts {
+            "--no-typescript"
+        } else {
+            "--typescript"
+        };
+        let bindgen_path = bindgen.binary("wasm-bindgen")?;
 
-    let mut cmd = Command::new(&bindgen_path);
-    cmd.arg(&wasm_path)
-        .arg("--out-dir")
-        .arg(out_dir)
-        .arg(dts_arg);
+        let mut cmd = Command::new(&bindgen_path);
+        cmd.arg(&wasm_path)
+            .arg("--out-dir")
+            .arg(out_dir)
+            .arg(dts_arg);
 
-    let target_arg = build_target_arg(target, &bindgen_path)?;
-    if supports_dash_dash_target(&bindgen_path)? {
-        cmd.arg("--target").arg(target_arg);
-    } else {
-        cmd.arg(target_arg);
+        let target_arg = build_target_arg(target, &bindgen_path)?;
+        if supports_dash_dash_target(&bindgen_path)? {
+            cmd.arg("--target").arg(target_arg);
+        } else {
+            cmd.arg(target_arg);
+        }
+
+        if let Some(value) = out_name {
+            cmd.arg("--out-name").arg(value);
+        }
+
+        let profile = data.configured_profile(profile);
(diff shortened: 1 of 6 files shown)


boringcactus:
currently the way to get examples built is to throw `--bins --examples` straight to `cargo build`. should we accept a `--example foo` argument instead? that's probably cleaner but also a little bit more work on the implementation side

boringcactus:
`--example foo` now works.

boringcactus:
issue with the current code: if a binary is defined but required features aren't available, it doesn't get built, but this code still tries to run wasm-bindgen against it, causing problems. i think i can fix that eventually, but i want to make a note of it here in case i forget.

src/bindgen.rs
```diff
@@ -28,49 +28,62 @@ pub fn wasm_bindgen_build(
...
-        .join(data.crate_name())
-        .with_extension("wasm");
-
-    let dts_arg = if disable_dts {
-        "--no-typescript"
-    } else {
-        "--typescript"
-    };
-    let bindgen_path = bindgen.binary("wasm-bindgen")?;
+    for target_name in data.targets() {
```
daxpedda: I think this PR could be a bit more simplified if we remove handling multiple crates here and leave that to #732.

src/manifest/mod.rs
```diff
@@ -524,6 +542,22 @@ impl CrateData {
         }
     }
 
+    /// Get the target names that will be built for the current crate.
+    pub fn targets(&self) -> impl Iterator<Item = String> + '_ {
```
daxpedda: As mentioned above, parts of this can be removed to make this PR simpler.

src/bindgen.rs
boringcactus: I think this is distinct from #732; if a crate has multiple binaries, a single `cargo build` will create multiple `.wasm` files for it, and the same invocation of `wasm-pack` should process all of them. #732 appears to be intended to allow repeated invocations of `wasm-pack` to build on top of each other; the example given uses two different feature sets, which is orthogonal to what this is doing.

daxpedda: > #732 appears to be intended to allow repeated invocations of `wasm-pack` to build on top of each other

I didn't realize that 😮! My intention was to slim down the PR to help push this over the finishing line, because I am assuming that the bandwidth of the wasm team is pretty strained.
It would be best if somebody from the wasm team says what he thinks about this of course.

alvinhochun:
Other than a rebase (which #816 tried to do), is there anything else that is blocking the merge of this PR?

PR #738, opened 2019-11-11
Create npm installer

### LINKED ISSUE #321: Create a `wasm-pack` npm module that installs wasm-pack
## 💡 Feature description

Publish a `wasm-pack` package to npm that runs the new wasm-pack installers, so JS users can get wasm-pack setup with just `npm i -g wasm-pack`.

#### 💻 Basic example

```
$ npm i -g wasm-pack
...
$ wasm-pack --version # and now wasm-pack is installed!
0.5.0
```

EverlastingBugstopper:
This PR creates an npm installer for `wasm-pack` that downloads and installs a binary. It uses this unpublished package for installing binaries https://github.com/everlastingbugstopper/binary-install. 

It currently downloads from a workers.cloudflare endpoint that we'd likely want to change to something that the rustwasm working group has control over.

We also probably want some more keywords before this would be published - currently I've just copied the same tags from the repository.

Proof that it works
```console
$ cd ./wasm-pack/npm
$ npm i -g

> wasm-pack@0.8.1 postinstall 
> node -e 'require("./binary.js").install()'

Downloading release https://workers.cloudflare.com/get-binary/rustwasm/wasm-pack/v0.8.1/x86_64-apple-darwin.tar.gz
wasm-pack has been installed!
+ wasm-pack@0.8.1
updated 1 package in 1.317s                                                                                                                                                        
$ which wasm-pack
~/.nvm/versions/node/v13.1.0/bin/wasm-pack                                                                                                                                                         
$ wasm-pack
wasm-pack 0.8.1
Ashley Williams <ashley666ashley@gmail.com>
📦 ✨  pack and publish your wasm!

USAGE:
    wasm-pack [FLAGS] <SUBCOMMAND>

FLAGS:
    -h, --help       Prints help information
    -V, --version    Prints version information
    -v, --verbose    Log verbosity is based off the number of v used

SUBCOMMANDS:
    build      🏗️  build your npm package!
    help       Prints this message or the help of the given subcommand(s)
    login      👤  Add an npm registry user account! (aliases: adduser, add-user)
    pack       🍱  create a tar of your npm package but don't publish!
    publish    🎆  pack up your npm package and publish!
    test       👩‍🔬  test your wasm!
```

Uninstalling will remove the downloaded binary, and installing a new version will overwrite the old downloaded binary.

Some caveats:

The README will need to be updated on every wasm-pack release, and the version number in package.json will also need to be bumped on every release.

Fixes #321

### DIFF
--- a/npm/.gitignore
+++ b/npm/.gitignore
@@ -1,0 +1,1 @@
+node_modules
--- a/npm/README.md
+++ b/npm/README.md
@@ -1,0 +1,94 @@
+<div align="center">
+
+  <h1>📦✨  wasm-pack</h1>
+
+  <p>
+    <strong>Your favorite Rust → Wasm workflow tool!</strong>
+  </p>
+
+  <p>
+    <a href="https://travis-ci.com/rustwasm/wasm-pack"><img alt="Build Status" src="https://travis-ci.com/rustwasm/wasm-pack.svg?branch=master"/></a>
+    <a href="https://ci.appveyor.com/project/ashleygwilliams/wasm-pack-071k0"><img alt="Build status" src="https://ci.appveyor.com/api/projects/status/iv1qtnqtv168ef8h?svg=true"/></a>
+    <a href="https://crates.io/crates/wasm-pack"><img alt="crates.io" src="https://meritbadge.herokuapp.com/wasm-pack"/></a>
+  </p>
+
+  <h3>
+    <a href="https://rustwasm.github.io/docs/wasm-pack/">Docs</a>
+    <span> | </span>
+    <a href="https://github.com/rustwasm/wasm-pack/blob/master/CONTRIBUTING.md">Contributing</a>
+    <span> | </span>
+    <a href="https://discordapp.com/channels/442252698964721669/443151097398296587">Chat</a>
+  </h3>
+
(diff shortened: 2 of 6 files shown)


Pauan:
Thanks a lot for starting on this!

I wonder if we could just create an npm package which contains the pre-built binaries (and a simple script as a fallback), and then just publish that.

That avoids needing to manage our own hosting, and it also means that versioning will work correctly (including version locking).

The alternative would be to point to the release files [already hosted on GitHub](https://github.com/rustwasm/wasm-pack/releases), this strategy is used by [PureScript](https://github.com/purescript/purescript/tree/master/npm-package).

EverlastingBugstopper:
Hosting is definitely something to consider.

> The alternative would be to point to the release files already hosted on GitHub, this strategy is used by PureScript.

That's what the `workers.cloudflare` link is doing except that it also caches downloads so we don't run into Github rate limits.

Pauan:
My understanding is that [GitHub has no rate limits on releases](https://help.github.com/en/github/managing-large-files/distributing-large-binaries). I assume that they also heavily cache it, because the releases are immutable and cannot change.

EverlastingBugstopper:
Well in that case let's just use the Github releases! I've updated the PR to pull directly from Github.

PR #760, opened 2019-12-16
[WIP] 🎅 0.9.0

EverlastingBugstopper:
Fixes #646

Issue #696, opened 2019-07-30
wasm-opt not working on latest master


torch2424:
relates to #625 

## 🐛 Bug description

The installed `wasm-opt` with `wasm-pack` on the [current master](https://github.com/rustwasm/wasm-pack/commit/439e523192676fb481b14a79ea88d03a953164e0) isn't working an exits with signal 6:

![Screenshot from 2019-07-29 07-36-12](https://user-images.githubusercontent.com/1448289/62056996-9525a980-b1d3-11e9-9b52-f5d6a804f21f.png)

#### 🤔 Expected Behavior

`wasm-opt` should run, optimize the wasm file, and not error/stop building the package with `wasm-pack`

#### 👟 Steps to reproduce

1. Clone the repo
2. `cargo run build [path to your rust wasm module]`

#### 🌍 Your environment
Include the relevant details of your environment.
wasm-pack version: [current master](https://github.com/rustwasm/wasm-pack/commit/439e523192676fb481b14a79ea88d03a953164e0)
rustc version: 1.34.1

cc @alexcrichton :smile:

alexcrichton:
What OS are you running on? (e.g. Ubuntu/Debian/etc)

Additionally can you upload the wasm file which causes this segfault?

torch2424:
@alexcrichton 

> What OS are you running on? (e.g. Ubuntu/Debian/etc)

Oops totally forgot that :joy: 

![Screenshot from 2019-07-31 01-11-56](https://user-images.githubusercontent.com/1448289/62195149-39246780-b330-11e9-9ca3-5fcf6a8af836.png)

pop_OS! is a derivative of Ubuntu :smile: 

> Additionally can you upload the wasm file which causes this segfault?

So here is the terminal output:

![Screenshot from 2019-07-31 01-14-41](https://user-images.githubusercontent.com/1448289/62195361-b18b2880-b330-11e9-9e22-4061f4f8e569.png)

So from that, I assume I need to get the `pkg/wasmboypluginhqx_bg.wasm`. Which you can [download here](https://files.aaronthedev.com/$/qhvs3).

Let me know if that helps! :smile:

alexcrichton:
Ok using the `release_78` binaries locally myself I can get `-O3` to segfault when running over that wasm file. Additionally using the latest `release_87` it segfaults. I've opened https://github.com/WebAssembly/binaryen/issues/2273 as an upstream issue.

torch2424:
[...]

PR #766, opened 2020-01-10
Add docs and actionable error message for failed wasm-opt executions

EverlastingBugstopper:
In light of our [upcoming release](#760) and the [reported bug](https://github.com/rustwasm/wasm-pack/issues/696) regarding `wasm-opt` on Linux, this PR updates the docs and includes a helpful error message on failed `wasm-opt` executions.

Output looks like this:

```console
$ wasm-pack build --release
[INFO]: 🎯  Checking for the Wasm target...
[INFO]: 🌀  Compiling to Wasm...
    Finished release [optimized] target(s) in 0.02s
[INFO]: Optimizing wasm binaries with `wasm-opt`...
Error: failed to execute `wasm-opt`: exited with exit code: 1
  full command: "/Users/averyharnish/Library/Caches/.wasm-pack/wasm-opt-86de3fed0e357d00/wasm-opt" "/Users/averyharnish/Documents/work/_tmp/test-wasm/pkg/test_wasm_bg.wasm" "-X" "/Users/averyharnish/Documents/work/_tmp/test-wasm/pkg/test_wasm_bg.wasm-opt.wasm" "-O"
To disable `wasm-opt`, add `wasm-opt = false` to your package metadata in your `Cargo.toml`.
```

### DIFF
--- a/docs/src/cargo-toml-configuration.md
+++ b/docs/src/cargo-toml-configuration.md
@@ -14,11 +14,11 @@ The available configuration options and their default values are shown below:
 # the Rust compiler has finished? Using `wasm-opt` can often further decrease
 # binary size or do clever tricks that haven't made their way into LLVM yet.
 #
-# Configuration can be set to `false` if you want to disable it (as is the
-# default for the dev profile), or it can be an array of strings which are
-# explicit arguments to pass to `wasm-opt`. For example `['-Os']` would optimize
-# for size while `['-O4']` would execute very expensive optimizations passes
-wasm-opt = false
+# Configuration is set to `false` by default for the dev profile, but it can
+# be set to an array of strings which are explicit arguments to pass to
+# `wasm-opt`. For example `['-Os']` would optimize for size while `['-O4']`
+# would execute very expensive optimizations passes
+wasm-opt = ['-O']
 
 [package.metadata.wasm-pack.profile.dev.wasm-bindgen]
 # Should we enable wasm-bindgen's debug assertions in its generated JS glue?
@@ -36,8 +36,10 @@ debug-js-glue = false
 demangle-name-section = true
 dwarf-debug-info = false
 
+# `wasm-opt` is on by default in for the release profile, but it can be
+# disabled by setting it to `false`
 [package.metadata.wasm-pack.profile.release]
-wasm-opt = ['-O']
+wasm-opt = false
 
 [package.metadata.wasm-pack.profile.release.wasm-bindgen]
 debug-js-glue = false
--- a/src/command/build.rs
+++ b/src/command/build.rs
@@ -392,7 +392,10 @@ impl Build {
             &self.out_dir,
             &args,
             self.mode.install_permitted(),
-        )?;
-        Ok(())
+        ).map_err(|e| {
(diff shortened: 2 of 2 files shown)


EverlastingBugstopper:
@torch2424 if you have time could you see if this branch lets you build w/o changes to your config?

torch2424:
Yep! It works! This is half for my future reference and for you but:

I went ahead and re-enabled wasm-opt in the `Cargo.toml` of the project:

Screenshot from 2020-01-11 00-48-31

Tried to build with the latest wasm-pack master, and it fails:

Screenshot from 2020-01-11 00-48-18

Then I cloned down your fork (I think it's your fork? Let me know if it isn't), switched to your branch, and I just got the warning and I was able to build :smile: 

Screenshot from 2020-01-11 00-51-51

EverlastingBugstopper:
Awesome! Thanks for checking that out for me :)

tests/all/wasm_opt.rs
```diff
@@ -147,7 +147,7 @@ fn custom_args() {
         .arg("build")
         .assert()
         .stderr(predicates::str::contains("--not-accepted-argument"))
-        .failure();
```
EverlastingBugstopper: this pr changes the expected behavior so i changed this failure to a success. i've talked with @ashleygwilliams about this change and we think it makes sense to build and warn if `wasm-opt` fails since it's not a necessary build step. That being said, I could see how you might want to make it fail instead of warn if it was custom arguments that failed and not the default `wasm-opt` enablement that failed.

cc @alexcrichton - this is related to https://github.com/rustwasm/wasm-pack/pull/625

ashleygwilliams: looks good to me

EverlastingBugstopper:
I could see this being a problem for people who have `wasm-pack` in their CI and `wasm-opt` enabled in their `Cargo.toml` - since this no longer fails it is easier to ignore if `wasm-opt` fails which could lead to people accidentally publishing a release that did not run `wasm-opt` without giving them time to debug _why_ `wasm-opt` failed.

ashleygwilliams:
@Pauan @fitzgen @alexcrichton do you have thoughts on this being a breaking change? and/or what behavior you'd want to see?

Pauan:
I'm not really comfortable with this change. Right now, wasm-opt is only run in release builds, and the point of release builds is to be as optimized as possible.

So if wasm-opt fails, then the file size and performance will be degraded. That's something important to notice and fix. And warnings can get lost easily and overlooked, leading to accidental performance regressions.

My solution to this problem is: if there's a bug causing wasm-opt to error, I'd rather fix that bug. And if users don't want to run wasm-opt at all, I'm fine with providing a flag or option to toggle that.

But I think the default should be to run wasm-opt (in release mode), and hard error if something goes wrong.

EverlastingBugstopper:
I've updated this PR to include docs changes and a (hopefully) more helpful error message. Happy to change the wording of either

ashleygwilliams:
cc @Pauan - ultimately i think you are right- do you think the error message and docs help?

ashleygwilliams:
mergin

PR #767, opened 2020-01-20
doc: Document installation via npm or yarn

### LINKED ISSUE #751: Document npm installer
Now that `wasm-pack` can be installed via `npm`, we should document that installation method. 

`npm install -g wasm-pack`

drager:
Closes #751.

Looks like this:

Screenshot from 2020-01-20 20-41-29

### DIFF
--- a/docs/_installer/index.html
+++ b/docs/_installer/index.html
@@ -116,6 +116,19 @@
           </p>
           <code>cargo install wasm-pack</code>
         </div>
+        <hr/>
+        <div>
+          <p>
+          Install using <b>npm</b> or <b>yarn</b>:
+          </p>
+          <code>
+            npm install -g wasm-pack
+          </code>
+          or
+          <code>
+            yarn global add wasm-pack
+          </code>
+        </div>
       </div>
     </div>
   </section>


docs/_installer/index.html
```diff
@@ -116,6 +116,19 @@ <h1>Install <code>wasm-pack</code></h1>
           </p>
           <code>cargo install wasm-pack</code>
         </div>
+        <hr/>
+        <div>
+          <p>
+          Install using <b>npm</b> or <b>yarn</b>:
+          </p>
+          <code>
+            npm install -g wasm-pack
```
Pauan: I don't think these should be installed globally. To ensure that builds are reproducible, they should instead be local, like this:

```
npm install --save-dev wasm-pack
yarn add --dev wasm-pack
```

ashleygwilliams: in general i would agree with you but this release bakes cargo-generate into wasm-pack.. so now it can also generate projects, which i think means it should be globally installed

Pauan: Shouldn't the generation be changed to install `wasm-pack` locally, then? There are a lot of benefits to local packages, and some definite long term pain with global packages.

Pauan: (We don't necessarily need to change it in *this* release, we can change the generation behavior in the next release.)

ashleygwilliams: generation means, cargo-generate, and uses the flow espouse by our docs, replacing the step of generating a project via cargo generate with wasm-pack new. this was proposed and approved and ships as part of this release. we are not going to change it because that would be a breaking change. 

i think we can advise people to install it globally. folks who have a project where it is better suited to install as a dev dep likely are aware of that. i tend to prefer to give instructions to folks who are more beginner, than folks with more advanced needs.

drager: @Pauan Yeah, I agree. I don't really recommend global installs myself. I always add as devDependencies. Maybe we can add a text under: "or install it was a local dev dependency with `npm install --save-dev wasm-pack` or `yarn add --dev wasm-pack`"

Pauan: @ashleygwilliams Changing the behavior is a non-breaking change, since only new projects would get the local install, old projects would keep using the global install, so nothing breaks.

It is precisely *because* of beginners that I recommend local installs, because local installs are safe, they do not cause problems, so they are better suited for beginners.

Global installs are a footgun that cause problems, which is very off-putting for beginners, especially because they don't understand *why* things are breaking.

We shouldn't be encouraging beginners (who don't know better) to do the wrong thing, which will cause them problems and also cause them to develop bad habits.

Not only is that harmful to them, but it's also harmful to the reputation of Rust + Wasm, because it makes us look incompetent, which means we can't be trusted to give good advice.

People used to install installs things like `webpack` and `typescript` and `eslint` globally, but it caused problems so now everybody installs them locally. Local installs are the de-facto norm in the npm ecosystem for good reason, we should follow those norms.

Pauan: Also, if the concern is about the docs, I can update the docs to include a paragraph like this:

> If you installed wasm-pack with npm or yarn, you can run it by using `npx wasm-pack ...` or `yarn wasm-pack ...` respectively.

That way users don't get tripped up when `wasm-pack` doesn't work.

ashleygwilliams: installing the project as a dev dep doesn't enable folks to use `wasm-pack new` to generate new projects. how do you account for that.

ashleygwilliams: i am ok recommending npx, but the dev dep suggestion doesn't fully work- as it only account for a subset of the potential usecases.

ashleygwilliams: ok here's my proposal:

```
Using npm:

$ npx wasm-pack new my-project to generate a new project, and
$ npm i wasm-pack --save-dev to build and publish
```

... yarn doesn't seem to have an equivalent that works (dlx doesn't work on wasm-pack  today)

would this work for both of you, @Pauan and @drager ?

Pauan: Using the dev dependency does work for all use cases. Here's an example: you can use `npx @angular/cli new foo` to create a new Angular app. This will do the following:

1. It creates a new folder `foo`

2. It puts all the necessary Angular glue into `foo`, including a `package.json` which contains `@angular/cli` as a `devDependencies`

3. Then it automatically runs `npm install`

In other words, `npx @angular/cli new` creates a new project which contains `@angular/cli` in `devDependencies`. So you're using the global package to create a new project which contains the local package.

This is the normal way of doing things, [`create-react-app` does the same thing](https://create-react-app.dev/docs/getting-started/#quick-start). They even explicitly recommend *not* installing `create-react-app` globally.

This has some advantages:

* When creating a new project, `npx` will always automatically use the most up to date version, so the user no longer needs to manage the version of wasm-pack.

* It installs wasm-pack locally, thus avoiding all of the issues with global installs.

* The user doesn't need to worry about the difference between global and local, they can just use `npx wasm-pack` and everything will work, both globally and locally.

As for yarn, the equivalent would be [`yarn create wasm-pack new`](https://legacy.yarnpkg.com/en/docs/cli/create) and `yarn wasm-pack`. Unfortunately they don't use the same command for both, but we can just explain that in the docs.

ashleygwilliams: i gave a proposal that sounds like  what you are suggesting, can you make an alternate proposal or comment on mine, i cannot actually tell what you are suggesting.

Pauan: The difference is that with your proposal the user has to manually install `wasm-pack` locally, but with my proposal `wasm-pack new` will automatically install wasm-pack locally.

ashleygwilliams:
the command doesnt do that right now. i am trying to get the release out. can we do that for next release and go with my proposal which matches the current behavior?

if u would like to change the behavior you can make a pr that also updates this documentarion

Pauan:
@ashleygwilliams Sure, that's why I was saying we can change the behavior in the next release, since it's a non-breaking change.

PR #774, opened 2020-01-25
[WIP] fix(test): drop geckodriver to v0.23.0 due windows regression > v0.24.0

### LINKED ISSUE #770: firefox tests are failing on windows
doesn't appear to line up with a change in wasm-pack but i'm genuinely not sure. i feel like this only started very recently, past few days at best- thoughts @EverlastingBugstopper ?

ashleygwilliams:
fixes #770 

waiting on some more input before merging


src/test/webdriver/geckodriver.rs
```diff
@@ -20,8 +21,18 @@ pub fn get_or_install_geckodriver(
...
-    if let Ok(path) = which::which("geckodriver") {
-        return Ok(path);
+    // geckodriver Windows binaries >v0.24.0 have an additional
+    // runtime dependency that we cannot be sure is present on the
+    // user's machine
+    //
+    // https://github.com/mozilla/geckodriver/issues/1617
+    //
+    // until this is resolved, always install v0.24.0 on windows
+    if !target::WINDOWS {
```
Pauan: Why don't we use the binary if it already exists?

Pauan: Ah, okay, I would have thought the user would have had lower than 0.26.0, but if it's failing on CI then I agree this is the right approach.

PR #785, opened 2020-02-07
0.9.1

ashleygwilliams:


CHANGELOG.md
```diff
@@ -1,5 +1,33 @@
...
+
+- ### 🤕 Fixes
+
+  - **Bump wasm-opt to version_90 - [ashleygwilliams], [issue/781] [issue/782] [pull/687]**
+
+    Previously, wasm-pack was hardcoded to install and attempt to execute wasm-opt on every build
+    using wasm-opt version 78. This version had various issues on Unix/Linux and caused broken CI
+    builds for many folks (we're so sorry!).
+
+    This PR updates the wasm-opt version to 90, which should fix the issues folks were having.
```
EverlastingBugstopper: nit: is it `wasm-opt` or `binaryen` that we version bump? or both?

ashleygwilliams: it's binaryen, though we only use it for wasm-opt, at least at the moment. do you think it's worth updating? i'm not sure, but if you think so i'm happy to

ashleygwilliams: i'll update it :)

PR #792, opened 2020-02-11
feat: npm install wasm-pack to local/global node_modules bin

### LINKED ISSUE #791: wasm-pack npm package should install wasm-pack binary to local node_modules
## 💡 Feature description
Currently, the npm package for wasm-pack does not provide a specific installation directory, which causes `binary-install` to fallback to using either a specific os env var or the users home dir. 

My proposal is to always install wasm-pack to $my_proj/node_modules/wasm-pack/bin/ (or $global/node_modules/wasm-pack/bin for npm install -g wasm-pack).

This strategy has a few advantages:

1) It follows a common practice among dev tools such as: tsc (typescript compiler), uglifyjs, node-sass, etc.
2) If you're working on multiple projects, then you're guaranteed to keep different versions distinct i.e. if proj 1 is using wasm-pack@0.1.0 and proj 2 is using wasm-pack@1.0.0, there is no risk of proj 2 using the wasm-pack@0.1.0 binary
3) CI pipeline caching becomes much simpler. For example, currently in Gitlab CI you must set the XDG_CONFIG_HOME env variable in order to make sure the wasm-pack binary installs to a cacheable location.

#### 💻 Basic example
Essentially, all that has to be added is something along these lines:

```javascript
// wasm-pack/npm/binary.js

const getBinary = () => {
  const platform = getPlatform();
  const version = require("./package.json").version;
  const author = "rustwasm";
  const name = "wasm-pack";
  const url = `https://github.com/${author}/${name}/releases/download/v${version}/${name}-v${version}-${platform}.tar.gz`;
  return new Binary(
    url,
    {
      name,
      installDirectory: "./node_modules/wasm-pack/bin" // or something along these lines
    }
  );
}
```

Zaba505:
Closes #791 

See related issue for reasoning and discussion about this change.

### DIFF
--- a/npm/binary.js
+++ b/npm/binary.js
@@ -27,7 +27,7 @@ const getBinary = () => {
   const url = `https://github.com/${author}/${name}/releases/download/v${version}/${name}-v${version}-${platform}.tar.gz`;
   return new Binary(url, {
     name,
-    installDirectory: join(__dirname, "..", "bin")
+    installDirectory: join(__dirname, "bin")
   });
 };
 
--- a/npm/binary.js
+++ b/npm/binary.js
@@ -25,7 +25,10 @@ const getBinary = () => {
   const author = "rustwasm";
   const name = "wasm-pack";
   const url = `https://github.com/${author}/${name}/releases/download/v${version}/${name}-v${version}-${platform}.tar.gz`;
-  return new Binary(url, { name });
+  return new Binary(url, {
+    name,
+    installDirectory: join(__dirname, "..", "bin")
+  });
 };
 
 const run = () => {


Zaba505:
@EverlastingBugstopper this should work for either `npm install -g` or `npm install`.

I took this from node-sass's install script, where there default binary install path is `join(__dirname, '..', 'vendor')`. The `'..'` is there because the script is running in a subfolder i.e. node-sass/scripts/install.js vs node-sass/install.js

EverlastingBugstopper:
Hey @Zaba505 - sorry for letting this sit so long. I think this change probably belongs upstream in https://github.com/everlastingbugstopper/binary-install. Do you think you can re-submit there so the default behavior is changed and I'll cut a point release?

Zaba505:
@EverlastingBugstopper I'm closing this PR since this has been resolved upstream by [binary-install#5](https://github.com/EverlastingBugstopper/binary-install/pull/5).

PR #799, opened 2020-02-21
Add --omit-imports flag

darinmorrison:
This PR adds an `--omit-imports` flag corresponding to the `wasm-bindgen-cli` flag added upstream in https://github.com/rustwasm/wasm-bindgen/pull/1958.

This flag is useful for some scenarios where imports must be controlled through separate preloads scripts where items are brought into the application scope by modifying `global` or `window` directly, instead of through the usual `import` or `require` mechanisms.

Examples of using this flag can be found in the `omit-imports` branch of [electron-sys](https://github.com/interfaces-rs/electron-sys/tree/omit-imports). See the [activity-monitor](https://github.com/interfaces-rs/electron-sys/tree/omit-imports/examples/activity-monitor) example (with [package.json](https://github.com/interfaces-rs/electron-sys/blob/10f467bae189a35c15e9206d42dfa952e697fb35/examples/activity-monitor/package.json#L9) and [preload.js](https://github.com/interfaces-rs/electron-sys/blob/omit-imports/examples/activity-monitor/preload.js)) and [hash](https://github.com/interfaces-rs/electron-sys/tree/omit-imports/examples/hash) example with (with [package.json](https://github.com/interfaces-rs/electron-sys/blob/10f467bae189a35c15e9206d42dfa952e697fb35/examples/hash/package.json#L9) and [preload.js](https://github.com/interfaces-rs/electron-sys/blob/omit-imports/examples/hash/preload.js)).

### DIFF
--- a/src/bindgen.rs
+++ b/src/bindgen.rs
@@ -17,6 +17,7 @@ pub fn wasm_bindgen_build(
     out_dir: &Path,
     out_name: &Option<String>,
     disable_dts: bool,
+    omit_imports: bool,
     target: Target,
     profile: BuildProfile,
 ) -> Result<(), failure::Error> {
@@ -34,19 +35,21 @@ pub fn wasm_bindgen_build(
         .join(data.crate_name())
         .with_extension("wasm");
 
-    let dts_arg = if disable_dts {
-        "--no-typescript"
-    } else {
-        "--typescript"
-    };
     let bindgen_path = install::get_tool_path(install_status, Tool::WasmBindgen)?
         .binary(&Tool::WasmBindgen.to_string())?;
 
     let mut cmd = Command::new(&bindgen_path);
-    cmd.arg(&wasm_path)
-        .arg("--out-dir")
-        .arg(out_dir)
-        .arg(dts_arg);
+    cmd.arg(&wasm_path).arg("--out-dir").arg(out_dir);
+
+    if disable_dts {
+        cmd.arg("--no-typescript");
+    } else {
+        cmd.arg("--typescript");
+    };
+
+    if omit_imports {
+        cmd.arg("--omit-imports");
+    }
 
     let target_arg = build_target_arg(target, &bindgen_path)?;
     if supports_dash_dash_target(bindgen_path.to_path_buf())? {
--- a/src/command/build.rs
+++ b/src/command/build.rs
@@ -27,6 +27,7 @@ pub struct Build {
     pub crate_data: manifest::CrateData,
     pub scope: Option<String>,
     pub disable_dts: bool,
+    pub omit_imports: bool,
     pub target: Target,
     pub profile: BuildProfile,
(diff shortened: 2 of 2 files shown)


Pauan:
Hmmm, I'm not sure if this is a good idea, we don't add *all* of the wasm-bindgen flags, especially the more obscure ones.

It's always possible to use the wasm-bindgen flags by using `--`, like this:

```
wasm-pack build -- --omit-imports
```

darinmorrison:
@Pauan are you sure that actually works? When I try modifying the npm script in the first example from

```
"pack-app-renderer": "cd crates/app-renderer && wasm-pack build --omit-imports --target web"
```

to

```
"pack-app-renderer": "cd crates/app-renderer && wasm-pack build --target web -- --omit-imports",
```

it does not work, but errors with the following:

```
> cd crates/app-renderer && wasm-pack build --target web -- --omit-imports

[INFO]: Compiling to Wasm...
error: Found argument '--omit-imports' which wasn't expected, or isn't valid in this context
```

From `wasm-pack build --help`, it sounds like `-- <extra_options>` get passed to `cargo build` but we need them to be passed to `wasm-bindgen`.

Am I missing something?

Pauan:
@darinmorrison You're right, I guess I got confused. We really should have a way to send arbitrary arguments to wasm-bindgen.

The only reason I see to send arguments to `cargo build` is for `--features`. So I think this would be better:

* Add in `--features`, `--all-features`, and `--no-default-features` to wasm-pack.

* Change `--` to pass the extra arguments to wasm-bindgen rather than cargo.

However, this is a breaking change. @ashleygwilliams what do you think?

darinmorrison:
Thanks for the clarification. A couple thoughts:

First, I agree exposing `--omit-imports` isn't great, since it is a more obscure flag. I was hesitant to add this but I didn't see an alternative.

Second, maybe it is worth considering another approach that would accomplish something similar to what you are describing but without being a breaking change.

We could add two new flags: `--cargo-options` and `--bindgen-options`. These would take something like a quoted argument string or comma separated arguments and pass them on to the respective tool. Then, `-- <extra_options>` could be kept as-is (for now anyway), perhaps with a deprecation warning.

I don't have a strong opinion about doing it this way versus what you proposed. I could imagine it might still be useful to be able to pass arguments to `cargo build` for some things, like experimenting with new features or debugging, but I don't have any real evidence for that.

But doing it this way with explicit `*-options` flags might be more flexible and would also generalize in case new tools are added to the pipeline in the future.

Pauan:
> We could add two new flags: `--cargo-options` and `--bindgen-options`.

I think that's reasonable. Though `--wasm-bindgen-options` is a bit of a mouthful. But it's only used for obscure options, so I think it's okay. Maybe it could be shortened to `--wasm-bindgen`?

darinmorrison:
Closing in favor of #805.

PR #827, opened 2020-04-06
Export the Binary from npm module

qm3ster:
This is not done as an optimization, but to intentionally leak other things about `Binary`.
For example, it enables me to do a cheeky
```js
child_process.spawn(require('wasm-pack')._getBinaryPath(), args)
```
in a consuming module, without spawning an extra node process.
This allows spawning the binary in a transient dependency, something that was rather awkward before: https://github.com/wasm-tool/rollup-plugin-rust/commit/9027e61e6426c2cae59f6eb7b71449efb07fca7c

An alternative  would be to export it manually, as `getBinaryPath` (no underscore) or something, but  whichever can be maintained if the implementation switches from `binary-installer` or if `binary-installer` changes this "private" method.

### DIFF
--- a/npm/binary.js
+++ b/npm/binary.js
@@ -28,23 +28,4 @@ const getBinary = () => {
   return new Binary(url, { name });
 };
 
-const run = () => {
-  const binary = getBinary();
-  binary.run();
-};
-
-const install = () => {
-  const binary = getBinary();
-  binary.install();
-};
-
-const uninstall = () => {
-  const binary = getBinary();
-  binary.uninstall();
-};
-
-module.exports = {
-  install,
-  run,
-  uninstall
-};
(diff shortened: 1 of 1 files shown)


EverlastingBugstopper:
Very interesting - I'm in the midst of doing some work to get `binary-install` to a place where it's more flexible and has solid tests. Keep an eye on that repo

qm3ster:
I'd call it "semi-public" on the `Binary` class, with its underscored name, and fully private considering the [indirection in the usage example](https://github.com/EverlastingBugstopper/binary-install/blob/4353376764a3d287e5b9b5161dfef94c37164031/packages/binary-install-example/binary.js#L68-L87), which is exactly what this package does, and this PR aims to remove.

PR #838, opened 2020-05-03
Added keywords (#707)

### LINKED ISSUE #707: Support custom section in Cargo.toml, ex. [wasm]
## 💡 Feature description
 _Prequel:_ The original issue I had faced was `wasm-pack` not pulling `keywords` from `Cargo.toml` into generated `package.json`. This affects the visibility of the package in the registry. The solution so far is manual editing of `package.json` which is not ideal, imho.

 _Idea evolution:_ Why not to add support for a custom `[wasm]` (or `[package.wasm]`) section to `Cargo.toml`? This can be  a set of key/value pairs to put into generated `package.json`. It can even help overriding same-name field from `[package]` specifically for wasm target.

#### 💻 Basic example
```toml
[package]
name = "my-lib" # this name will be used on crates.io
                # package may be used as a native lib without possible confusion
[package.wasm]
name = "my-lib-wasm" # override, this name will be used on npm
keywords = ["one", "two", "three"] # original to generated package.json
```

What do you think?

PS: Would be happy to pick this one up :nerd_face:

lucashorward:

```bash
```

Closes #707 

package.json files usually contain a `keywords` array so that npm can make searching easier. Currently the keywords from the Cargo.toml are not exported to the package.json, meaning they have to be done manually before a package is published to npm, or be difficult to find on npm.

This is my first edit, so please let me know if I have made some junior mistakes

### DIFF
--- a/src/manifest/mod.rs
+++ b/src/manifest/mod.rs
@@ -390,7 +390,8 @@ struct NpmData {
     files: Vec<String>,
     dts_file: Option<String>,
     main: String,
-    homepage: Option<String>, // https://docs.npmjs.com/files/package.json#homepage
+    homepage: Option<String>, // https://docs.npmjs.com/files/package.json#homepage,
+    keywords: Option<Vec<String>>, // https://docs.npmjs.com/files/package.json#keywords
 }
 
 #[doc(hidden)]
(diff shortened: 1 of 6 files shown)


lucashorward:
The test is failing for `build_with_and_without_wasm_bindgen_debug` - but when I run `cargo test` on master, this failure is also there..
Do I have a wrong configuration setting or something?

Issue #698, opened 2019-08-08
wasm-pack test <extra_options> does not work as expected


alexlapa:
## 🐛 Bug description

`wasm-pack test` does not pass  <extra_options> to `cargo build --tests --target wasm32-unknown-unknown`.

#### 🤔 Expected Behavior
```
wasm-pack test --firefox -- --features some_feature
```

should invoke

```
cargo build --tests --target wasm32-unknown-unknown --features some_feature
```

#### 👟 Steps to reproduce

Running 
```
wasm-pack test --firefox -- --features some_feature
```

Causes error during tested crate compilation.

Last log entry:
``` 
INFO 2019-08-08T19:58:57Z: wasm_pack::child: Running "cargo" "build" "--tests" "--target" "wasm32-unknown-unknown"
```

However, running:
```
cargo test --target wasm32-unknown-unknown --features some_feature
```
works.

Tell me if you need more details.

#### Possible solution

It seems to me that there is no reason to run `cargo build --tests --target wasm32-unknown-unknown` when running `wasm-pack test`. 

 `wasm-pack test` will run `wasm-bindgen-test`, which will run `cargo test`, which will run `cargo build --tests` correctly propagating all extra options.

i.e. you can just remove [step_build_tests](https://github.com/rustwasm/wasm-pack/blob/439e523192676fb481b14a79ea88d03a953164e0/src/command/test.rs#L193)

#### 🌍 Your environment
Include the relevant details of your environment.
wasm-pack version: 0.8.1
rustc version: 1.36.0 (a53f9df32 2019-07-03)

alexlapa:
@ashleygwilliams ,

Added reproduction https://github.com/alexlapa/wasm-pack-698-reproduction

najamelan:
yes, this is an annoying bug.

lukidoescode:
@ashleygwilliams is the label `needs reproduction` still valid?

najamelan:
I am using it with `-- --all-features` and it works, but I had to remove the `.cargo/config` with:

```toml
[target.wasm32-unknown-unknown]
runner = 'wasm-bindgen-test-runner'
```
It seems `wasm-pack` doesn't like this one, so it's not possible to switch between `cargo test` and `wasm-pack test`.

alexlapa:
Retested [reproduction](https://github.com/alexlapa/wasm-pack-698-reproduction) with latest wasm-pack (0.9.1), it is still reproducible.

```
[...]

PR #851, opened 2020-05-18
Passes `extra_options` to `cargo build --tests` when running `wasm-pack test`.

azriel91:
Fixes https://github.com/rustwasm/wasm-pack/issues/698.

Before this change, when running `wasm-pack test -- <extra_options>`, `wasm-pack` runs

```bash
cargo build --tests --target wasm32-unknown-unknown
cargo test --target wasm32-unknown-unknown <extra_options>
```

Now it does:

```bash
cargo build --tests --target wasm32-unknown-unknown <extra_options>
cargo test --target wasm32-unknown-unknown <extra_options>
```

The command to build tests is only used once throughout the whole code base, that's why I passed the `extra_options` through without wrapping in `Option<_>`.

Comes with a `compile_error!` test too :v:.

---

```bash
```

:v: welcome :bowing_man:!


najamelan:
So if I get it, users must correctly use `--` to pass arguments to both build and test and then `--` again to pass options to `cargo test` only. If this is considered the correct UI design, then it should at least be documented in the help message of `wasm-pack help test` and in the rust wasm book. A helpful reminder if the user passes extra arguments and build fails might be good as well.

azriel91:
no, the part after the first `--` is exactly the same as what you'd do for `cargo test`, i.e. if you normally would run:

```bash
cargo test --features "feature" -- test_name
```

under `wasm-pack` you would run:

```bash
wasm-pack test -- --features "feature" -- test_name
```

The special handling for `build` is, we don't pass through the `test_name` part because when running `cargo build`, the `test_name` filter isn't accepted / doesn't make sense. 

> Not sure if `cargo build` emits a special exit status when it doesn't recognize input arguments.

It does, which is how I discovered the `test_name` filter being passed through causing the command to fail.

Pauan:
Thanks, though I think this needs to wait for https://github.com/rustwasm/wasm-pack/pull/805 to land, since it changes the way that `--` works.

azriel91:
Heya, I've rebased over `master`. Given #805 is closed, this one should be ready to review.

najamelan:
Just noticed I didn't get back to this before. Just know that with cargo test you can do:

`cargo test test_name` or even a partial test name, without `--`. 

That's why I think this isn't very intuitive and should be well documented in `wasm-pack help test` and the wasm guide.

azriel91:
@najamelan Heya, I had one more go, and managed to convince `clap` to not require the extra `--` before forwarding all arguments through (https://github.com/clap-rs/clap/issues/1847's investigation says it's impossible, but I had another go).
To enable this, I had to:

* Change the `Cargo.toml` from being a positional argument, to be under the `--manifest-path` option.
* Use some not-so-recommended `AppSettings`.

If that doesn't sit well for this project, I could undo it and just add `--` to the docs, but the user experience is much better without the extra `--`. Also, the `--manifest-path` is consistent with `cargo`'s interface, so I thought that's alright

azriel91:
Heya @najamelan / @Pauan, I managed to get `wasm-pack` to now have the same command line interface as `cargo`.

So now users can run:

```bash
wasm-pack test --features "feature" -- test_name  # which corresponds to:
#   cargo test --features "feature" -- test_name
```

The `path` positional parameter is retained as the first (optional) argument

In the implementation, I tell `clap` to take in `path` and all `--hyphen-options` in a single `Vec` field, and if the first String does not begin with `-`, then we take that to be the crate path. The remainder is passed through to `cargo`.

I'm really keen on seeing this through, as it would enable the `governor` crate to be published with WASM support.

Pauan:
@azriel91 Unfortunately it's impossible, because only @ashleygwilliams has the rights to publish new changes, and she has disappeared.

kigawas:
Any ETA on merging this?

Saruniks:
It would be good to have this functionality.

PR #870, opened 2020-06-29
Search wasm-opt in PATH before trying to download the prebuilt binary

### LINKED ISSUE #869: wasm-pack fails to locate wasm-opt in PATH
## 🐛 Bug description
wasm-pack does not search PATH for wasm-opt, at least not on targets where there is no prebuilt binary

#### 🤔 Expected Behavior
wasm-pack should find wasm-opt in PATH

#### 👟 Steps to reproduce
Not sure if this is specific to FreeBSD or if it also happens on other operating systems.
- run wasm-pack
- observe it fails because wasm-opt is not available
- install wasm-opt
- run wasm-pack
- observe that it does not find wasm-opt even though it is in path.

#### 🌍 Your environment
wasm-pack version: 0.9.1
rustc version: 1.44.1 (FreeBSD ports version)
FreeBSD version: FreeBSD 12.1-RELEASE-p3 GENERIC  amd64
wasm-opt version: 93

hyd-dev:
Fixes #869.

### DIFF
--- a/src/wasm_opt.rs
+++ b/src/wasm_opt.rs
@@ -3,7 +3,8 @@
 use crate::child;
 use crate::install::{self, Tool};
 use crate::PBAR;
-use binary_install::Cache;
+use binary_install::{Cache, Download};
+use log::debug;
 use std::path::Path;
 use std::process::Command;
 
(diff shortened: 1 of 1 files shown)


RReverser:
I made an alternative PR for the same issue - https://github.com/rustwasm/wasm-pack/pull/876 - which, I believe, is slightly better because it's a more generic fix that would help future tooling as well.

hyd-dev:
> I made an alternative PR for the same issue - #876 - which, I believe, is slightly better because it's a more generic fix that would help future tooling as well.

Yes, it's better, so I'll close this PR. I'm looking forward to #876.

Issue #823, opened 2020-04-02
65214 segmentation fault (core dumped)  wasm-pack help


KSXGitHub:
## 🐛 Bug description

I run `wasm-pack help` and it crashes (sometimes).

```
wasm-pack 0.9.1
Ashley Williams <ashley666ashley@gmail.com>
📦 ✨  pack and publish your wasm!

USAGE:
    wasm-pack [FLAGS] [OPTIONS] <SUBCOMMAND>

FLAGS:
    -h, --help       Prints help information
    -q, --quiet      No output printed to stdout
    -V, --version    Prints version information
    -v, --verbose    Log verbosity is based off the number of v used

OPTIONS:
        --log-level <log_level>    The maximum level of messages that should be logged by wasm-pack. [possible values:
                                   info, warn, error] [default: info]

SUBCOMMANDS:
    build      🏗️  build your npm package!
    help       Prints this message or the help of the given subcommand(s)
    login      👤  Add an npm registry user account! (aliases: adduser, add-user)
    new        🐑 create a new project with a template
    pack       🍱  create a tar of your npm package but don't publish!
    publish    🎆  pack up your npm package and publish!
    test       👩‍🔬  test your wasm!
malloc_consolidate(): invalid chunk size
[1]    65214 segmentation fault (core dumped)  wasm-pack help
```

#### 🤔 Expected Behavior

Run without crash

#### 👟 Steps to reproduce

```sh
cargo +stable install wasm-pack
wasm-pack help
```

#### 🌍 Your environment

wasm-pack version: 0.9.1
rustc version: 1.42.0

ahwatts:
I'm getting this, too. I installed the prebuild binary.

Interestingly, if I run it with `RUST_BACKTRACE=1`, the segfault goes away. If I run it in `gdb`, I get this:

```
awatts@ironic-mullet ~ $ gdb $(which wasm-pack)
GNU gdb (GDB) Fedora 9.1-5.fc32
Copyright (C) 2020 Free Software Foundation, Inc.
License GPLv3+: GNU GPL version 3 or later <http://gnu.org/licenses/gpl.html>
This is free software: you are free to change and redistribute it.
There is NO WARRANTY, to the extent permitted by law.
Type "show copying" and "show warranty" for details.
This GDB was configured as "x86_64-redhat-linux-gnu".
Type "show configuration" for configuration details.
For bug reporting instructions, please see:
<http://www.gnu.org/software/gdb/bugs/>.
[...]

PR #871, opened 2020-06-30
Update curl to 0.4.30

hyd-dev:
~~Fixes~~ Addresses #823.
0.4.30 includes the alexcrichton/curl-rust#333 fix (alexcrichton/curl-rust#334) and makes triggering sfackler/rust-openssl#1293 less possible - at least `wasm-pack` does not segfault that frequently.
We can do nothing here except for calling `libc::_exit()` at the end of the program until sfackler/rust-openssl#1293 gets fixed (edit: there is sfackler/rust-openssl#1324 - this PR may not be needed after sfackler/rust-openssl#1324 gets merged).

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -100,7 +100,7 @@ name = "binary-install"
 version = "0.0.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 dependencies = [
- "curl 0.4.25 (registry+https://github.com/rust-lang/crates.io-index)",
+ "curl 0.4.30 (registry+https://github.com/rust-lang/crates.io-index)",
  "dirs 1.0.5 (registry+https://github.com/rust-lang/crates.io-index)",
  "failure 0.1.6 (registry+https://github.com/rust-lang/crates.io-index)",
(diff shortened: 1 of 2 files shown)


0xd4d:
Is there a reason you didn't update `Cargo.toml`? If 0.4.30 is required to fix the bug the fix should be there as well.

https://github.com/rustwasm/wasm-pack/blob/9f9634ca9ffa9147c044f947a31808f02715aa44/Cargo.toml#L17

hyd-dev:
> Is there a reason you didn't update Cargo.toml? If 0.4.30 is required to fix the bug the fix should be there as well.

Sorry, I ignored it. I've updated it.

hyd-dev:
sfackler/rust-openssl#1324 has been [merged](https://github.com/sfackler/rust-openssl/commit/08d8367ff2a5a1e34102090296765bff12044f37). I tested it after updating and found it doesn't crash anymore - it fixed #823 completely, so this PR is not needed anymore. [`openssl`](https://github.com/sfackler/rust-openssl) should be updated (possibly by using `[patch.crates-io]`) instead.

PR #888, opened 2020-07-31
feat: reference types argument

Tarnadas:
This PR adds a new argument `--reference-types`, which enables reftype compilation for wasm-bindgen.

This feature is hoped to enable more efficient communication between the host (JS) and the wasm module, but not yet all browsers support it.

For more information about reference types, you should read the [wasm-bindgen Guide](https://rustwasm.github.io/docs/wasm-bindgen/reference/reference-types.html#support-for-reference-types).

This argument will automatically disable `wasm-opt`.
Otherwise compilation would crash with the following error:

```
[parse exception: Only 1 table definition allowed in MVP (at 0:4234)]
Fatal: error in parsing input
```

### DIFF
--- a/src/command/build.rs
+++ b/src/command/build.rs
@@ -27,6 +27,7 @@ pub struct Build {
     pub crate_data: manifest::CrateData,
     pub scope: Option<String>,
     pub disable_dts: bool,
+    pub reference_types: bool,
     pub target: Target,
     pub profile: BuildProfile,
     pub mode: InstallMode,
@@ -119,6 +120,11 @@ pub struct BuildOptions {
     /// this flag will disable generating this TypeScript file.
     pub disable_dts: bool,
 
+    #[structopt(long = "reference-types")]
+    /// Enable support for reference types. This feature is hoped to enable more efficient
+    /// communication between the host (JS) and the wasm module, but not yet all browsers support it.
+    pub reference_types: bool,
+
     #[structopt(long = "target", short = "t", default_value = "bundler")]
     /// Sets the target environment. [possible values: bundler, nodejs, web, no-modules]
     pub target: Target,
@@ -160,6 +166,7 @@ impl Default for BuildOptions {
             scope: None,
             mode: InstallMode::default(),
             disable_dts: false,
+            reference_types: false,
             target: Target::default(),
             debug: false,
             dev: false,
@@ -196,6 +203,7 @@ impl Build {
             crate_data,
             scope: build_opts.scope,
             disable_dts: build_opts.disable_dts,
+            reference_types: build_opts.reference_types,
             target: build_opts.target,
             profile,
             mode: build_opts.mode,
@@ -370,6 +378,7 @@ impl Build {
             &self.out_dir,
             &self.out_name,
             self.disable_dts,
+            self.reference_types,
             self.target,
             self.profile,
         )?;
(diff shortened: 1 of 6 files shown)


Urhengulas:
Hello @Tarnadas 👋🏽

The patch looks great :)

You should be able to make `wasm-opt` run successfully by passing the `--enable-reference-types` command line flag. This way you could still get the optimization 🚀

richardlett:
I tried this out, and found this bug. Not sure if its related to this particular commit, or due to wasm-bindgen.

https://github.com/rustwasm/wasm-bindgen/issues/2291

Tarnadas:
This shouldn’t be related to this particular commit.

It rather sounds like a memory allocation issue, which might be totally normal. I’m not sure if there is an option for wasm-bindgen to increase the initial memory size, but you could do it manually in the generated JS glue code file and see if that fixes your problem:
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/WebAssembly/Memory

Also I don’t quite get why this shouldn’t happen without reference types enabled.

Btw today I will be fixing the wasm-opt stuff. Thanks for the hint @Urhengulas

richardlett:
Excuse my ignorance on javascript/wasm.

I cant seem to find in the glue code where to memory is allocated -- I can find  a buffer with ref-types disabled, as there is a "heap", but I can't see anything of the like here.

Strangely, even if I'm just missing that, I preallocated  using `Vec::with_capacity(1 million)`, (which is not included in the benchmarks), and the issue is totally outside of that (and there's no further allocations are made besides the Vec, unless the gluecode is doing it somehow, but that does not seem to be the case).

Tarnadas:
Apparently the JS glue code does not instantiate a WebAssembly memory, which would result in the default values being used.

You should see a line in your JS glue code which looks somewhat like this:
```js
const wasmModule = new WebAssembly.Module(bytes);
const wasmInstance = new WebAssembly.Instance(wasmModule, imports);
```

Here you can instantiate a WebAssembly Memory with a larger initial size:
```js
const wasmModule = new WebAssembly.Module(bytes);
const wasmInstance = new WebAssembly.Instance(wasmModule, {
  js: {
    mem: new WebAssembly.Memory({ initial: /* large number */, maximum: /* large number */ })
  },
  ...imports
});
```

whatisaphone:
The corresponding flag for wasm-bindgen is called `--reference-types`. Maybe we should use the same longer name here for consistency?

Tarnadas:
That sounds definitely like a good idea. I’ll update my PR, but I’m a little worried about the maintenance of this repository

Tarnadas:
Hey @drager @azriel91,
I just rebased my PR with the latest master branch.
Any chance this could be reviewed and merged?

azriel91:
@Tarnadas heya, I'm simply a fellow contributor; that said the changes look good :v:

src/command/build.rs
```diff
@@ -378,6 +387,9 @@ impl Build {
     }
 
     fn step_run_wasm_opt(&mut self) -> Result<(), Error> {
+        if self.reference_types {
+            return Ok(());
```
mikefarmer01, line 391: This should generate at least a debug message to notify the user that wasm-opt had been disabled.

mikefarmer01 (changes requested):

PR #897, opened 2020-08-09
feat: added possibility to add a version for gecko driver to install

arn-the-long-beard:
Hello guys !!! 

I checked! 📦✅

- [x] I have the latest version of `rustfmt` installed
```bash
->  info: component 'rustfmt' for target 'x86_64-unknown-linux-gnu' is up to date

```
- [x] I ran `cargo fmt` on the code base before submitting 
- [x] I reference which issue is being closed in the PR text
->  https://github.com/rustwasm/wasm-bindgen/issues/2261#issuecomment-670961373

Since the 27th of July there is a bug with headless mode for firefox on the latest **geckodriver**.
So, it could be nice to give us the opportunity to choose which version we want to install.

This PR is draft, because I am not pro and maybe we can find a good solution together 😄

My first reflex is that we could use environment variable to do that, but maybe we can use `feature` or something else.

Thank you for your inputs.

By the way, the test I made is failing, and even with debugging in my IDE, I do not know where it is failing. Can somebody give me a hand on this ? 

✨ Thank you for your inputs ✨

Issue : https://github.com/rustwasm/wasm-bindgen/issues/2261#issuecomment-670961373

### DIFF
--- a/src/test/webdriver/geckodriver.rs
+++ b/src/test/webdriver/geckodriver.rs
@@ -4,6 +4,7 @@ use chrono::DateTime;
 use failure::{self, ResultExt};
 use install::InstallMode;
 use stamps;
+use std::env::VarError;
 use std::path::PathBuf;
 use target;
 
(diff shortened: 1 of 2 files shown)


src/test/webdriver/geckodriver.rs
```diff
@@ -88,13 +89,21 @@ fn get_geckodriver_url(target: &str, ext: &str) -> String {
...
-
+    let specified_geckodriver_version: Result<String, VarError> =
+        std::env::var("SPECIFIED_GECKO_DRIVER_VERSION");
     let geckodriver_version = if target::WINDOWS {
         log::info!(
             "[geckodriver] Windows detected, holding geckodriver version to {}",
             DEFAULT_WINDOWS_GECKODRIVER_VERSION
         );
         DEFAULT_WINDOWS_GECKODRIVER_VERSION.to_owned()
+    } else if specified_geckodriver_version.is_ok() {
```
MartinKavik, line 100: I would probably switch conditions `if target::WINDOWS` and `if specified_geckodriver_version.is_ok()` otherwise you wouldn't be able to specify the version on Windows.

arn-the-long-beard: I wasn't sure because it seems the version for windows had to be 0.24.0.

arn-the-long-beard: Done :)

arn-the-long-beard:
Hello guys, what should we do with this PR ?

Issue #879, opened 2020-07-19
Targeting Deno


eliorivero:
## 🐛 Bug description
I saw this https://rustwasm.github.io/docs/wasm-bindgen/reference/deployment.html#deno mentioning that you could pass `--target deno` and build to Deno. I tried:

```
wasm-pack build --target deno
```

and got 

```
error: Invalid value for '--target <target>': Unknown target: deno
```

#### 🤔 Expected Behavior
A package suitable to be used in Deno should've been created.

#### 👟 Steps to reproduce

```
cargo new rust-deno-test --lib
cd rust-deno-test
```

Edit `src/lib.rs` and replace its contents with:
```
use wasm_bindgen::prelude::*;

#[wasm_bindgen]
pub fn factorial( x: u64 ) -> u64 {
    match x {
        0 | 1 => 1,
        _ => x * factorial( x - 1 ),
    }
}
```

Edit `Cargo.toml` and merge:
```
[dependencies]
wasm-bindgen = "0.2.65"

[lib]
crate-type =["cdylib"]
```

Then run
```
wasm-pack build --target deno
```

#### 🌍 Your environment
wasm-pack version: 0.9.1
rustc version: 1.44.1

#### Notes
The tool `ssvmup` can successfully convert the same code to a package that's usable by Deno, with a similar command `ssvmup build --target deno`.

stevebest:
`wasm-bingen` seems to have recently gained support for `deno` target, but the option is not forwarded to it by `wasm-pack`.

I have quickly hacked a [patch](https://github.com/stevebest/wasm-pack/commit/81804faf883547ad74065da0e50adc0ac6184af8) which seems to work at first glance. It does a bit more than required, even: for example, it generates a `package.json`, which Deno doesn't use.

I could work on a PR, but consider the fact that I only had a proper look at Deno yesterday.

drager:
Fixed in [v0.11.0](https://github.com/rustwasm/wasm-pack/releases/tag/v0.11.0).

Issue #672, opened 2019-07-01
I want to build wasm for Deno.


kahirokunn:
## 💡 Feature description
I want to build wasm for Deno.
https://github.com/denoland/deno

I try write rust and do wasm-pack.
And then I got .wasm file, I put it file on the Deno.
But the wasm file never work....(´；ω；｀)

#### 💻 Basic example

```sh
$ wasm-pack build --target deno
```

csmoe:
`wasm-pack`'s build target reiles on `wasm-bindgen` cc https://rustwasm.github.io/docs/wasm-bindgen/reference/cli.html?highlight=target#--target

ibaryshnikov:
@kahirokunn is there `require` function in deno?

ashleygwilliams:
related https://github.com/denoland/deno/issues/1325

ashleygwilliams:
@kahirokunn in order to run wasm-packed wasm, you'll need to require the js wrapper generated by wasm-bindgen. can deno load JS?

kevinkassimo:
fyi Deno has resolved the original issue posted above and have recently just introduced WASM imports (should be very similar to Node.js's `--experimental-wasm-modules`):
```ts
import { exportedFunc1, exportedFunc2 } from "./my.wasm";
```
If `my.wasm` imports from other JS/TS files, it could be done by specifying import names like
```wat
(module
  (import "./wasm-dep.js" "jsFn" (func $jsFn (result i32)))
  (import "http://127.0.0.1:4545/cli/tests/051_wasm_import/remote.ts" "jsRemoteFn" (func $jsRemoteFn (result i32)))
)
```
with corresponding local/remote file set up.

ytoune:
I use below for a while

```sh
cargo generate --git https://github.com/rustwasm/wasm-pack-template -n myproject
cd myproject
# deno has not alert but has console.log.
sed -i -e "s/fn alert(s: &str);/#[wasm_bindgen(js_namespace = console, js_name = log)] fn alert(s: \\&str);/" src/lib.rs

wasm-pack build --target web --out-name index
# deno hasn't supported fetch('file:// ...') yet.
[...]

PR #908, opened 2020-08-27
Support deno target

juzi5201314:
- [x ] You have the latest version of `rustfmt` installed
- [x ] You ran `cargo fmt` on the code base before submitting

- Related issues
#672 
#879


OliverBrotchie:
Is there any update on this? It is mentioned in the docs already [here](https://rustwasm.github.io/docs/wasm-bindgen/reference/deployment.html?highlight=deno#deno).

juzi5201314:
As https://github.com/rustwasm/wasm-pack/issues/879#issuecomment-662835589 said, `wasm-bingen seems to have recently gained support for deno target, but the option is not forwarded to it by wasm-pack.`

OliverBrotchie (approved):
The changes are very minor, I cannot foresee any conflicts emerging.

Sh4rK:
Hey everyone,

This seems like a relatively small change, but would help greatly in writing wasm modules for Deno, is there anything blocking this? Is there anything to do to get it merged?

src/command/utils.rs
```diff
@@ -36,9 +37,14 @@ fn find_manifest_from_cwd() -> Result<PathBuf, failure::Error> {
...
-pub fn create_pkg_dir(out_dir: &Path) -> Result<(), failure::Error> {
+pub fn create_pkg_dir(out_dir: &Path, target: &Target) -> Result<(), failure::Error> {
     fs::create_dir_all(&out_dir)?;
-    fs::write(out_dir.join(".gitignore"), "*")?;
+    let mut gitignore_contents = String::new();
+    match target {
+        Target::Deno => {}
+        _ => gitignore_contents.push_str("*")
+    }
+    fs::write(out_dir.join(".gitignore"), gitignore_contents)?;
```
Sh4rK, line 47: Instead of writing an empty file, it would be cleaner to not write it at all, I think.

But also, I'm nut sure this is necessary, why would targeting Deno be different in this regard? You might want to package up the resulting files into a separate repository (equivalent to putting it on npm for node), but at the same time want to keep the rust lib's source clean.

Having a wildcard .gitignore doesn't make sense in any other way, you obviously want to keep the results _somewhere_.

olanod:
This looks helpful, is it missing anything?

Mipsters:
I was using this branch, couple of problems I had
1. `wasm-pack build --help` didn't list `deno` as a target option
2. the `.js` generated in pkg doesn't manipulate the url correctly for me
    1. I use windows, yet the path string still comes with `/` and not `\`, so `Deno.build.os === 'windows' ? '\\' : '/'` give a wrong result
    2. `new URL(import.meta.url).pathname` gives a path with `\` before `C:` for some reason?

except for this, it works perfectly!
took me some time to realize that I needed the `.js` in my deno code, but it did work!

egfx-notifications:
> I am using this branch, couple of problems I had
> 
>     1. `wasm-pack build --help` didn't list `deno` as a target option
> 
>     2. the `.js` generated in pkg doesn't manipulate the file path correctly for me
>        (may be related to the fact that I use git bash as my main console)
>        
>        1. I use windows, yet the path string still comes with `/` and not `\`, so `Deno.build.os === 'windows' ? '\\' : '/'` give a wrong result
>        2. `new URL(import.meta.url).pathname` gives a path with `\` before `C:` for some reason?
> 
> 
> except for this, it works perfectly! took me some time to realize that I needed the `.js` suffix in the import in my deno code, but it did work!
> 
> **edit:** after thinking about it a bit, the problems in 2 are probably problems in `wasm-bingen` and not this pr sorry if that's the case

@Mipsters Your point 2.2 seems to be fixed in the current wasm-bindgen release, the path is concatenated differently now.  
2.1 I don't get right now, is `Deno.build.os === 'windows' ? '\\' : '/'` something you use in your code or something that wasm-bindgen created? Either way, I can't see this in output of current wasm-bindgen.  
Your first point would definitely be an improvement to this PR.

Mipsters:
@egfx-notifications wow, I barely remember what I did then, but I think that the ternary condition was not mine  
if I remember correctly, it was of some inner library, don't remember if it was part of the pr or something else
you can run a simple script to check if this is resolved, if `Deno.build.os === 'windows'` is true on windows then there is no issue
if you don't have a windows machine available, hit me up, but it'll take time for me to remind myself how to use any of this things...

src/command/utils.rs
Mipsters: I think this was the default behavior, he just removed that option for deno
the deleted line in 41 wrote `*` to `.gitignore` in every case, but @juzi5201314 probably concluded that it is not relevant for deno
[this is still the state in the main branch (as of 2nd Feb, 2022)](https://github.com/rustwasm/wasm-pack/blob/1dc41bc09745b5c5f87fca9eddfb65b5b3586ddc/src/command/utils.rs#L42)
[and in the latest release atm](https://github.com/rustwasm/wasm-pack/blob/v0.10.2/src/command/utils.rs#L42)
I agree about preferring not to create the `.gitignore` at all, but keeping the `*` for all other targets is just keeping the same behavior
changing this behavior is outside of the scope of this pr :)

egfx-notifications:
> @egfx-notifications wow, I barely remember what I did then, but I think that the ternary condition was not mine if I remember correctly, it was of some inner library, don't remember if it was part of the pr or something else you can run a simple script to check if this is resolved, if `Deno.build.os === 'windows'` is true on windows then there is no issue if you don't have a windows machine available, hit me up, but it'll take time for me to remind myself how to use any of this things...

Then this seems to be fixed now. Tried cmd, Powershell and Git Bash on Windows, all give the correct result.

src/command/utils.rs
Sh4rK: My comment is not very clear, reading it again now. I think my point was that this function should be left as-is, no need to special case it for Deno.

drager:
Fixed in #1117.

PR #883, opened 2020-07-21
Remove parking_lot from direct dependency

tetsuharuohzeki:
We don't use this crate now. This fix https://github.com/rustwasm/wasm-pack/issues/882

Make sure these boxes are checked! 📦✅

- [x] You have the latest version of `rustfmt` installed
```bash
$ rustup component add rustfmt
```
- [x] You ran `cargo fmt` on the code base before submitting
- [x] You reference which issue is being closed in the PR text

✨✨ 😄 Thanks so much for contributing to wasm-pack! 😄 ✨✨

PR #924, opened 2020-10-10
Cargo opts

### LINKED ISSUE #923: Would like a way to pass arguments to cargo, before "build"
### Motivation

I would like to be able to pass global options to cargo.  In particular, I use `--locked` and `--offline` quite heavily.  `--locked` is useful to avoid unwanted updates.  `--offline` is important if I'm in a poor (or expensive) network environment.

## 💡 Feature description

Currently, `--` introduces options to `cargo build`.  It's not so easy to shoehorn an additional free-form list of argualments.  I suggest we take a leaf from compiler drivers like `gcc`, and introduce a repeatable option that adds one additional cargo argument.

#### 💻 Basic example

So, you could say something like
```
   wasm-pack --cargo-opt=--locked --cargo-opt=--verbose --cargo-opt=--color=never build
```

If your cargo is actually rustup pretending to be cargo (so normally you can say `cargo +nightly build`) and you encounter #629, you could say
```
  wasm-pack --cargo-opt=+nightly build
```

### Alternatives

The only plausible alternative for a user like me right now, without some change to `wasm-pack`, is to live with the unwanted lockfile updates, or to put a stunt wrapper for cargo on the PATH.  Or to abandon `wasm-pack` and go to `wasm-bindgen-cli`.

As for the proposed command line syntax for wasm-pack, I think `--cargo-opt=...` is probably the best but I'm happy to implement whatever syntax you prefer.  If you are looking for prior art, `gcc` has `-Wl,--some-linker-option`, but that's not very readable.  `-C...` would be a possibility but steals a single-letter option.

I'm sure the value ought to be the whole thing passed to cargo, not just the part after `--`.  Anything else would be really very confusing.

Another possibilty would be to do something more complicated, perhaps with `---` meaning something special, but it doesn't seem desirable for wasm-pack's command line parsing to be funky and exciting.  It should be straightforward, and rather unusual use cases like mine definitely don't need super-brief syntax.

ijackson:
Hi.  Here is a series let to the user influence the way wasm-pack runs cargo.  Closes #923.  Please see that issue for rationale and other options I considered.

I hope this meets with your approval.  Please let me know what you think.

#### Template checklist

```bash
```


ijackson:
FTAOD these conflicts are due to #883. I am no longer using wasm-pack, but if it is being maintained and this MR is welcome, I could fix it.

Issue #919, opened 2020-09-29
wasm-opt Tool version should be bumped 


michelhe:
## 🐛 Bug description
The prebuilt wasm-opt version_90 segfaults when optimizing my crate with `-enable-mutable-globals`.
When I manually use version_97, it works 

I think it will also help fix #886 

#### 👟 Steps to reproduce
I didn't try to test any other project, but It specifically segfaults in mine: https://github.com/michelhe/rustboyadvance-ng/tree/master/platform/rustboyadvance-wasm
add to Cargo.toml
```toml
[package.metadata.wasm-pack.profile.release]
wasm-opt = ["-Oz", "--enable-mutable-globals"]
```
run `wasm-pack build --release`

#### 🌍 Your environment
wasm-pack version: 0.9.1
rustc version: rustc 1.43.0 (4fb7144ed 2020-04-20)

panstromek:
For the record, I had the same issue, reported it in binaryen:

https://github.com/WebAssembly/binaryen/issues/3513

PR #926, opened 2020-10-19
bump: wasm_opt, chromedriver & geckodriver version

meck93:

### Description
Fixes the issue reported in #919

### DIFF
--- a/src/install/mod.rs
+++ b/src/install/mod.rs
@@ -165,7 +165,7 @@ fn prebuilt_url(tool: &Tool, version: &str) -> Result<String, failure::Error> {
     let target = if target::LINUX && target::x86_64 {
         match tool {
             Tool::WasmOpt => "x86_64-linux",
-            _ => "x86_64-unknown-linux-musl",
+            _ => bail!("Unrecognized target!"),
         }
     } else if target::LINUX && target::x86 {
         match tool {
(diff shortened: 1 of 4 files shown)


lucashorward:
Seems good, but the download urls have changed (since the binaries in the release have a different naming convention now, it seems - https://github.com/WebAssembly/binaryen/releases), so if you could rename those accordingly, that would be great!

meck93:
Hi @lucashorward 
Thanks for the reply. Maybe I don't fully understand what you mean. Could you be a bit more specific? 

The download URL for **binaryen** version_97 looks like this: https://github.com/WebAssembly/binaryen/releases/download/version_97/binaryen-version_97-x86_64-linux.tar.gz

And, this is the way that we already compose it:  
```Rust
Ok(format!(
    "https://github.com/WebAssembly/binaryen/releases/download/{vers}/binaryen-{vers}-{target}.tar.gz",
    vers = "version_97",
    target = target,
))
```         

Or, did you mean something different?

lucashorward:
Yes, that is the url I mean, and I see you've correctly updated the target variable that's used in that link for linux (which has changed from `x86-linux` to `x86_64-linux`- in `src/install/mod.rs` line 167), but the path has also changed for windows and macos in this newer release of wasm_opt. So those paths should be updated here as well:
For 64-bit macos, it should go from `x86_64-apple-darwin` to `x86_64-macos` (line 176)
For 64-bit windows, it should go from `x86-windows` to `x86_64-windows` (line 179)

The failing CI test is because of this, the url for linux is correct but for the other systems a 404 is returned.

On top of that, I see there are no 32-bit versions listed - not sure what the impact of that will be to be honest. Maybe @ashleygwilliams has an idea on that. On the one hand, we don't want to keep using this older version, but I'm unsure if this will break all 32-bit builds.

meck93:
Thanks a lot for the clarification! That makes a lot of sense. I'm on Linux and that's probably why I only updated the Linux URL when I compiled `wasm-pack` locally (before opening the PR). But I'll fix the other ones as well!

drager:
Already fixed in new releases.

PR #936, opened 2020-11-17
Recent node versions can import ESModulesPackage output if "type": "module" is specified

serprex:
With this change ESModulePackage can be imported by nodejs without getting raising SyntaxError about how the pkg directory is a CommonJS module

https://nodejs.org/api/packages.html#packages_determining_module_system

### DIFF
--- a/src/manifest/mod.rs
+++ b/src/manifest/mod.rs
@@ -424,7 +424,10 @@ impl CrateData {
         let current_idx = data
             .packages
             .iter()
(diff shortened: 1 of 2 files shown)


vovacodes:
Can we merge it? "type": "module" is also a best practice for the browser-targeted packages that export code using esm format so that the modern bundlers can do a better job handling them. We should also include the `"exports": "./<crate-name>.js"` field (https://nodejs.org/api/packages.html#packages_exports_sugar)

PR #955, opened 2021-01-04
Support chromedriver on Apple Silicon.

### LINKED ISSUE #954: Support chromedriver on apple silicon
Right now if I use wasm-pack to run my tests on my new M1-using macbook, I run into [this error message](https://github.com/rustwasm/wasm-pack/blob/9f9634ca9ffa9147c044f947a31808f02715aa44/src/test/webdriver/chromedriver.rs#L41). The text of the error is [currently correct](https://bugs.chromium.org/p/chromedriver/issues/detail?id=3688) but won't be once the binaries are available.

anp:
~~From local testing this works fine and native binaries aren't yet
available: https://bugs.chromium.org/p/chromedriver/issues/detail?id=3688.~~

Uses the newly-available `_m1` releases of chromedriver.

Resolves #954.

(checked the boxes!)

### DIFF
--- a/src/test/webdriver/chromedriver.rs
+++ b/src/test/webdriver/chromedriver.rs
@@ -9,7 +9,7 @@ use target;
 
 // Keep it up to date with each `wasm-pack` release.
 // https://chromedriver.storage.googleapis.com/LATEST_RELEASE
-const DEFAULT_CHROMEDRIVER_VERSION: &str = "79.0.3945.36";
+const DEFAULT_CHROMEDRIVER_VERSION: &str = "88.0.4324.96";
 
 const CHROMEDRIVER_LAST_UPDATED_STAMP: &str = "chromedriver_last_updated";
 const CHROMEDRIVER_VERSION_STAMP: &str = "chromedriver_version";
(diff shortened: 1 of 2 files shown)


src/test/webdriver/chromedriver.rs
```diff
@@ -33,7 +33,7 @@ pub fn install_chromedriver(
 ) -> Result<PathBuf, failure::Error> {
     let target = if target::LINUX && target::x86_64 {
         "linux64"
-    } else if target::MACOS && target::x86_64 {
+    } else if target::MACOS {
```
richiksc, line 36:
```suggestion
    } else if target::MACOS && target::aarch64 {
        "mac64_m1"
    } else if target::MACOS && target::x86_64 {
```

@anp The chromium issue you linked to now mentions that `chromedriver` has been released with native binaries for Apple Silicon for versions 87 and 88, and sure enough, they have: https://chromedriver.chromium.org/.

The suffixes for these builds are `mac64_m1`. `target.rs` would also have to be updated, as well as `DEFAULT_CHROMEDRIVER_VERSION`.

richiksc (changes requested):

src/test/webdriver/chromedriver.rs
anp: Thank you for the heads up!

PR #961, opened 2021-01-14
Use native chromedriver binaries on Apple Silicon

### LINKED ISSUE #954: Support chromedriver on apple silicon

richiksc:
Closes #954, closes #955.

---

```bash
```


richiksc:
Closing in favor of #955.

PR #977, opened 2021-02-16
Unstable versions of cargo-generate on crates.io breaks prebuilt binaries and previously installed version checks.

### LINKED ISSUE #976: Unstable versions of cargo-generate on crates.io breaks prebuilt binary downloads.
## 🐛 Bug description
As title.

#### 🤔 Expected Behavior
The last stable version of the prebuilt binary should have been downloaded.

#### 👟 Steps to reproduce
1. Start with clean wasm-pack installation
2. Run `RUST_LOG=warn wasm-pack new foo`
3. Run into installation error:
```
WARN 2021-02-16T22:21:44Z: wasm_pack::install: could not download pre-built `cargo-generate`: failed to download from https://github.com/cargo-generate/cargo-generate/releases/download/v0.6.0-alpha.1/cargo-generate-v0.6.0-alpha.1-x86_64-unknown-linux-musl.tar.gz. Falling back to `cargo install`.
```

#### 🌍 Your environment
wasm-pack version: 0.9.1
rustc version: 1.50.0

I am currently working on a PR to fix this but would just like to report it first.

ThePuzzlemaker:
This PR fixes #976.

Recently there was a version of `cargo-generate` published to `crates.io`, `0.6.0-alpha.1`. This caused an issue because the previous API for `Krate` to check the latest version used `max_version` instead of `max_stable_version`, and there were no prebuilt binary for `0.6.0-alpha.1`. This PR fixes this issue and therefore fixes prebuilt binaries as well as the checks for previously installed versions of cargo-generate. I also changed the GitHub address to use `cargo-generate/cargo-generate` instead of `ashleygwilliams/cargo-generate`. While it still works without it, it's still just generally helpful as it saves one level of redirection which on some connections may be kind of slow.

### DIFF
--- a/docs/src/commands/new.md
+++ b/docs/src/commands/new.md
@@ -30,7 +30,7 @@ wasm-pack new myproject --template https://github.com/rustwasm/wasm-pack-templat
 The template can be an address to a git repo that contains a [`cargo-generate`]
 template.
 
-[`cargo-generate`]: https://github.com/ashleygwilliams/cargo-generate
+[`cargo-generate`]: https://github.com/cargo-generate/cargo-generate
 
 ## Mode
 
--- a/src/install/krate.rs
+++ b/src/install/krate.rs
@@ -3,7 +3,7 @@ use serde::Deserialize;
 
 #[derive(Debug, Deserialize)]
 pub struct Krate {
-    pub max_version: String,
+    pub max_stable_version: String,
 }
 
 #[derive(Debug, Deserialize)]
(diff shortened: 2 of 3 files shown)


ThePuzzlemaker:
I am expecting that some tests will fail on this PR but I'm pretty sure none of those are due to this PR itself, but due to outstanding issues.

pcardune:
Just ran into this same issue myself. Tried out your change but it appears to still be broken for me. As far as I can tell looking at [the cargo-generate releases page](https://github.com/cargo-generate/cargo-generate/releases), prebuilt binaries stopped being published since version 0.5.2, probably because of [this PR](https://github.com/cargo-generate/cargo-generate/pull/271) that refactored the CI pipeline from travis to github actions. Unclear whether that change was intentional or not.

PR #947, opened 2020-12-21
fix: Remove appveyor and travis and use Github actions for tests

ashleygwilliams:
🤕 do not merge 🤕 

fixes #594

PR #978, opened 2021-02-19
feat: adds support for M1 macbooks to curl installer

EverlastingBugstopper:
M1 macbooks can run x86_64 code in emulation mode! this PR adds support for detecting if a machine is an M1 macbook, and then directs them towards the x86_64-apple-darwin binary in both the curl installer and the npm installer :)

### DIFF
--- a/docs/_installer/init.sh
+++ b/docs/_installer/init.sh
@@ -92,6 +92,13 @@ get_architecture() {
         fi
     fi
 
+    if [ "$_ostype" = Darwin -a "$_cputype" = arm64 ]; then
+        # Darwin `uname -s` doesn't seem to lie on Big Sur
+        # but we want to serve x86_64 binaries anyway that they can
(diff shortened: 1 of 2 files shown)


chriskrycho:
Wouldn't it be preferable to just pre-build for M1? That's quite doable now (though it may require switching to GitHub Actions _a la_ #947). See for example [the commit adding support for Arm Mac builds on Volta](https://github.com/volta-cli/volta/commit/7ac38410e7347498babc10237b08d8e5575106e8).

EverlastingBugstopper:
this would be the easiest lift right now with minimal changes to the code. github actions also only supports macos-11 for preview builds, so switching to github actions at this time is not feasible.

unfortunately i think that by default, big sur requires that binaries be signed or they'll be SIGKILLed by gateway. i haven't included that work in this PR.

EverlastingBugstopper:
should merge #1000 instead of this

Issue #984, opened 2021-03-03
Don't overwrite an existing .gitignore file


sinking-point:
If a developer has a custom .gitignore in place, it is unexpected for wasm-pack to overwrite it. wasm-pack should check for an existing .gitignore before writing a new one. Currently, if a developer wants a custom .gitignore, they have to reset it every time they run wasm-pack.

Implementing this makes it no harder for developers who want the default .gitignore. They just have to leave it alone after it's generated.

PR #985, opened 2021-03-04
Preserve existing .gitignore file in out directory

sinking-point:
Closes https://github.com/rustwasm/wasm-pack/issues/984 .

Rather than check for an existing .gitignore before creating one, I opted to use `create_new` and check the error. This makes it atomic - it'll probably never matter, but hey.

---

```bash
```

### DIFF
--- a/src/command/utils.rs
+++ b/src/command/utils.rs
@@ -3,6 +3,8 @@
 
 use failure;
 use std::fs;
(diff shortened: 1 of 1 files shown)


andrewhamon:
This is much needed in a monorepo. We _want_ to check in the js files so that typescript can function as normal.

PR #986, opened 2021-03-05
Implements RFC 8, fixing #606

### LINKED ISSUE #606: Implement support for RFC 8, transitive NPM dependencies
Specified upstream [RFC 8](https://github.com/rustwasm/rfcs/blob/master/text/008-npm-dependencies.md) is a way for NPM dependencies to work transitively throughout the crate graph. The first phase of supporting this is [now implemented in wasm-bindgen](https://github.com/rustwasm/wasm-bindgen/pull/1305) and the next phase is in `wasm-pack` itself. Thankfully it shouldn't be too hard!

Currently today `wasm-pack` will write out a `package.json`, but this will overwrite a `package.json` created by `wasm-bindgen`, if one exists. Instead what `wasm-pack` should do is:

* When creating `package.json`, first check to see if one already exists at the target location
* Deserialize the `dependencies` map, if it exists, otherwise start with an empty object
* Add in all the current configuration (name, scope, urls, etc)
* Write out the `package.json` again, overwriting what was previously there.

Basically the `package.json` that `wasm-pack` writes should just be merged with the `package.json` emitted by `wasm-bindgen` if it's already there.

jpgneves:

```bash
```

Fixes #606, test included.


simlay:
I've pulled this branch and tested it. It looks correct. @jpgneves could you pull in the recent changes from master?

jpgneves:
> I've pulled this branch and tested it. It looks correct. @jpgneves could you pull in the recent changes from master?

@simlay Thanks! I just rebased off of master.

jpgneves:
@simlay Not wanting to put undue pressure, is there any chance this can be looked at? Thanks!

src/command/utils.rs
```diff
@@ -37,6 +37,7 @@ fn find_manifest_from_cwd() -> Result<PathBuf, failure::Error> {
 
 /// Construct our `pkg` directory in the crate.
 pub fn create_pkg_dir(out_dir: &Path) -> Result<(), failure::Error> {
+    let _ = fs::remove_dir_all(&out_dir); // Clean up any existing directory and ignore errors
```
simlay, line 40: Hmm. It's unclear if we need to remove the entire out directory. It may not be a big deal but say someone has some `npm link` going on pointed at the `pkg` directory, this may break it on every rebuild. I'm not sure if `npm link` does a symlink or a hard link (or is that not a thing with directories) which I think would be the difference.

simlay (approved):
I think this is good minus my one concern about removing the `out_dir` on every build.

@drager is the one who actually has merge/publish rights.

src/command/utils.rs
jpgneves: Yeah, I can see your point. `npm link` [documentation claims they are symlinks](https://docs.npmjs.com/cli/v7/commands/npm-link).

That said, I don't feel strongly about this. I personally tend to want my tests to run on a as clean as possible environment to avoid spurious failures due to other tests leaving cruft behind, but I'm happy to remove this cleanup if that makes sense to you.

simlay:
@drager bump.

drager (approved):
Thank you!

src/command/utils.rs
huntc: We're seeing errors that I think are in relation to this as reported at https://github.com/rustwasm/wasm-pack/issues/1099#issuecomment-1033430935

huntc: I've raised a PR to revert this change: https://github.com/rustwasm/wasm-pack/pull/1110

callensm:
Is there a timeline for when this RFC fix is going to be released? NPM dependencies still seem to be broken in `v0.10.2`.

Issue #952, opened 2021-01-04
Add check for Apple Silicon Macs


Figments:
## 💡 Feature description
Include support for Apple Silicon Macs, e.g. the late 2020 MacBook Air, late 2020 MacBook Pro 13", and late 2020 Mac mini.

#### 💻 Basic example
Currently, trying to install the `wasm-pack` library from either NPM or Yarn results in the following error:

```
Error: Unsupported platform: Darwin arm64
```

Installing `wasm-pack` via cargo (using the latest Darwin arm64 support included in Rust 1.49) works flawlessly. It's only the NPM and Yarn version that produces this error.

anp:
I'm also getting set up on an M1 machine, although I build my wasm-pack from source so didn't run into this. Based on the [NPM installer](https://github.com/rustwasm/wasm-pack/blob/master/npm/binary.js#L27) I think the project will also need to upload aarch64-apple-darwin binaries to GitHub Releases.

anp:
Also FYI once you get it installed there are still [some limitations](https://github.com/rustwasm/wasm-pack/issues/954), through no fault of wasm-pack's maintainers AIUI.

richiksc:
You mention wasm-pack working flawlessly on M1 if installed via cargo or compiled from source. I run into the issue described at #907 when running wasm-pack compiled from source on an M1-based Mac.

iugo:
```
➜  curl https://rustwasm.github.io/wasm-pack/installer/init.sh -sSf | sh
wasm-pack-init: no precompiled binaries available for CPU architecture: arm64
```

tholu:
@iugo Ran into the same issue, tried `cargo install wasm-pack` instead and it works so far.

Xayaan:
Curl command doesn't work on M1, as @tholu stated above. Running `cargo install wasm-pack` works perfectly.

huntc:
Having followed the `cargo install` route, I note that the resultant `wasm-pack` is built for x86:

```
huntc@MBP14inch2021 ~ % file `which wasm-pack`
/Users/huntc/.cargo/bin/wasm-pack: Mach-O 64-bit executable x86_64
```

fishcharlie:
Is there any plans to fix the curl installer and make the bundle universal or Apple Silicon compatible?

d3lm:
[...]

PR #1000, opened 2021-04-21
fix "no pre-built binaries" error for aarch64, use x86_64 via rosetta

nilslice:
Adds support for installing binary deps (`wasm-opt`, etc) `aarch64-apple-darwin` target, via usage of `x86_64` binary run through the Apple-provided Rosetta instruction translator. As long as _Rosetta 2_ is installed on the target Apple device, all downloaded binaries will be executed using it automatically. 

More info on Rosetta 2: https://support.apple.com/en-us/HT211861

closes: #952

---

### DIFF
--- a/docs/_installer/init.sh
+++ b/docs/_installer/init.sh
@@ -99,6 +99,11 @@ get_architecture() {
 
         Darwin)
             local _ostype=apple-darwin
+            # override for case where target machine is using Apple Silicon CPU, 
+            # use x86_64 binary via rosetta
+            if [ $_cputype == arm64 ]; then
+                local _cputype=x86_64
+            fi
             ;;
 
         MINGW* | MSYS* | CYGWIN*)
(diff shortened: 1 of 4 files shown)


nilslice:
looks like appveyor ci check is failing here: 
```
stderr=```[INFO]: Installing cargo-generate...
error: the `--vers` provided, `latest`, is not a valid semver version: cannot parse 'latest' as a semver
Error: Installing cargo-generate with cargo
Caused by: failed to execute `cargo install`: exited with exit code: 101
  full command: "cargo" "install" "--force" "cargo-generate" "--version" "latest" "--root" "C:\\projects\\wasm-pack-071k0\\target\\test_cache\\.cargo-generate-cargo-install-latest"
  ```

doesn't like `"latest"` as a version, so we may need to pull the semver compliant version.

nilslice:
In a different failing test, an attempt is made to download a pre-built binary at this URL: 
https://github.com/ashleygwilliams/cargo-generate/releases/download/v0.6.1/cargo-generate-v0.6.1-x86_64-pc-windows-msvc.tar (which would be redirected to the new org cargo-generate)

The issue is that releases haven't contained these assets since Sept. 2020 (v0.5.1, https://github.com/cargo-generate/cargo-generate/tags) 

Was there a reason builds were excluded from the releases since then?

PR #973, opened 2021-02-13
Fix #958: Update binary-install in package.json to fix security alert

Rizary:
Make sure these boxes are checked! 📦✅

- [ ] You have the latest version of `rustfmt` installed
```bash
$ rustup component add rustfmt
```
- [ ] You ran `cargo fmt` on the code base before submitting
- [X] You reference which issue is being closed in the PR text

This PR fixes #958.

✨✨ 😄 Thanks so much for contributing to wasm-pack! 😄 ✨✨

PR #983, opened 2021-03-03
Fix cargo-generate installation

bradyjoslin:
Make sure these boxes are checked! 📦✅

- [x] You have the latest version of `rustfmt` installed
```bash
$ rustup component add rustfmt
```
- [x] You ran `cargo fmt` on the code base before submitting
- [x] You reference which issue is being closed in the PR text

✨✨ 😄 Thanks so much for contributing to wasm-pack! 😄 ✨✨

Closes #975

PR #1012, opened 2021-06-03
Binary install update for axios security vulnerability.

### LINKED ISSUE #958: Security alert for axios 0.19.0 (npm dependency)
## 🐛 Bug description
There is a security alert for axios 0.19.0, which is in npm/package-lock.json.

#### 🤔 Expected Behavior
It should be upgraded to 0.21.1.  (Upgrading binary-install to 0.1.0 and then running npm update should fix it.)

#### 👟 Steps to reproduce
Any package that uses wasm-pack has this warning:
image

#### 🌍 Your environment
wasm-pack version: 0.9.1

simlay:
Closes #958. Co-PR to #973.

Based on https://github.com/rustwasm/wasm-pack/pull/973#pullrequestreview-589981499, I've ran `node run.js` and verified that this works.

If we'd like, I'm down to add a test subcrate that's runs `node run.js build` to actually run the full build. Thoughts?

```bash
```

### DIFF
--- a/npm/binary.js
+++ b/npm/binary.js
@@ -25,7 +25,7 @@ const getBinary = () => {
   const author = "rustwasm";
   const name = "wasm-pack";
   const url = `https://github.com/${author}/${name}/releases/download/v${version}/${name}-v${version}-${platform}.tar.gz`;
-  return new Binary(url, { name });
+  return new Binary(name, url );
 };
 
 const run = () => {
--- a/npm/package-lock.json
+++ b/npm/package-lock.json
@@ -5,12 +5,11 @@
   "requires": true,
   "dependencies": {
     "axios": {
-      "version": "0.19.0",
-      "resolved": "https://registry.npmjs.org/axios/-/axios-0.19.0.tgz",
-      "integrity": "sha512-1uvKqKQta3KBxIz14F2v06AEHZ/dIoeKfbTRkK1E5oqjDnuEerLmYTgJB5AiQZHJcljpg1TuRzdjDR06qNk0DQ==",
+      "version": "0.21.1",
+      "resolved": "https://registry.npmjs.org/axios/-/axios-0.21.1.tgz",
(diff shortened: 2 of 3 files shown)


Keavon:
Thanks for this! I will email @ashleygwilliams since hopefully she just has GitHub notifications disabled but hasn't fallen off the face of the earth.

Keavon:
Do we also want to update the version number to `0.9.2`? I believe (some or all of) these are the relevant files?:

* `Cargo.lock:2284`
* `Cargo.toml:4`
* `CHANGELOG.md` (add a section indicating the bumped binary-install version)
* `docs\index.html:45` and `:47`
* `npm\package.json:3`
* `npm\package-lock.json` (but you aren't supposed to hand-edit this file, does this perhaps get updated by running `npm update` or `npm install` after making the other changes?)

Keavon:
Hmm, what about these test failures?

simlay:
> Hmm, what about these test failures?
> 
> 
> 
> 

I'm pretty sure this is fixed in #983.

Keavon:
I see, so we would need to merge that one before this one?

simlay:
> I see, so we would need to merge that one before this one?

There have been a couple of PRs to help clean up CI. I think fixing CI is outside the scope of this security fix.

Keavon:
Ok, so that would mean we just merge even though CI is failing, correct? In other words, CI would fail if it were run on the current state of the `master` branch?

simlay:
> Ok, so that would mean we just merge even though CI is failing, correct? In other words, CI would fail if it were run on the current state of the `master` branch?

It seems that way. The only PR since December to "pass" ci was #983 it seems. If whoever has merge and publish access to this repo wants CI to pass for this PR, I'm fine with fixing it but I'm not gonna look much more into it without someone showing up.

drager (approved):
Thanks a lot!

PR #1032, opened 2021-07-10
fixes docs offscreen warning label 

RileyMShea:
I just put the warning header in a `p` tag so that it wouldn't render offscreen.

There's some extra whitespace in the bottom margin of the warning header but I thought I'd keep the change atomic.

- [ ✅] You have the latest version of `rustfmt` installed
```bash
```

- [ ✅] You ran `cargo fmt` on the code base before submitting
- [ ✅] You reference which issue is being closed in the PR text

### DIFF
--- a/docs/_theme/header.hbs
+++ b/docs/_theme/header.hbs
@@ -35,10 +35,12 @@
   }
 </style>
 <header class='warning'>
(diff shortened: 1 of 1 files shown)


drager (approved):
Thanks!

PR #1035, opened 2021-07-20
Fix example commands in docs broken by PR #851

fkettelhoit:
Since #851 was merged, both `wasm-pack build` and `wasm-pack test` directly pass extra arguments to `cargo build` and `cargo test` without the need for an additional `--`. This PR updates the docs to reflect this change.

EDIT: I just noticed that `wasm-pack build` did not yet use the same syntax as `wasm-pack test`, which would be fixed by #1037 and should be merged first for the doc changes of this PR to make sense.

---

```bash
```

### DIFF
--- a/docs/src/commands/build.md
+++ b/docs/src/commands/build.md
@@ -127,13 +127,13 @@ wasm-pack build examples/js-hello-world --mode no-install
 
 ## Extra options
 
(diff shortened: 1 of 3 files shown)


drager (approved):
Thank you!

PR #1046, opened 2021-08-03
Fix mdbook build

palfrey:
All 1.x versions of [ammonia](https://crates.io/crates/ammonia/versions) have been yanked, so mdbook 0.2 isn't installable anymore. This PR upgrades to 0.3, and runs the build part of mdbook on PRs so we can pick up issues like this before master merge.

---

```bash
```


palfrey:
@drager Not quite sure why the windows build decided to timeout (suspect random unrelated failure), but this should fix the current doc build issue

drager (approved):
Awesome. Thanks a lot!

PR #1054, opened 2021-09-05
fix: todo, minor typos, extra clones

workingjubilee:
While struggling with an issue I reviewed some of wasm-pack's source to understand how it works better. En route, I found minor details I wound up poking at: borrows that are re-borrowed and clones that could be skipped, mostly.


drager (approved):
Thanks, looks great!

PR #1061, opened 2021-09-23
fix: declare ES module in package.json

### LINKED ISSUE #1039: Add type = "module" in package.json to support nodejs in bundler mode
## 💡 Feature description

When you build using the default "bundler" mode, it can *almost* run directly from nodejs, using nodejs's built in ESM support.

Unfortunately, the created package.json file is missing `"type": "module"`. Adding this to the created package.json not only fixes some bundler issues, but also makes nodejs understand the created package too:

foo.js:

```javascript
import {greet} from 'my_module'
greet()
```

```
$ node --experimental-wasm-modules foo.js
(node:414396) ExperimentalWarning: Importing Web Assembly modules is an experimental feature. This feature could change at any time
(Use `node --trace-warnings ...` to show where the warning was created)
Hello world!
```

gthb:
In bundler mode, generate `package.json` with `"type": "module"` and use the `"main"` attribute instead of the `"module"` attribute.

This change makes the built ES module palatable to Node.js (when run with `--experimental-vm-modules --experimental-wasm-modules`), while it remains also palatable to webpack as illustrated in https://github.com/webpack/webpack/pull/14313 (where the `pkg` subfolder is generated with `wasm-pack` built with this change). This resolves the headache of using a wasm-pack-built package in a library that one needs to both run directly in Node and include in a webpack build.

Fixes #1039

```bash
```


josephg:
Ping - I really want this feature merged too. I've settled for writing a post-processing script which inserts `"type": "module"` myself, but its quite annoying!

mkermani144:
I don't know if this is going to be merged that soon, but isn't it better if we use the new `exports` field of `package.json` instead?

josephg:
I think [`exports`](https://nodejs.org/api/packages.html#exports) or main would be fine, though I’m curious why you think exports is better @mkermani144 other than it’s newer.

It doesn’t matter to me, just so long as package.json isn’t broken out of the gate. (It’s broken because it claims the module is a cjs module when it’s actually an esm module). I’m disappointed this has been in a bad state so long. What do we need to do to fix this?

mkermani144:
I just think it's more modern than the `main` field. From [Node docs](https://nodejs.org/api/packages.html#package-entry-points):

> For new packages targeting the currently supported versions of Node.js, the "exports"field is recommended. For packages supporting Node.js 10 and below, the "main" field is required.

Anyway, I agree with you. Merging this PR is far more important than using either of the fields 😅

hanayashiki:
when is this out?

hanayashiki:
For myself, this library is not sufficient for package publishing. I use a wrapper typescript file and rollup to publish npm package that is usable out of box. I'm using target --web and i think that is the only way.

kachkaev:
It’d be great to see this PR merged, yes! I’ve crafted a [demo](https://github.com/hasharchives/wasm-ts-esm-in-node-jest-and-nextjs) where I use `--targer=bundler` and then manually patch `package.json`. It confirms that the solution in this PR works in Node.js, Jest and Next.js 🎉

tdelmas:
https://github.com/rustwasm/wasm-pack/issues/1039#issuecomment-1809632849 :

> Yes, need to rebase master into the current https://github.com/rustwasm/wasm-pack/pull/1061 in order to get that merged.

drager (approved):
Thank you!

dukejones:
So stoked about this!! Any idea of when the next release is? Last release looks like Jun 27.

I'm writing a plugin to integrate wasm-pack with Vite & Sveltekit, & I'd love to not have to manually rewrite package.json with "type":"module"

Issue #1064, opened 2021-09-29
Publish pre-built binary for linux/arm64 (aarch64)


nacardin:
#### 💻 Example
I would like to be able to run the install script on arm64 linux machines.
```
$ curl https://rustwasm.github.io/wasm-pack/installer/init.sh -sSf | sh 
```
Currently it fails with
```
wasm-pack-init: no precompiled binaries available for CPU architecture: aarch64
```

#### 💡 Description
* The release workflow should build and publish a binary for `aarch64-unknown-linux-musl`.
* The install script should look up and install pre-built binary for aarch64.

Serveny:
Use `cargo install wasm-pack` instead

d3lm:
Yep, using `cargo install` works fine.

drager:
Fixed in [v0.11.0](https://github.com/rustwasm/wasm-pack/releases/tag/v0.11.0).

PR #1065, opened 2021-09-30
Add linux/arm64 to release workflow

nacardin:
Build and release `aarch64-unknown-linux-musl` target. Solves first part of https://github.com/rustwasm/wasm-pack/issues/1064.

With this change we can easily add more targets in the future. Once GH updates `macos-latest` to using MacOS 11, we can add `aarch64-apple-darwin`.

You can see that the release workflow succeeds here: https://github.com/nacardin/wasm-pack/actions/runs/1291954371

- [ x ] You have the latest version of `rustfmt` installed
- [ x ] You ran `cargo fmt` on the code base before submitting
- [ x ] You reference which issue is being closed in the PR text

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -56,7 +56,7 @@ jobs:
         if: matrix.os == 'ubuntu-latest'
         run: |
           sudo apt-get update -y
-          sudo apt-get install clang -y
+          sudo apt-get install clang libc6-dev-arm64-cross -y
           echo "TARGET_CC=clang" >> $GITHUB_ENV
 
       - name: Install p7zip (MacOS)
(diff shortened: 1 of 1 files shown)


drager (approved):
Thanks!

d3lm:
Yay! Looking forward for a `aarch64` so it runs on an M1.

d3lm:
Couldn't we use `macos-11` for the time being until `macos-latest` gets updated? @nacardin

Sotatek-NguyenNguyen4:
I'm using macos monterey, I try to do step by step on above, and it doesn't work. I'm really not clearly step: "You ran cargo fmt on the code base before submitting". Could anyone explain it for me?

Issue #1076, opened 2021-10-31
Extra options `--target-dir` build error


utherpally:
## 🐛 Bug description
Run wasm-pack with `--target-dir` extra-options
```
wasm-pack build frontend --target web --no-typescript --dev -- --target-dir target/frontend
```
Raise an error:
```
error: failed to read `/home/dev/Projects/maple/target/wasm32-unknown-unknown/debug/frontend.wasm`

Caused by:
    0: failed to read `/home/dev/Projects/maple/target/wasm32-unknown-unknown/debug/frontend.wasm`
    1: No such file or directory (os error 2)
Error: Running the wasm-bindgen CLI
Caused by: failed to execute `wasm-bindgen`: exited with exit status: 1
  full command: "/home/dev/.cargo/bin/wasm-bindgen" "/home/dev/Projects/maple/target/wasm32-unknown-unknown/debug/frontend.wasm" "--out-dir" "frontend/pkg" "--no-typescript" "--target" "web" "--debug"
```

#### 🌍 Your environment
Include the relevant details of your environment.
wasm-pack version: 0.10.1
rustc version: 1.58.0-nightly (514b38779 2021-10-22)

drager:
Fixed in #1082 and released in [v0.11.0](https://github.com/rustwasm/wasm-pack/releases/tag/v0.11.0).

PR #1082, opened 2021-11-17
fix(#1076): `--target-dir` as extra option is now considered as expected

sassman:

```bash
```


wdanilo:
@drager - do you think we can merge it? It would save a lot of debugging on our end this week :(

wdanilo:
Thanks so so so much @drager !

PR #1086, opened 2021-12-02
Replace two mentions of `wasm-pack init` with `wasm-pack build` in the docs

mstange:
Two small documentation fixes.


maccesch (approved):

drager (approved):
Thank you!

PR #1089, opened 2021-12-05
Recent node versions can import ESModulesPackage output if "type": "module" is specified

serprex:
I accidentally closed #936 while rebasing/updating


gthb:
See https://github.com/rustwasm/wasm-pack/pull/1061 — is this not the same thing?

serprex:
@gthb yes

Issue #913, opened 2020-09-07
no prebuilt wasm-opt binaries error


JHBonarius:
## 🐛 Bug description
I'm running wasm-pack on a raspberry pie 3b (arm v71). and are trying the game of life tutorial. Build 'fails' with an error. (I say 'fails', as I get something that works after all.)

`~/rust/wasm-game-of-life $ wasm-pack build
[INFO]: Checking for the Wasm target...
[INFO]: Compiling to Wasm...
    Finished release [optimized] target(s) in 0.50s
[INFO]: Installing wasm-bindgen...
Error: no prebuilt wasm-opt binaries are available for this platform: Unrecognized target!
To disable `wasm-opt`, add `wasm-opt = false` to your package metadata in your 'Cargo.toml'.`

#### 🤔 Expected Behavior
Looking at `wasm-opt.rs`, I would have expected the build to successfully complete, giving the message "Skipping wasm-opt because it is not supported on this platform".
It also doesn't seem to look for a local wasm-opt in the path, as described above `pub fn find_wasm_opt`. I downloaded the [wasm-opt source](https://github.com/WebAssembly/binaryen), build it locally and added the bin dir it to the path... source doesn't show any signs of looking at the path...

#### 👟 Steps to reproduce
Get a raspberry pie, install raspbian OS, install git, rust, wasm, etc
git clone https://github.com/rustwasm/wasm_game_of_life.git
cd wasm_game_of_life/
wasm_pack build

#### 🌍 Your environment
Include the relevant details of your environment.
wasm-pack version: 0.9.1
rustc version: 1.46.0

PiepsC:
I ran into a completely different issue with wasm-opt as well which was solved by downgrading to 0.8.1. I reckon that will not solve anything for you considering the lack of binaries.

Have you tried disabling wasm-opt altogether? `To disable `wasm-opt`, add `wasm-opt = false` to your package metadata in your `Cargo.toml`.` Obviously doesn't address the issue here, but perhaps it'll at least allow you to continue.

SebastienGllmt:
I think this is a duplicate of https://github.com/rustwasm/wasm-pack/issues/886

JHBonarius:
> I think this is a duplicate of #886

No, because that bug occurs when wasm-opt is actually running. In my case, the wasm-opt binaries are not found at all.

AntoninPe:
Same issue here with a raspberry-pi 4

everything compiles well but wasm-opt fails.

[...]

PR #1102, opened 2022-01-09
Add aarch64-linux support

Nasafato:
No issue, but related to #913. `wasm-opt` does include pre-built binaries for `aarch64-linux`, so we add a conditional to handle that case. Let me know if you still want me to create an issue


yaleman:
Any chance this can be merged? It's still a problem.

wegylexy:
What about _Windows on ARM_?

github2023spring:
May I know why this pull is closed? I think it is still a problem.

Issue #1099, opened 2021-12-27
Change in v0.10.2 that causes output directory to first be removed breaks my build


zzzachzzz:
In reference to this change which causes the output directory to first be removed: https://github.com/rustwasm/wasm-pack/pull/986#pullrequestreview-715310944

While I understand this may be intended behavior, I thought I'd share my use case that illustrates why my build is dependent on files already residing in the output `pkg` folder, files which are tracked in git.

I have a `.gitignore` that looks like this:
```
pkg/*
!pkg/index.ts
!pkg/insight.worker.ts
!pkg/insight.d.ts
```
As you can see, I keep some essential files tracked by git in the `pkg` folder, and these modules then reference the output files generated by wasm-pack. Maybe there's a better way for me to go about this? Such as copying the files into `pkg` as part of the build process, or perhaps I should keep these files outside of the `pkg` folder, and reference the build output that goes into `pkg`.

If it is decided to keep this change, please add it to the v0.10.2 changelog.

huntc:
We are also seeing this problem introduced with 0.10.2 and in conjunction with webpack. Webpack is watching the files in the pkg folder and seems to go into a panic, and is unable to recover until we restart its watching.

@zzzachzzz Did you manage to overcome this somehow?

zzzachzzz:
@huntc My temporary solution is to lock wasm-pack @ 0.10.1, installing via npm

huntc:
Fixed by my PR.

PR #1110, opened 2022-02-11
Do not remove the pkg directory

huntc:
A recent commit (d79e72189c523586debcee2ebbf6a2fd065e8abf) ensured that the `pkg` directory was removed as the first step of attempting to create it. Unfortunately, this causes a problem for [webpack](https://webpack.js.org/) when watching the `pkg` directory. Webpack is unable to recover its watching and so any watch server must be restarted, which is a blocker when using it.

Please note that there was [some debate](https://github.com/rustwasm/wasm-pack/pull/986#discussion_r676967231) about whether to remove the `pkg` directory. It is unclear to me what the actual benefits of it are, and whether this PR re-introduces some other problem. However, wasmpack appears to have existed as per the behaviour of this PR for a few years.

This PR reverses the change made in the above commit and fixes https://github.com/rustwasm/wasm-pack/issues/1099.

### DIFF
--- a/src/command/utils.rs
+++ b/src/command/utils.rs
@@ -37,7 +37,6 @@ fn find_manifest_from_cwd() -> Result<PathBuf, failure::Error> {
 
 /// Construct our `pkg` directory in the crate.
 pub fn create_pkg_dir(out_dir: &Path) -> Result<(), failure::Error> {
-    let _ = fs::remove_dir_all(&out_dir); // Clean up any existing directory and ignore errors
     fs::create_dir_all(&out_dir)?;
     fs::write(out_dir.join(".gitignore"), "*")?;
     Ok(())


simlay (approved):
Oof. Well, sorry about. Thanks for mentioning the discussion. It's nice to know that it was at least discussed about even if the incorrect choice was made.

@drager Could you merge when you get a chance?

huntc:
Are these test failures mine? I simply removed a line... I couldn't successfully run `cargo test` for perhaps other reasons as I couldn't afford to invest a great deal of time on this. I will look further in terms of getting a good setup though if required.

huntc:
> Are these test failures mine? I simply removed a line... I couldn't successfully run `cargo test` for perhaps other reasons as I couldn't afford to invest a great deal of time on this. I will look further in terms of getting a good setup though if required.

Would it be possible to follow up on the above as wasmpack is broken for us.  Thanks.

drager (approved):

drager:
> > Are these test failures mine? I simply removed a line... I couldn't successfully run `cargo test` for perhaps other reasons as I couldn't afford to invest a great deal of time on this. I will look further in terms of getting a good setup though if required.
> 
> Would it be possible to follow up on the above as wasmpack is broken for us. Thanks.

I think this wasn't this PRs fault. Will merge and investigate why it fails. Thanks for the PR!

huntc:
Thank you, and thanks for your efforts here too!

main--:
This change - while necessary to fix interaction with e.g. the webpack dev server - causes a new problem: Now that wasm-pack doesn't clear the output folder, it will pick up the `package.json` created by a previous run and try to deserialize this as if it was created by wasm-bindgen. This fails and causes wasm-pack to return an error code.

huntc:
> This change - while necessary to fix interaction with e.g. the webpack dev server - causes a new problem: Now that wasm-pack doesn't clear the output folder, it will pick up the `package.json` created by a previous run and try to deserialize this as if it was created by wasm-bindgen. This fails and causes wasm-pack to return an error code.

The change here restores behaviour to what it was, so is the problem you describe really a new issue?

main--:
Yes. Deleting the output folder was introduced along with the changes that implement the package.json generation, so prior to that, the problem I describe did not exist.

huntc:
Perhaps the solution is to just remove the package.json? I don’t know, but I’m thinking a new issue should be raised. Wdyt?

main--:
Simply removing package.json is what I'm doing now: https://github.com/main--/wasm-pack/commit/13369d26ed1668d04f9e521d731294791bb5747e

And it does resolve the problem for me.

huntc:
> 

Do you know if this upsets webpack again though as per https://github.com/rustwasm/wasm-pack/pull/1110#issue-1131781769?

main--:
At least in my setup, this makes webpack even happier than before. Before, I was (on Windows) just launching an instance of cmd in the pkg directory to prevent it from getting deleted - this still made webpack sad because it would see files getting deleted, rebuild (with an error) and then build *again* once the files are back. Now that the files are overwritten directly, it only builds once (and also much faster, presumably because caching works better). My webpack doesn't care about the package.json at all.

huntc:
> At least in my setup, this makes webpack even happier than before. Before, I was (on Windows) just launching an instance of cmd in the pkg directory to prevent it from getting deleted - this still made webpack sad because it would see files getting deleted, rebuild (with an error) and then build _again_ once the files are back. Now that the files are overwritten directly, it only builds once (and also much faster, presumably because caching works better). My webpack doesn't care about the package.json at all.

Thanks for following up. Fingers crossed then!

drager:
Not sure I follow correctly but will this PR when released solve your problem @main-- ?

egfx-notifications:
Just opened an issue (https://github.com/rustwasm/wasm-pack/issues/1118) regarding this and arrived at the same conclusion as @main-- during my investigation.
https://github.com/main--/wasm-pack/commit/13369d26ed1668d04f9e521d731294791bb5747e would fix the issue introduced by this PR.
@main--  Could you make a PR with this commit?

Another option would be to rewrite [write_package_json()](https://github.com/rustwasm/wasm-pack/blob/4ae6306570a0011246c39c8028a4f11a4236f54b/src/manifest/mod.rs#L570) so that it can distinguish between a package.json from wasm-bindgen and a package.json already processed by wasm-pack. I have no idea if and how that would work, though.

PR #1119, opened 2022-03-04
Clean up package.json from previous runs

main--:
Remove the package.json file from previous runs to avoid crashes, as discussed in https://github.com/rustwasm/wasm-pack/pull/1110#issuecomment-1059008962


egfx-notifications (approved):

drager (approved):
Thanks!

PR #1127, opened 2022-04-15
Read WebDriver paths from env at highest priority

### LINKED ISSUE #1126: Headless chrome test fails due to incorrect chromedriver version
EDIT: I fibbed a bit on a couple of points (~~struckout~~ below). Not that it's relevant, but to be clear, chrome is not a snap on Ubuntu, but rather a Debian package (it's chrom*ium* that's a snap). Also, I got to wondering why Ubuntu 20.x wasn't on the latest version of chrome, and it turned out I wasn't actually up-to-date after all (whoops). I'm now on 100.0.4896.60, and everything of course works as expected since the driver and browser major versions match.

So I guess the question is now if wasm-pack wants to support users who, intentionally or otherwise, aren't on the latest version. I would argue that being able to test on arbitrary browser versions, within reason, is a very sensible thing to support.

## 🐛 Bug description
I'm following the [wasm rust tutorial](https://rustwasm.github.io/docs/book/game-of-life/testing.html), which instructs me to run a simple test using `wasm-pack test --chrome --headless`. However, the test fails with:

```
Error: non-200 response code: 404
{"value":{"error":"invalid session id","message":"invalid session id","stacktrace":"#0 0x5581c6dd1ad3 ... (a long list of addresses) }
```

Following #611, I reran with `RUST_LOG=wasm_bindgen_test_runner`, which revealed that the incorrect version of `chromedriver` had been selected:

```
[2022-03-31T21:31:30Z DEBUG wasm_bindgen_test_runner::headless] got: {"sessionId":"ccc87ee615a5b257d818da42ca72618e","status":33,"value":{"message":"session not created: This version of ChromeDriver only supports Chrome version 100\nCurrent browser version is 99.0.4844.84 with binary path /usr/bin/google-chrome\n  (Driver info: chromedriver=100.0.4896.60 (6a5d10861ce8de5fce22564658033b43cb7de047-refs/branch-heads/4896@{#875}),platform=Linux 5.4.0-105-generic x86_64)"}}
```

Following [the docs](https://rustwasm.github.io/wasm-bindgen/wasm-bindgen-test/browsers.html#chromedriverpathtochromedriver), I downloaded the correct version of `chromedriver` and attempted to use `CHROMEDRIVER=path/to/chromedriver`, but it seems to be ignored:

```
[2022-03-31T21:39:00Z DEBUG wasm_bindgen_test_runner::headless] executing "/home/randy/.cache/.wasm-pack/chromedriver-806033fccbd67392/chromedriver" "--port=44385"
```

Possibly I misunderstood if that variable is read by `wasm-pack`, seeing as it's in a section entitled "Appendix: Testing in headless browsers without wasm-pack".

#### 🤔 Expected Behavior
If I understood #611 correctly, the fix (#703) was, per @alexcrichton, "Let's download the most recent stable version." I don't think that's actually the correct behavior, though. Instead, the browser version should be detected and the correct `chromedriver` downloaded as appropriate. ~~I am using the up-to-date chrome snap on Ubuntu 20.04.4 LTS, which is at version 99.0.4844.84 (also seen in logs above).~~

#### 👟 Steps to reproduce
Attempt to run `wasm-pack test --chrome --headless` for any test on a system with chrome 99.0.4844.84 (and probably any pre-100.x.x.x version) and observe that the incorrect `chromedriver` version is used.

#### 🌍 Your environment
wasm-pack version: 0.10.2
rustc version: 1.57.0
~~chrome version: 99.0.4844.84, which as above, is the up-to-date version from the snap on Ubuntu 20.04.4~~

thisisrandy:

- [ x] You have the latest version of `rustfmt` installed
```bash
```
- [ x] You ran `cargo fmt` on the code base before submitting
- [ x] You reference which issue is being closed in the PR text

This closes #1126. A separate PR will be submitted for wasm-bindgen to update the docs.

### DIFF
--- a/src/command/test.rs
+++ b/src/command/test.rs
@@ -10,6 +10,7 @@ use install::{self, InstallMode, Tool};
 use lockfile::Lockfile;
 use log::info;
 use manifest;
+use std::env;
 use std::path::PathBuf;
 use std::str::FromStr;
 use std::time::Instant;
@@ -337,10 +338,12 @@ impl Test {
     fn step_get_chromedriver(&mut self) -> Result<(), Error> {
         assert!(self.chrome && self.chromedriver.is_none());
 
-        self.chromedriver = Some(webdriver::get_or_install_chromedriver(
-            &self.cache,
-            self.mode,
-        )?);
+        self.chromedriver = Some(env::var("CHROMEDRIVER")
+            .map(|path_str| PathBuf::from(path_str))
+            .or_else(|_| webdriver::get_or_install_chromedriver(
+                    &self.cache,
+                    self.mode,
+            ))?);
         Ok(())
     }
 
@@ -362,10 +365,12 @@ impl Test {
     fn step_get_geckodriver(&mut self) -> Result<(), Error> {
         assert!(self.firefox && self.geckodriver.is_none());
 
-        self.geckodriver = Some(webdriver::get_or_install_geckodriver(
-            &self.cache,
-            self.mode,
-        )?);
+        self.geckodriver = Some(env::var("GECKODRIVER")
+            .map(|path_str| PathBuf::from(path_str))
+            .or_else(|_| webdriver::get_or_install_geckodriver(
+                    &self.cache,
+                    self.mode,
+            ))?);
         Ok(())
     }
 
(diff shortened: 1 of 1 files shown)


chris13524:
Thank you for this contribution! Your branch solves my use case

thisisrandy:
The failing test seems to be failing in `rustwasm:master` and unrelated to this PR.

PR #1130, opened 2022-04-21
chore: updates binary-install to v1.0.1

EverlastingBugstopper:
also removes `run.js` and `uninstall.js` as they are not needed


drager (approved):
Thanks a lot!

Issue #1009, opened 2021-05-26
package.json should include an explicit "main" entry


mstange:
This was previously discussed in #94: It's good practice to include a `"main": "main-file.js"` entry in package.json, so that the generated npm package is usable without specifying a file name.

The current workaround is to use `wasm-pack build --out-name index`, or to manually adjust package.json after the build. However, even with the main file being called "index.js", it would be better to have an explicit `"main": "index.js"` entry, according to [this comment](https://github.com/mstange/gecko-profiler-demangle/issues/1#issuecomment-848996652) by @ljharb, who is on the nodejs Modules team.

ljharb:
As long as you're making changes, it'd be also good to add:
```js
"exports": {
  ".": "./index.js",
  "./package.json": "./package.json"
},
```
since adding that field is usually a breaking change, so it's good to add it earlier

DoctorBracewell:
Running into this exact issue - one of the build tools I'm using to bundle the output package uses node's `require.resolve(packageName)` function which complains that it can't find the package if the `package.json` doesn't contain a `main` field.

I might take a look into submitting a PR that adds an explicit `"main": "file-name.js"` field to the `package.json` - can anyone think if that would be a breaking change or have any unintended side-effects?

DoctorBracewell:
Taking a look at the `package.json` generation code and output for each build target it seems like the `nodejs` target is the only one that includes a `main` field, other targets such as `bundler` or `no-modules` use properties such as `module` or `browser`, which according to [this SO answer](https://stackoverflow.com/a/42817320/10509335) are not officially supported by node (but still recognised by a number of different bunders/package managers etc.)

Given Ijharb's comment, I'd assume it best to just include a `main` property for the `package.json` regardless of the build target?

ljharb:
Yes, always include "main" as CJS; there's no advantage to omitting it.

DogLooksGood:
Hey, any news. I saw the PR is blocked?
The `main` property is kinda necessary for using with bundlers like Vite.

ronsims2:
@DogLooksGood, I ran into the same problem and I wrote a couple of scripts to work around the missing main by adding it: https://gist.github.com/ronsims2/78ed52583f84056969577286b881c0bc

DogLooksGood:
@ronsims2 Yeah, we are using a patch to add a "main" entry

leighmcculloch:
[...]

PR #1151, opened 2022-06-20
Updated package.json structs and populated new main field

DoctorBracewell:
#1009 discusses how the generated `package.json` should have a `main` entry regardless of target type, this has been confirmed by a member of the NodeJS module team.

Added `main` field to  NoModulesPackage and ESModulesPackage in the generated package.json.

Cloned the entry filename during struct instantiation for when the `package.json` is generated. Cloning might not be optimal, it is probably possible to borrow the `String` and generate the filename with that, but I think clones here should be fine.

- [✅ ] You have the latest version of `rustfmt` installed
```bash
```
- [✅ ] You ran `cargo fmt` on the code base before submitting
- [✅ ] You reference which issue is being closed in the PR text


zensh:
Look forward to merge, `module` field is not work in node.js.

PR #1176, opened 2022-09-10
Fix invalid binaryen URL

frewsxcv:
Fixes this bug:

```
Error: failed to download from https://github.com/WebAssembly/binaryen/releases/download/version_108/binaryen-version_108-x86-linux.tar.gz
```

For reference: https://github.com/WebAssembly/binaryen/releases/tag/version_108


serprex:
#1163 is a wider fix, but like this PR it also hasn't found how to fix CI for macos

PR #1188, opened 2022-10-30
Fix binaryen URL and use updated `binary-install` to fix installation on macOS (adding to @matheus23's PR)

printfn:
This includes all the commits from #1163, uses the updated `binary-install` crate (https://github.com/rustwasm/binary-install/pull/21), switches from `failure` to `anyhow` to match what `binary-install` uses, and fixes `wasm-opt` installation on macOS.

Close #1163
Close #1175
Close #1176

---

```bash
```


printfn:
Hi @drager, apologies for the ping. Do you think you might potentially be able to take a look at this PR? It’s gotten a bit large unfortunately due to the error handling changes, but I believe it should work now. Thanks!

src/bindgen.rs
```diff
@@ -19,7 +19,7 @@ pub fn wasm_bindgen_build(
     disable_dts: bool,
     target: Target,
     profile: BuildProfile,
-) -> Result<(), failure::Error> {
```
joshuataylor, line 22: This has changed in the main branch, need to add `extra_options`.:
```rust
    extra_options: &Vec<String>,
) -> Result<()> {
```

src/install/mod.rs
```diff
@@ -199,15 +194,14 @@ fn prebuilt_url(tool: &Tool, version: &str) -> Result<String, failure::Error> {
...
-                "0.5.1", // latest released binary [#907](https://github.com/rustwasm/wasm-pack/issues/907)
+                "0.17.3",
                 target
             ))
         },
         Tool::WasmOpt => {
             Ok(format!(
         "https://github.com/WebAssembly/binaryen/releases/download/{vers}/binaryen-{vers}-{target}.tar.gz",
-        vers = "version_108",
+        vers = "version_110",
```
joshuataylor, line 204: 111 is out, looks like not a big change. It would be great to bump to that, and hopefully no more past that and this is merged soon :crossed_fingers:.

https://github.com/WebAssembly/binaryen/releases/tag/version_111
https://github.com/WebAssembly/binaryen/releases/tag/version_111

joshuataylor (commented):
This is great, I'll do more testing on x86 linux and macos arm64.

It no longer errors for me though :-).

src/bindgen.rs
printfn: fixed

src/install/mod.rs
printfn: fixed

joshuataylor:
I've been testing this on both MacOS Monterey ARM64 and Linux AMD64 and it works great! No issues so far. :+1:

Cargo.toml
```diff
@@ -10,38 +10,38 @@ categories = ["wasm"]
...
-siphasher = "0.2.3"
+strsim = "0.10"
+siphasher = "0.3"
 structopt = "0.3"
-toml = "0.4"
-which = "2.0.0"
-binary-install = "0.0.2"
+toml = "0.5"
+which = "4.3.0"
+binary-install = { git = "https://github.com/rustwasm/binary-install", rev = "6d60f6731c8294000ea8f3a0260659b025cf7f94" }
```
drager, line 36: Just published [0.1.0](https://crates.io/crates/binary-install/0.1.0) of binary-install now, so we can use that now :)

drager (approved):
I think this looks good. Thanks a lot for all the work and effort!

printfn:
Updated to use the new version of binary-install :)

PR #1195, opened 2022-11-28
Retry CI from where it last succeeded on rustwasm/master

gthb:
Empty commit to retry CI tests on exactly the state of the `master` branch on which they last succeeded.

```bash
```


gthb:
This was me starting to troubleshoot `wasm-pack test --headless --firefox` failing in our own workflow, but the cause of that turned out to be the `ubuntu-latest` runner alias, which changed to ubuntu 22.04 in which Firefox ceased to be bundled — which is being fixed in the runner, see https://github.com/actions/runner-images/issues/5490 — so that problem (a) is not relevant in this repo, and (b) is going away.

PR #1208, opened 2023-01-11
Tell bundler that the generated package has side effects. Closes #972.

### LINKED ISSUE #972: There should be a way to disable `sideEffects: false`
## 🐛 Bug description

Custom JS snippets can contain top-level side-effects. Right now, wasm-pack generated package.json includes `sideEffects: false` which marks every file as side-effect free, leading to potential bugs after bundling.

#### 🤔 Expected Behavior

When JS snippet files are generated, they should be put into the `sideEffects` array, and only the JS file generated by wasm-bindgen should be assumed to be side-effect free since it's the only one wasm-bindgen has control over.

`package.json` allows to list potentially side-effect-y files as

```json
{
  "sideEffects": [
    "path/to/snippet.js"
  ]
}
```

This will avoid bugs for snippets with top-level side-effects, while it won't decrease optimization level for common scenarios where the module is used via its main entry point.

stephanemagnenat:
This is a brutal fix for #972, that sets `sideEffects` to true for ES modules so that the Rust main function (annotated `#[wasm_bindgen(start)]`) is not tree-shaked. A better solution would be to detect if the main function is present in the Rust code and emit the `sideEffects` flag accordingly.

### DIFF
--- a/src/manifest/mod.rs
+++ b/src/manifest/mod.rs
@@ -737,7 +737,7 @@ impl CrateData {
             module: data.main,
             homepage: data.homepage,
             types: data.dts_file,
-            side_effects: false,
+            side_effects: true,
             keywords: data.keywords,
             dependencies,
         })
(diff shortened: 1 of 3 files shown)


drager:
Fixed in #1224

PR #1185, opened 2022-10-26
Add support for workspace inheritance

printfn:
This PR adds support for reading metadata from crates using the [workspace inheritance](https://doc.rust-lang.org/cargo/reference/workspaces.html#the-package-table) feature introduced in [Rust 1.64](https://blog.rust-lang.org/2022/09/22/Rust-1.64.0.html#cargo-improvements-workspace-inheritance-and-multi-target-builds).

I've implemented this by bumping the `cargo_metadata` dependency to the latest version, and using its metadata for the `description`, `license`, `license-file`, `repository` and `homepage` fields, instead of parsing the `Cargo.toml` file directly. This should also make `wasm-pack` more resilient to future `Cargo.toml` format changes.

I've tested the change locally and it seems to work perfectly.

Fixes #1180.

---

Make sure these boxes are checked! 📦✅

- [x] You have the latest version of `rustfmt` installed
```bash
$ rustup component add rustfmt
```
- [x] You ran `cargo fmt` on the code base before submitting
- [x] You reference which issue is being closed in the PR text

✨✨ 😄 Thanks so much for contributing to wasm-pack! 😄 ✨✨

PR #1216, opened 2023-01-21
Respect package.readme in Cargo.toml

### LINKED ISSUE #1215: Warning about no README when `readme = false`
## 🐛 Bug description
I have some crates nested as subdirectories, one of which is the crate I'm using with _wasm-pack_. When building it, it reports:

```text
[WARN]: :-) origin crate has no README
```

It similarly warned about having no license until I changed my _Cargo.toml_ to the following (relevant to both points above):

```toml
[package]
readme = false
license-file = "../../LICENSE.txt"
```

This stopped the complaints about the missing license, at least, which I have only in my repo/workspace root.

#### 🤔 Expected Behavior
While not a critical problem, it would be nice if the Cargo-standard setting of `package.readme = false` meant that _wasm-pack_ didn't try looking for it.

#### 👟 Steps to reproduce
My repo (simplified) is set up like this:

```text
README.md
LICENSE.txt
Cargo.toml <- workspace.members = [ "crates/foo", "crates/bar" ]
crates/
  foo/
    Cargo.toml <- package.readme = false, package.license-file = "../../LICENSE.txt"
  bar/ <- the wasm project
    Cargo.toml <- package.readme = false, package.license-file = "../../LICENSE.txt"
```

#### 🌍 Your environment
wasm-pack version: 0.10.3
rustc version: 1.66.1 (90743e729 2023-01-10)

heaths:
Fixes #1215 by honoring how package.readme is authored in Cargo.toml, if authored. This could be a path or a boolean to disable searching for the path.

Also fixes the existing `it_copies_a_readme_provided_path` test which was an exact copy - just with fewer empty lines - of `it_copies_a_readme_default_path` and wasn't honoring any authored path since there was no way to get the authored path.

### DIFF
--- a/src/command/build.rs
+++ b/src/command/build.rs
@@ -383,7 +383,7 @@ impl Build {
 
     fn step_copy_readme(&mut self) -> Result<()> {
         info!("Copying readme from crate...");
-        readme::copy_from_crate(&self.crate_path, &self.out_dir)?;
+        readme::copy_from_crate(&self.crate_data, &self.crate_path, &self.out_dir)?;
         info!("Copied readme from crate to {:#?}.", &self.out_dir);
         Ok(())
     }
--- a/src/manifest/mod.rs
+++ b/src/manifest/mod.rs
@@ -529,6 +529,14 @@ impl CrateData {
         }
     }
 
+    /// Gets the optional path to the readme, or None if disabled.
+    pub fn crate_readme(&self) -> Option<String> {
+        self.pkg()
+            .readme
+            .clone()
+            .map(|readme_file| readme_file.into_string())
+    }
+
     /// Get the license for the crate at the given path.
     pub fn crate_license(&self) -> &Option<String> {
         &self.pkg().license
--- a/src/readme.rs
+++ b/src/readme.rs
@@ -4,10 +4,11 @@ use anyhow::{Context, Result};
 use std::fs;
 use std::path::Path;
 
+use crate::manifest::CrateData;
 use crate::PBAR;
 
 /// Copy the crate's README into the `pkg` directory.
-pub fn copy_from_crate(path: &Path, out_dir: &Path) -> Result<()> {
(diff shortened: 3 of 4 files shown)


drager (approved):
Thanks! Looks good to me! Can you rebase latest master into your branch perhaps?

heaths:
> Thanks! Looks good to me! Can you rebase latest master into your branch perhaps?

Unfortunately, because #1185 took a dependency on a new crate, the problem needs to fix there as well. It has the same problem as this crate did where it expects a readme to be a path only (present or absent) so I can't tell the difference between explicitly disabled or just not specified.

heaths:
Ready. I misunderstood that cargo_metadata was consuming from the command `cargo metadata`, which results in `null` in this case so we can still detect when `readme = false` was specified.

heaths:
Looks like the test failures started in or around 3073db8f18fcd452404929d22762389b50d2cfe7

drager:
> Ready. I misunderstood that cargo_metadata was consuming from the command `cargo metadata`, which results in `null` in this case so we can still detect when `readme = false` was specified.

Great, thanks a lot! Appreceite you fixing it!

> Looks like the test failures started in or around [3073db8](https://github.com/rustwasm/wasm-pack/commit/3073db8f18fcd452404929d22762389b50d2cfe7)

Yeah, I will look into the failures. Thanks

drager:
Sorry @heaths. I had to revert this PR since we got some actual test failures after this was merged. Both locally and in Github actions. Could you perhaps take a look at that and submit a new PR? Sorry for the inconvenience...

heaths:
Understood. Will take a look.

heaths:
To note, the two failing log_level tests started earlier in 3073db8f18fcd452404929d22762389b50d2cfe7 - sometime around when the deserialization of the manifest was removed from this crate in lieu of running a tool.

heaths:
@drager please see https://github.com/rustwasm/wasm-pack/pull/1298

drager:
Thank you for taking the time to investigate and fix it! Appreciate it!

Issue #650, opened 2019-05-19
Segmentation fault after small patch


derekdreery:
## 🐛 Bug description
I'm getting a segfault when I run `wasm-pack` with no arguments. I've made some [small non-unsafe changes to the codebase](https://github.com/derekdreery/wasm-pack/tree/custom-binary).

Sometimes, instead of a segfault, I get a failed assertion in curl-sys:

```
thread '<unnamed>' panicked at 'assertion failed: `(left == right)`
  left: `2`,
 right: `0`', **/home/.cargo/registry/src/github.com-1ecc6299db9ec823/curl-0.4.20/src/lib.rs:92:13**
```
which is here: [curl assert fail](https://github.com/alexcrichton/curl-rust/blob/0.4.20/src/lib.rs#L92).

Here is the stack trace from the segfault core dump

```
May 19 15:44:30 systemd-coredump[11016]: Process 11013 (wasm-pack) of user 1000 dumped core.

Stack trace of thread 11014:
#0  0x00007fb5df46836d __pthread_rwlock_unlock (libpthread.so.0)
#1  0x00007fb5df17e3ba CRYPTO_THREAD_unlock (libcrypto.so.1.1)
#2  0x00007fb5df0f1e07 n/a (libcrypto.so.1.1)
#3  0x00007fb5df0f21c8 ERR_load_strings_const (libcrypto.so.1.1)
#4  0x00007fb5df679aea ERR_load_SSL_strings (libssl.so.1.1)
#5  0x00007fb5df679b1a n/a (libssl.so.1.1)
#6  0x00007fb5df46a5cf __pthread_once_slow (libpthread.so.0)
#7  0x00007fb5df17e40a CRYPTO_THREAD_run_once (libcrypto.so.1.1)
#8  0x00007fb5df679f24 OPENSSL_init_ssl (libssl.so.1.1)
#9  0x0000560063ae2fff n/a (/home/devel/non-work/wasm-pack/target/debug/wasm-pack
```

#### 👟 Steps to reproduce
1. Clone the repo above
2. run `wasm-pack` (e.g. `cargo run`)

#### 🌍 Your environment
wasm-pack version: latest git
rustc version: 1.34.2

derekdreery:
Update: I've updated my system, rebooted my computer and done a full rebuild. The segfault has changed slightly: now I occasionally get a message like:

```
double free or corruption (out)
```

and the stacktrace has changed

```
[...]

PR #1224, opened 2023-02-05
Mark snippets and the bundler target's main file as having side effects

### LINKED ISSUE #972: There should be a way to disable `sideEffects: false`

Liamolucko:
This is a less extreme version of #1208, which only marks snippets and the main file on the bundler target as having side effects instead of all files.

This means that the JS shim file which contains the vast majority of the JS code is still properly marked as having no side effects, allowing bundlers to get rid of things like unused `new TextEncoder` calls which could theoretically have side effects but don't.

Fixes #972
Fixes rustwasm/wasm-bindgen#3276

### DIFF
--- a/src/manifest/mod.rs
+++ b/src/manifest/mod.rs
@@ -723,10 +723,10 @@ impl CrateData {
                 url: repo_url,
             }),
             files: data.files,
-            module: data.main,
+            module: data.main.clone(),
             homepage: data.homepage,
             types: data.dts_file,
-            side_effects: false,
+            side_effects: vec![format!("./{}", data.main), "./snippets/*".to_owned()],
             keywords: data.keywords,
             dependencies,
         })
@@ -758,7 +758,7 @@ impl CrateData {
             module: data.main,
             homepage: data.homepage,
             types: data.dts_file,
-            side_effects: false,
+            side_effects: vec!["./snippets/*".to_owned()],
             keywords: data.keywords,
             dependencies,
         })
--- a/src/manifest/npm/esmodules.rs
+++ b/src/manifest/npm/esmodules.rs
@@ -22,7 +22,7 @@ pub struct ESModulesPackage {
     #[serde(skip_serializing_if = "Option::is_none")]
     pub types: Option<String>,
     #[serde(rename = "sideEffects")]
-    pub side_effects: bool,
+    pub side_effects: Vec<String>,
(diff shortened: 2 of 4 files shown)


cryptoquick:
I tried testing out your branch, but unfortunately I ran into this issue here: https://github.com/rustwasm/wasm-pack/issues/650#issuecomment-1436057987

It's unrelated to what's solved here, but if I can solve that, I can test if this resolves https://github.com/rustwasm/wasm-bindgen/issues/3306 and https://github.com/rustwasm/wasm-pack/issues/1230.

Menci:
I think only `bundler` target's `sideEffects` should be set to the entry file. `web` target should keep `false`.

Liamolucko:
> I think only `bundler` target's `sideEffects` should be set to the entry file. `web` target should keep `false`.

It's already effectively `false` for the entry file. On the web target, `sideEffects` is set to `["./snippets/*"]`, which means that only the files inside of `snippets` are treated as though they have side effects. Everything else, including the entry file, is treated as though `sideEffects` is set to false.

The reason why snippets are marked as having side effects is because they're arbitrary user code; we've got no idea what they do, and we don't want to assume that they have no side effects and cause them to break.

martinxyz:
I've tested this branch (after a merge with master) and it fixed the issue for me.

Sytten:
What's missing to move this forward? It's an annoying bug right now

Liamolucko:
> What's missing to move this forward? It's an annoying bug right now

It's just waiting on a review from @drager.

drager (approved):
Thanks a lot!

cryptoquick:
Thanks! 🙏

PR #1237, opened 2023-02-26
fix(manifest/npm): set correct `sideEffects` for bundler target

### LINKED ISSUE #1235: `sideEffects` should not be `false` now
## 🐛 Bug description
To enable tree-shaking of bundlers, wasm-pack generates `package.json` with `sideEffects: false`. It's OK before since the entry module generated by wasm-bindgen didn't have side effects:

```js
import * as wasm from "./{wasm_name}.wasm";
export * from "./{js_name}.js";
```

But after https://github.com/rustwasm/wasm-bindgen/commit/721c86c3467750bf2a46bf85941e6c211ae9e87a, it now has side effects:

```js
import * as wasm from "./{wasm_name}.wasm";
import { __wbg_set_wasm } from "./{js_name}.js";
__wbg_set_wasm(wasm);
export * from "./{js_name}.js";
```

Now `sideEffects: false` will let bundle remove the `__wbg_set_wasm` call from the entry module, which causes the `let wasm;` variable in binding module to be `undefined`.

#### 🤔 Expected Behavior
What should have happened?

The `sideEffects` property should be set to a list of files with side effects. I think it should be the entry module's filename, i.e. `sideEffects: ["./{entry_name}.js"]`.

#### 👟 Steps to reproduce
Clear steps describing how to reproduce the issue, including commands and flags run. If you are seeing an error, please include the full error message and stack trace.

* [rustwasm-bug-repro.zip](https://github.com/rustwasm/wasm-pack/files/10831720/rustwasm-bug-repro.zip)
* Unzip the project, run `yarn` to install dependencies and run `yarn build` to build a bundle.
* Open `dist/assets/index-*.js`. The file is not minified. You can see the `__wbg_set_wasm` function and assignment to `let wasm;` variable is tree-shaked.
* Modify `node_modules/pkg/package.json` (not `pkg/package.json`, the one to be used by Vite is in node_modules), set `sideEffects` to `["./game_of_life.js"]`.
* Rerun `yarn build`
* Open `dist/assets/index-*.js`. The file is not minified. You can see the `__wbg_set_wasm` function is present and called correctly.

The `pkg` directory in the archive is generated from https://github.com/BLaZeKiLL/game-of-life/tree/69443864176a39415c93c82d361caf2393126b80. With wasm-pack 0.10.3.

#### 🌍 Your environment
wasm-pack version: 0.10.3
rustc version: 1.64.0

Menci:

```bash
```

Fix #1235. Set `sideEffects` to the entry file for `bundler` target while keep `false` for `web` target.

### DIFF
--- a/src/manifest/mod.rs
+++ b/src/manifest/mod.rs
@@ -14,6 +14,7 @@ use std::{collections::HashMap, fs};
 
 use self::npm::{
     repository::Repository, CommonJSPackage, ESModulesPackage, NoModulesPackage, NpmPackage,
+    SideEffects,
 };
 use crate::command::build::{BuildProfile, Target};
 use crate::PBAR;
@@ -723,10 +724,10 @@ impl CrateData {
                 url: repo_url,
             }),
             files: data.files,
-            module: data.main,
+            module: data.main.clone(),
             homepage: data.homepage,
             types: data.dts_file,
-            side_effects: false,
+            side_effects: SideEffects::StringVec([data.main].to_vec()),
             keywords: data.keywords,
             dependencies,
         })
@@ -758,7 +759,7 @@ impl CrateData {
             module: data.main,
             homepage: data.homepage,
             types: data.dts_file,
-            side_effects: false,
+            side_effects: SideEffects::Bool(false),
             keywords: data.keywords,
             dependencies,
         })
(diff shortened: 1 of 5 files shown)


Liamolucko:
#1224 already does the same thing as this PR.

PR #1250, opened 2023-03-22
fix: wasm-pack bin script entry

### LINKED ISSUE #1248: Unable to find `wasm-pack` on 0.11.0
## 🐛 Bug description
After installing `wasm-pack@0.11.0` on Ubuntu 22.04, running `wasm-pack` fails. The issue does not occur using `wasm-pack@0.10.3` in the same manner. The issue occurs both locally and on the `ubuntu-latest` GitHub Actions runner.

#### 🤔 Expected Behavior
The `wasm-pack` command should run. Instead, the system reports that the command is not found.

#### 👟 Steps to reproduce
Using an Ubuntu 22.04 installation, run `npm install wasm-pack@0.11.0`, which is currently the latest version. Then, run `wasm-pack`. The system reports that `wasm-pack` cannot be found.

To show this is a new issue, run `npm install wasm-pack@0.10.3` and then run `wasm-pack`. The command succeeds. 

#### 🌍 Your environment
wasm-pack version: 0.11.0
rustc version: 1.68.0

ahippler:
A bin entry in a package.json has to be a local file and not a command to be linked by a package manager.

See https://docs.npmjs.com/cli/v9/configuring-npm/package-json#bin
> To use this, supply a bin field in your package.json which is a map of command name to local file name.

this fixes #1248

### DIFF
--- a/npm/package.json
+++ b/npm/package.json
@@ -7,7 +7,7 @@
     "postinstall": "node ./install.js"
   },
   "bin": {
-    "wasm-pack": "node ./run.js"
+    "wasm-pack": "./run.js"
   },
   "repository": {
     "type": "git",


ahippler:
@drager could you please review and make a new release?

drager (approved):
Thanks!

PR #1255, opened 2023-03-25
Use correct path name for locally-installed wasm-opt.

jamesyoungman:
If wasm-opt is installed in `/usr/bin/wasm-opt`, the previous code tried to execute `/usr/bin/bin/wasm-opt` which of course doesn't exist.

```bash
```


jamesyoungman:
This is the wrong fix, because it fails to work for the downloaded case.    Better fix in another PR.

PR #1272, opened 2023-04-14
Add wasmbindgen option: `omit_default_module_path`

matthiasgeihs:
There is currently no option to run `wasm-bindgen` with `--omit_default_module_path`.

This is relevant for projects that run in a context where usage of `input = new URL('file.wasm', import.meta.url);` is [causing problems](https://github.com/vitejs/vite/issues/5087) (e.g., in web workers).

This PR adds the option to run `wasm-bindgen` with `--omit_default_module_path`.
Reference: https://rustwasm.github.io/wasm-bindgen/reference/cli.html?highlight=omit#--omit-default-module-path

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1031,6 +1031,8 @@
     demangle-name-section = true
     # Should we emit the DWARF debug info custom sections?
     dwarf-debug-info = false
+    # Should we omit the default import path?
+    omit-default-module-path = false
     ```
 
     As always- there are defaults for you to use, but if you love to configure (or have a project that requires it),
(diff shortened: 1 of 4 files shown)


CHANGELOG.md
```diff
@@ -1031,6 +1031,8 @@
     demangle-name-section = true
     # Should we emit the DWARF debug info custom sections?
     dwarf-debug-info = false
+    # Should we omit the default import path?
+    omit-default-module-path = false
```
drager, line 1035: I don't think the old changelog should be edited.

matthiasgeihs: Oops 🙈 Thanks for catching this one 🙏 Added a revert commit for CHANGELOG.

drager (approved):
Thanks!

PR #1288, opened 2023-05-21
fix: removes `--always-auth` from `npm login`

EstebanBorai:
`npm login` doesn't support `--always-auth` anymore, instead it is under the [`adduser` subcommand][1].

[1]: https://docs.npmjs.com/cli/v6/commands/npm-adduser#always-auth

```bash
```
- [-] You reference which issue is being closed in the PR text


drager (approved):
Thank you!

PR #1298, opened 2023-06-16
Respect package.readme in Cargo.toml

### LINKED ISSUE #1215: Warning about no README when `readme = false`

heaths:
Fixes #1215 by honoring how package.readme is authored in Cargo.toml, if authored. This could be a path or a boolean to disable searching for the path.

Also fixes the existing `it_copies_a_readme_provided_path` test which was an exact copy - just with fewer empty lines - of `it_copies_a_readme_default_path` and wasn't honoring any authored path since there was no way to get the authored path.

### DIFF
--- a/src/command/build.rs
+++ b/src/command/build.rs
@@ -376,7 +376,7 @@ impl Build {
 
     fn step_copy_readme(&mut self) -> Result<()> {
         info!("Copying readme from crate...");
-        readme::copy_from_crate(&self.crate_path, &self.out_dir)?;
+        readme::copy_from_crate(&self.crate_data, &self.crate_path, &self.out_dir)?;
         info!("Copied readme from crate to {:#?}.", &self.out_dir);
         Ok(())
     }
--- a/src/manifest/mod.rs
+++ b/src/manifest/mod.rs
@@ -541,6 +541,14 @@ impl CrateData {
         }
     }
 
+    /// Gets the optional path to the readme, or None if disabled.
+    pub fn crate_readme(&self) -> Option<String> {
+        self.pkg()
+            .readme
+            .clone()
+            .map(|readme_file| readme_file.into_string())
+    }
+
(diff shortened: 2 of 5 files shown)


heaths:
This fixes some issues with a couple tests that were actually broken before the original PR this replaces. Looking at old runs, it seems that they were effectively broken when the change was made to stop deserializing the manifest in this crate directly and to use a tool instead. Maybe the breaks in the tests weren't fully exposed then - though there were subsequent build breaks exactly the same as after the original PR for this change - but I am happy to fix them because they certainly are related to what I was trying to achieve here: to ignore when a README was explicitly disabled (as I'd expect for a wasm).

drager (approved):
Thanks a lot!

PR #1321, opened 2023-08-10
Add option to skip optimization with wasm-opt

sisou:
I am currently running `wasm-pack` three times for our project. Once for the main-thread wasm file, once for the web-worker wasm file, and once to generate types that include both of those builds.

This `--no-opt` option would help me immensely by reducing our build time by ~30% by not having to wait for the `wasm-opt` step for the type generation. I cannot disable `wasm-opt` generally in `Cargo.toml`, because I still want it for the other two actual build steps.

An alternative to this PR would be a `--types-only` option, which _only_ outputs a .d.ts file.

### DIFF
--- a/src/command/build.rs
+++ b/src/command/build.rs
@@ -32,6 +32,7 @@ pub struct Build {
     pub reference_types: bool,
     pub target: Target,
     pub no_pack: bool,
+    pub no_opt: bool,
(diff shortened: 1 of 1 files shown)


drager (approved):
Thank you!

PR #1330, opened 2023-09-16
Fix binary tool version check

cyqsimon:
I was wondering why `wasm-pack` is insisting on pulling and building `cargo-generate` despite there already being one at `/usr/bin/cargo-generate`. I'm on EndeavourOS too so it is the newest version `0.18.4`.

Turns out, the version check logic no longer works for the newest `cargo-generate`. Right now `cargo-generate --version` prints `cargo generate 0.18.4`, which means the current logic is actually evaluating `"generate" == "0.18.4"`. The same problem also exists for `wasm-opt`. This patch fixes this, along with adding a few helpful debug prints.

You can test before and after using `RUST_LOG=debug cargo run -- new --mode no-install version-test`.

### DIFF
--- a/src/install/krate.rs
+++ b/src/install/krate.rs
@@ -1,5 +1,6 @@
 use crate::install::Tool;
 use anyhow::Result;
+use log::debug;
 use serde::Deserialize;
 const VERSION: Option<&str> = option_env!("CARGO_PKG_VERSION");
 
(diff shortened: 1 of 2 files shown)


cyqsimon:
Hmm. Seems like depending on the specific build used, `wasm-bindgen --version` may contain a commit hash (e.g. `wasm-bindgen 0.2.87 (f0a8ae3b9)`). Annoying.

I guess this is now a situation that calls for regex.

Issue #1252, opened 2023-03-22
--target-dir causes build errors when used in workspace


chanks:
## 🐛 Bug description
I have a workspace which contains both WASM and non-WASM binaries. I find that using the default target directory in the workspace root for both types of binary seems to thrash the target cache and cause unnecessary recompilations, so I'd like to compile the WASM binaries to a different directory. One WASM binary is in `{workspace_root}/crates/binary/user_interface_wasm`.

For reference, my default wasm-pack invocation is:
`wasm-pack build crates/binary/user_interface_wasm --dev -t web --no-typescript`
This works, but thrashes the cache:
```
[INFO]: License key is set in Cargo.toml but no LICENSE file(s) were found; Please add the LICENSE file(s) to your project directory
[INFO]: ⬇️  Installing wasm-bindgen...
[INFO]: ✨   Done in 1m 23s
[INFO]: 📦   Your wasm pkg is ready to publish at crates/binary/user_interface_wasm/pkg.
```

When I pass a --target-dir:
`wasm-pack build crates/binary/user_interface_wasm --dev -t web --no-typescript --target-dir target-web`
It builds the WASM app in `crates/binary/user_interface_wasm/target-web` rather than `{workspace_root}/target-web`. I guess that's fine, but it also errors, seemingly because it's looking for the built binary in the wrong place:
```
[INFO]: License key is set in Cargo.toml but no LICENSE file(s) were found; Please add the LICENSE file(s) to your project directory
[INFO]: ⬇️  Installing wasm-bindgen...
error: failed reading 'target-web/wasm32-unknown-unknown/debug/user_interface_wasm.wasm'

Caused by:
    No such file or directory (os error 2)
Error: Running the wasm-bindgen CLI
Caused by: Running the wasm-bindgen CLI
Caused by: failed to execute `wasm-bindgen`: exited with exit status: 1
  full command: "/home/chris/.cache/.wasm-pack/wasm-bindgen-5f26acfc988649a3/wasm-bindgen" "target-web/wasm32-unknown-unknown/debug/user_interface_wasm.wasm" "--out-dir" "crates/binary/user_interface_wasm/pkg" "--no-typescript" "--target" "web" "--debug"
```

Passing a target-dir that puts the build directory at the workspace root is a bit nicer place for it:
`wasm-pack build crates/binary/user_interface_wasm --dev -t web --no-typescript --target-dir ../../../target-web`
But also fails with a similar error:
```
[...]

Issue #1156, opened 2022-07-19
Specify cargo build target dir?


shrinktofit:
## 💡 Feature description

I want to specify all the intermediate compilation files or wasm output into specified dir. I used `wasm-pack build <input-dir> --out-dir <wasm-pack-output-dir> -- --target-dir <cargo-target-dir>`, but the following error was emitted:

```txt
[INFO]: Installing wasm-bindgen...
error: failed reading '<input-dir>\target\wasm32-unknown-unknown\release\<not-important>.wasm'

Caused by:
    系统找不到指定的路径。 (os error 3)
    ^^^^^^^^^^^^^^^
    (Which means "system can not find specified path")
```

Seems like wasm-pack always assumes the cargo target output is emitted to `<input-dir>/target`?

#### 💻 Basic example
Include a basic code example if possible. Omit this section if not applicable.

kaimast:
To me it looks like as wasm-pack changes directories incorrectly.

I specify `CARGO_TARGET_DIR=wasm-target` and it ends up compiling everything in the following folder: `<workspace_dir>/<crate_dir>/wasm-target` (which isn't the correct path).

But then wasm-pack looks for the binary in `<workspace_dir>/wasm-target` (which is the correct path) and fails.

drager:
Fixed in #1331.

PR #1331, opened 2023-09-22
Fix passing relative paths to cargo

dfaust:

```bash
```

When building a crate located in a sub-directory, relative paths, passed as extra options to cargo (like `--target-dir`), are now handled correctly.

This closes issues #704, #1156 and #1252.
And makes PR #712 obsolete.


drager (approved):
Thank you!

Issue #1338, opened 2023-10-11
Multiple dependencies have critical defects identified by cargo-audit


Teebor-Choka:
## 🐛 Bug description
Some of the dependencies used in wasm-pack should be updated due to critical advisories.

Running `cargo audit` in a project that uses wasm-pack results in the following advisories issued as errors:
```shell
Crate:     rustls-webpki
Version:   0.100.1
Title:     rustls-webpki: CPU denial of service in certificate path building
Date:      2023-08-22
ID:        RUSTSEC-2023-0053
URL:       https://rustsec.org/advisories/RUSTSEC-2023-0053
Severity:  7.5 (high)
Solution:  Upgrade to >=0.100.2, <0.101.0 OR >=0.101.4
Dependency tree:
rustls-webpki 0.100.1
├── webpki-roots 0.23.1
│   └── ureq 2.7.1
│       ├── wasm-pack 0.12.1
...
Crate:     time
Version:   0.1.45
Title:     Potential segfault in the time crate
Date:      2020-11-18
ID:        RUSTSEC-2020-0071
URL:       https://rustsec.org/advisories/RUSTSEC-2020-0071
Severity:  6.2 (medium)
Solution:  Upgrade to >=0.2.23
Dependency tree:
time 0.1.45
└── chrono 0.4.26
    ├── wasm-pack 0.12.1
...
Crate:     atty
Version:   0.2.14
Warning:   unsound
Title:     Potential unaligned read
Date:      2021-07-04
ID:        RUSTSEC-2021-0145
URL:       https://rustsec.org/advisories/RUSTSEC-2021-0145
Dependency tree:
atty 0.2.14
└── wasm-pack 0.12.1
```

#### 🤔 Expected Behavior
No advisories for dependencies.

#### 👟 Steps to reproduce
Install the `cargo audit` utility as `cargo install cargo-audit` and run the audit with `cargo audit`

#### 🌍 Your environment
Include the relevant details of your environment.
wasm-pack version: wasm-pack 0.12.1
rustc version: rustc 1.72.1 (d5c2e9c34 2023-09-13)

PR #1339, opened 2023-10-11
Update dependencies to fix active advisories

Teebor-Choka:
Remove active advisories by updating the dependencies to versions that fixed the issues causing the advisories.

```bash
```

Fixes parts of the #1338 


Teebor-Choka:
Lock file needs to be updated in the native development environment.

Issue #1342, opened 2023-10-27
OS System Rust package pecularities cause wasm-pack build not to work on Quickstart from wasm-book


daidoji:
## 🐛 Bug description
Summary: 
1. I attempted to complete the Quickstart "hello-wasm" from the wasm-book.  
2. Got error for no wasm32-unknown-unknown target.
3. Followed instructions for non-rustup install of that target with slight alterations for my particular OS below.
4. Still got a weird error
5. Followed up and explored a bit and point out where I think the issue is in wasm-build.

Got this error.
```
(keri) daidoji@darkstar:~/ssi/cesride/wasm$ wasm-pack build
[INFO]: 🎯  Checking for the Wasm target...
Error: wasm32-unknown-unknown target not found in sysroot: "/usr"

Used rustc from the following path: "/usr/bin/rustc"
It looks like Rustup is not being used. For non-Rustup setups, the wasm32-unknown-unknown target needs to be installed manually. See https://rustwasm.github.io/wasm-pack/book/prerequisites/non-rustup-setups.html on how to do this.

Caused by: wasm32-unknown-unknown target not found in sysroot: "/usr"

Used rustc from the following path: "/usr/bin/rustc"
It looks like Rustup is not being used. For non-Rustup setups, the wasm32-unknown-unknown target needs to be installed manually. See https://rustwasm.github.io/wasm-pack/book/prerequisites/non-rustup-setups.html on how to do this.
```

Attempted to follow the instructions in the error message (ie https://rustwasm.github.io/wasm-pack/book/prerequisites/non-rustup-setups.html).  However, I'm using Slackware Linux (-current) and the system rust package (not installed with rustup).  Slackware seems to install `rustlib` to `/usr/lib64/rustlib/`.  So I put the files there thinking it would be no big deal.

Got the same error.  ie) 
```
(keri) daidoji@darkstar:~/ssi/cesride/wasm$ rustc --print sysroot
/usr
(keri) daidoji@darkstar:~/ssi/cesride/wasm$ ls /usr/lib64/rustlib/
components   manifest-cargo.gz           manifest-rust-src.gz                           manifest-rustfmt-preview.gz  uninstall.sh*
etc/         manifest-clippy-preview.gz  manifest-rust-std-x86_64-unknown-linux-gnu.gz  rust-installer-version       wasm32-unknown-unknown/
install.log  manifest-rust-docs.gz       manifest-rustc.gz                              src/                         x86_64-unknown-linux-gnu/
[...]

PR #1343, opened 2023-10-30
Rewrite wasm_target to use target-libdir

daidoji:
Rewritten wasm_target to use target libdir from the rustc tool rather than looking through sysroot.  This is to accomodate non-rustup installations.

Fixes [bug-1342](https://github.com/rustwasm/wasm-pack/issues/1342) 

Can clean up unused sysroot stuff too in this module if that's appropriate.  Just wanted someone to agree with this before I put anymore work into it.

```bash
```


drager (approved):
Thank you!

PR #1344, opened 2023-10-31
Docs: Add a description of `build --target deno` 

puxiao:
emm...

> Outputs JS that can be natively imported as an ES module in deno.

Is this description correct?


drager (approved):
Thanks!

PR #1348, opened 2023-11-24
chore: document deno in build target

sigmaSd:


sigmaSd:
There is a similar thing in publish https://github.com/rustwasm/wasm-pack/blob/bdc4884fc5dcb50d78e521acfb2f613fb6588b48/src/command/build.rs#L141 but I don't know if that one works

drager (approved):
Thanks!

PR #1391, opened 2024-05-14
fix: update chromedriver download URL, resolve test issues

### LINKED ISSUE #1390: CI is not working
## 🐛 Bug description

CI for this repo doesn't appear to be working.

#### 🤔 Expected Behavior

CI jobs should be passing.

#### 👟 Steps to reproduce

1. Look at the Github Actions results on `master`.

#### 🌍 Your environment

n/a

kade-robertson:
This fixes a few things to get CI working for the repo again:
- The chromedriver releases were previously 404ing, causing these tests to fail. The URL for where these are hosted was changed, so it's just a simple update to get this working again.
- In later Cargo releases, the output format for the "Finished..." log line changed. I've fixed this by making `matches_cargo` accept both the old and new formats.
- With MacOS 14 (and by extension, M1) runners becoming the `macos-latest` default, the current geckodriver setup fails for that platform. ~It's not entirely obvious to me why at this point. Downgrading to the `macos-13` runner has these tests working again. This is probably the most.. questionable change as it'd be better to track down why the latest runners don't work.~ Edit: It seems like the only thing preventing macOS from working was that I needed to manually install Firefox. I'm guessing this was available in the previous macOS runners by default.

---

```bash
```

Closes #1390

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -21,7 +21,7 @@ jobs:
             os: ubuntu-latest
             rust: stable
           - build: macos-stable
-            os: macos-latest
+            os: macos-13
             rust: stable
           - build: windows-stable
             os: windows-latest
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -21,7 +21,7 @@ jobs:
             os: ubuntu-latest
             rust: stable
           - build: macos-stable
-            os: macos-13
+            os: macos-latest
             rust: stable
           - build: windows-stable
             os: windows-latest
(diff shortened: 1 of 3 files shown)


kade-robertson:
Seems like there's some unfortunate caching issues I'll have to dig into.

drager:
> Seems like there's some unfortunate caching issues I'll have to dig into.

Thank you for looking into this! Appreciated!

kade-robertson:
> Thank you for looking into this! Appreciated!

This issue _seems_ to be resolved now for the sake of this PR. However, in general I've found some consistency issues with trying to determine what paths to cache when manually in many of my other projects. Are you open to switching to something like [Swatinem/rust-cache](https://github.com/Swatinem/rust-cache) or [mozilla-actions/sccache-action](https://github.com/Mozilla-Actions/sccache-action)? I've used both successfully quite a bit, though recently all of my projects have moved to the latter. If so I can open up another PR after this is cleared to merge to switch out the caching.

drager (approved):
Thanks a lot!

PR #1131, opened 2022-05-01
Add skip-gitignore cli arg

daggy1234:
Make sure these boxes are checked! 📦✅

- [x] You have the latest version of `rustfmt` installed
```bash
$ rustup component add rustfmt
```
- [x] You ran `cargo fmt` on the code base before submitting
- [x] You reference which issue is being closed in the PR text

✨✨ 😄 Thanks so much for contributing to wasm-pack! 😄 ✨✨

This fixes #728 with a simple cli flag `--skip-gitignore`

PR #1408, opened 2024-07-06
Don't create .gitignore in pkg directory

### LINKED ISSUE #728: Allow skipping `.gitignore` generation
## 💡 Feature description

Since implementing self-`.gitignore` generation in https://github.com/rustwasm/wasm-pack/issues/473, I've come across few codebases that prefer not to publish npm artifacts elsewhere nor integrate Rust build into their build process, but keep any artifacts in the Git repo.

I suggest adding `--no-gitignore` option for such usecases.

## Alternative

Since in such cases module is not intended to be published anywhere, another alternative could be to implement this (omit `.gitignore`) as part of `--no-pack` implementation proposed here https://github.com/rustwasm/wasm-pack/issues/691.

cc @ashleygwilliams

### LINKED ISSUE #909: 'pkg/.gitignore' in deno target
In Deno, packages are usually hosted on github or other git hosting platforms.
But now the content of generated pkg/.gitignore is `*`, which will prevent pkg from being submitted to git.
Should the `*` in the .gitignore of the Deno target be deleted?

### LINKED ISSUE #984: Don't overwrite an existing .gitignore file

### LINKED ISSUE #999: `build --target web`: have an option not to add .gitignore to pkg output folder
## 💡 Feature description
I recently found out that it's possible to have wasm-powered static websites on Github pages straight from master branch, only problem is that right now I have to keep remembering to delete .gitignore from pkg output folder before commit.

I know that it is not intended use, but it makes working with Pages a bit easier, and by that I guess that others may follow. I'm pretty sure adding that flag is near-trivial, so why not just have it as an option?

#### 💻 Basic example
`wasm-pack build --target web --no-gitignore` builds just like `wasm-pack build --target web`, but ommits creation of additional .gitignore inside output folder.

mrothNET:
Closes #728 
Closes #984,
Closes #909
Closes #999.
Supersedes #1131, #985 

### Problematic Behaviour

Currently, `wasm-pack` generates a `.gitignore` file in the `pkg` directory during each build.

### Suggested Solution

This PR proposes removing the automatic creation of the `.gitignore` file without introducing a replacement.

### Justification

While many tools generate an appropriate `.gitignore` during project *initialization*, it’s relatively uncommon (in my experience) for build tools, like compilers, to create `.gitignore` files during a *build*.

Moreover, this behavior has led to several issues, as outlined in the following requests:

- #728
- #984
- #909
- #999

This PR aims to remove the generation of the problematic `.gitignore` file, addressing and potentially superseding these other PRs:

- #1131
- #985

### Changes made

- Remove one line from [src/command/utils.rs](https://github.com/rustwasm/wasm-pack/pull/1408/files#diff-36d20b5b00d9e4e25c74b1cebb9dd935d248c1a9231bf0ebda7d54c6e8cc9119)
- Update documentation in [docs/src/commands/build.md](https://github.com/rustwasm/wasm-pack/pull/1408/files#diff-b2a4b3d52fd1f77d27b83df800176970fd451f1e37aa42266c28c250f402c036)

### DIFF
--- a/docs/src/commands/build.md
+++ b/docs/src/commands/build.md
@@ -3,11 +3,7 @@
 The `wasm-pack build` command creates the files necessary for JavaScript
 interoperability and for publishing a package to npm. This involves compiling
 your code to wasm and generating a pkg folder. This pkg folder will contain the
-wasm binary, a JS wrapper file, your `README`, and a `package.json` file.
-
-The `pkg` directory is automatically `.gitignore`d by default, since it contains
-build artifacts which are not intended to be checked into version
-control.<sup>[0](#footnote-0)</sup>
+wasm binary, a JS wrapper file, your `README`, and a `package.json` file.<sup>[0](#footnote-0)</sup>
 
 ## Path
 
@@ -140,6 +136,5 @@ wasm-pack build examples/js-hello-world --mode no-install -- --offline
 
 <hr style="font-size: 1.5em; margin-top: 2.5em"/>
 
-<sup id="footnote-0">0</sup> If you need to include additional assets in the pkg
-directory and your NPM package, we intend to have a solution for your use case
-soon. [↩](#wasm-pack-build)
+<sup id="footnote-0">0</sup> In contrast to previous releases, a `.gitignore`
+file is no longer created in the `pkg` directory. [↩](#wasm-pack-build)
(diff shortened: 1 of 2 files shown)


rimutaka (approved):
LGTM. I wonder why it hasn't been merged.

satindra-r:
was looking for this, hope it gets merged soon

BadMachine:
any progress?

jefshe:
bumping! it seems very strange that this is on by default as it's very easy for a dev to add it themselves and causes issues when we want to commit type files

Enduriel:
Would be really nice to see get added, with `--no-pack` being a thing it's rather annoying to have a custom package.json to then have it get gitignored because of running a wasm-pack build.

PR #1423, opened 2024-08-24
Be more explicit with available wasm-opt flags in documentation

### LINKED ISSUE #1422: Be more explicit with available wasm-opt flags in documentation
The existing documentation about wasm-opt optimization flags are easy to glance over, and the available wasm-opt flags from binaryen are already obscure to find. Outside of `-O4`, I have personally found these flags to be useful, so I think a documentation note specifying the available wasm-opt flags for extreme optimization cases is worth adding.

ArhanChaudhary:
Closes #1422

This PR explicitly specifies the available wasm-opt optimization flags for extreme optimization cases.

- [X] You have the latest version of `rustfmt` installed
```bash
```
- [X] You ran `cargo fmt` on the code base before submitting
- [X] You reference which issue is being closed in the PR text


drager (approved):
Thank you!

PR #1436, opened 2024-09-30
chore: Remove unmaintained dependency atty in favor of stdlib

mariusvniekerk:
As per https://github.com/softprops/atty/issues/57#issuecomment-2367878741 we can just use the stdlib functionality that exists in Rust 1.70


drager (approved):
Thanks!

PR #1437, opened 2024-09-30
Replace `atty` with `std::io::IsTerminal`

mbrobbel:
`atty` is unmaintained: https://rustsec.org/advisories/RUSTSEC-2024-0375.html.

This PR removes it from the dependencies and instead uses https://doc.rust-lang.org/stable/std/io/trait.IsTerminal.html. This means the `msrv` is bumped to `1.70.0`.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -132,17 +132,6 @@ dependencies = [
 ]
 
 [[package]]
-name = "atty"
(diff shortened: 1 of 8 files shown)


mbrobbel:
Closing in favor of #1436.

PR #1438, opened 2024-10-11
Requests using proxy settings from ENV

jjyr:
This PR enables ureq to use proxy settings from ENV, it solves lots of pain in network restricted environments.

I also send a PR to binary-install https://github.com/rustwasm/binary-install/pull/34


drager (approved):
Thank!

PR #1450, opened 2024-11-24
fix: Fix the parse error when package.json is generated by wasm-bindgen

mohebifar:
When building one of my projects, I kept running into this issue:

```bash
wasm-pack build --target web
```

**Error:**

```txt
Finished `release` profile [optimized] target(s) in 0.05s
Error: invalid type: map, expected a string at line 2 column 18
Caused by: invalid type: map, expected a string at line 2 column 18
```

### Root Cause

The issue originates in the `step_create_json` and the `fn write_package_json`. During the build process, the following `package.json` was generated by `wasm-bindgen`:

```json
{
  "dependencies": {
    "web-tree-sitter-sg": "0.22.2"
  }
}
```

While this `package.json` is valid, the code incorrectly attempted to parse the entire file as a `HashMap<String, String>`. This led to a failure when merging dependencies because the `dependencies` field itself is a map, not a string.

### Fix

This PR updates the parsing logic to correctly handle the `package.json` structure.

---

```bash
```


mohebifar:
Bumping this PR. Any blockers for merging this?

Issue #1206, opened 2023-01-09
`snippets` directory is not included in `files` of `package.json`


Gu-Miao:
`/src/lib.rs`:

```rust
#[wasm_bindgen(module = "/www/utils.js")]
extern "C" {
  fn random(max: usize) -> usize;
}
```

`/www/utils.js`:

```js
export function random(max) {
  return Math.round(Math.random() * max)
}
```

`wasm-pack` builds correctly:

![image](https://user-images.githubusercontent.com/28031505/211259147-afb00367-03b5-4ebc-ada1-eee45bfbfa34.png)

But in `package.json`, it is not inclueded:

![image](https://user-images.githubusercontent.com/28031505/211259217-c46457a1-7935-4406-91e8-df59465efb5f.png)

I found similar issues: #663, #557, #881. But all of them were not resloved. So I create this.

PR #1479, opened 2025-02-27
fix: add the snippets directory by default

jdrouet:
This should fix https://github.com/rustwasm/wasm-pack/issues/1206

I don't know what is _the way_ to detect if the `snippets` directory should be included so, by default, I include it considering it's already in the `sideEffect` section.
Another possibility would be to add to the `Cargo.toml` metadata a `package-json` section that allows to add some files and enforce it there.

Tell me if it's ok with you or if you need some updates 😉 

```bash
```

### DIFF
--- a/src/manifest/mod.rs
+++ b/src/manifest/mod.rs
@@ -668,6 +668,7 @@ impl CrateData {
             let js_bg_file = format!("{}_bg.js", name_prefix);
             files.push(js_bg_file);
         }
+        files.push("snippets".into());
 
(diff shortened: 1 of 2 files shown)


LuisOsta:
@jdrouet When do you think this PR would be merged? Do you have any idea of who we should reach out to about this?

jdrouet:
Honestly, an approval from the maintainer 😁

LuisOsta:
I tested this on my repo and verified that it worked for us FWIW

Issue #1508, opened 2025-07-21
Many wasm-pack tests fail on rustc 1.88


mshroyer:
## 🐛 Bug description
When running `cargo test` on wasm-pack itself, many tests will fail

#### 🤔 Expected Behavior
`cargo test` should succeed

#### 👟 Steps to reproduce
```
rustup toolchain install --profile minimal 1.88
rustup override set 1.88
cargo test
```

#### 🌍 Your environment
Include the relevant details of your environment.
wasm-pack version: master (commit id 32e52ca893e26c97bbf478c57ae1e9abed23c449)
rustc version: 1.88.0 (6b00bc388 2025-06-23)
macOS 15.5

PR #1509, opened 2025-07-21
Fix many broken cargo tests

mshroyer:
This fixes most, but not all, test failures mentioned in #1508 . There are still one or two different failures on macOS 15.5 depending on the rust version used.

## Template

```bash
```


drager (approved):
Thank you!

PR #1513, opened 2025-08-02
Update the document URL and the repsitory URL to `drager/wasm-pack`

yutannihilation:
Thanks for maintaining wasm-pack!

I think one of the confusion like https://github.com/drager/wasm-pack/issues/1511 is that most URL still refers to the previous repository. For example, <https://rustwasm.github.io/wasm-pack/installer/> doesn't exist, so it should be pointed to <https://drager.github.io/wasm-pack/installer/>. Also, the repository URL should also be `drager/wasm-pack` instead of `rustwasm/wasm-pack`.

---

```bash
```


drager (approved):
Thank you!

viliml:
https://github.com/drager/wasm-pack-template doesn't exist.

PR #1534, opened 2025-11-12
Pass arguments to wasm-bindgen

### LINKED ISSUE #1092: Pass arguments to wasm-bindgen
## 💡 Feature description
Right now wasm-pack allows passing arguments to cargo build, but doesn't do the same for wasm-bindgen. I feel like this should be a pretty important feature to have. In my case, I want to use `--weak-refs` flag with wasm-bindgen, and at the moment I see no way to use this feature while using wasm-pack.

For the syntax, we could add a flag:
```console
# Like this:
wasm-pack --wbg-arg --weak-refs --wbg-arg --reference-types
# Or this:
wasm-pack --wbg-args '--weak-refs --reference-types'
```
The only problem I see with this syntax is it's different from the syntax for passing arguments to cargo, which is done with the `--` flag. That's a bit unfortunate, but ¯\\\_(ツ)_/¯.

So, what do people think?

rafaelbeckel:
This PR implements support for passing arbitrary flags to wasm-bindgen via wasm-pack build.

Addresses https://github.com/drager/wasm-pack/issues/1092

## CLI changes
- New repeatable flag: --wbg-arg <ARG> (alias: --wbg)
  - Example usage:
    - wasm-pack build --wbg-arg --experimental-reset-state-function
    - wasm-pack build --wbg-arg --weak-refs --wbg-arg --reference-types

## Implementation
- BuildOptions: add wbg_arg: Vec<String> (Clap v4, allow_hyphen_values, visible_alias = "wbg").
- Build: store bindgen_args and plumb through the build pipeline.
- bindgen::wasm_bindgen_build: accept &[String] of args and append them to the wasm-bindgen Command (after defaults, so users can override if desired).
- docs: Document the new flag with examples.
- tests: Add an integration test asserting that a forwarded unknown flag is visible in the wasm-bindgen invocation (proves passthrough).

Closes #1092

### DIFF
--- a/src/manifest/mod.rs
+++ b/src/manifest/mod.rs
@@ -683,7 +683,7 @@ impl CrateData {
             None
         };
 
-        let keywords = if !pkg.keywords.is_empty() {
+        let keywords = if pkg.keywords.len() > 0 {
             Some(pkg.keywords.clone())
         } else {
             None


src/command/build.rs
```diff
@@ -40,15 +40,18 @@ pub struct Build {
...
     pub bindgen: Option<install::Status>,
     pub cache: Cache,
+    pub bindgen_args: Vec<String>,
     pub extra_options: Vec<String>,
 }
 
 /// What sort of output we're going to be generating and flags we're invoking
 /// `wasm-bindgen` with.
 #[derive(Clone, Copy, Debug)]
+#[derive(Default)]
```
rafaelbeckel:
```suggestion
#[derive(Clone, Copy, Debug, Default)]
```

src/command/build.rs
```diff
@@ -65,11 +67,6 @@ pub enum Target {
     Deno,
 }
```
rafaelbeckel:
```suggestion
```

src/command/build.rs
```diff
@@ -182,33 +178,14 @@ pub struct BuildOptions {
     /// Option to skip optimization with wasm-opt
     pub no_opt: bool,
 
+    #[clap(long = "wbg-arg", visible_alias = "wbg", allow_hyphen_values = true)]
+    /// Pass an argument to wasm-bindgen. May be used multiple times.
+    pub wbg_arg: Vec<String>,
+
     /// List of extra options to pass to `cargo build`
     pub extra_options: Vec<String>,
 }
```
rafaelbeckel:
```suggestion
```

src/command/login.rs
```diff
@@ -16,9 +16,9 @@ pub fn login(
         &scope, &registry, &auth_type
     );
     info!("npm info located in the npm debug log");
-    npm::npm_login(&registry, &scope, &auth_type)?;
+    npm::npm_login(&registry, scope, auth_type)?;
```
rafaelbeckel:
```suggestion
    npm::npm_login(&registry, &scope, &auth_type)?;
```

src/command/login.rs
```diff
@@ -19,6 +19,6 @@ pub fn login(
     npm::npm_login(&registry, &scope, &auth_type)?;
     info!("Logged you in!");
 
-    PBAR.info(&"👋  logged you in!".to_string());
+    PBAR.info("👋  logged you in!");
```
rafaelbeckel:
```suggestion
    PBAR.info(&"👋  logged you in!".to_string());
```

src/manifest/mod.rs
```diff
@@ -473,8 +473,8 @@ impl CrateData {
     }
 
     fn is_same_path(path1: &Path, path2: &Path) -> bool {
-        if let Ok(path1) = fs::canonicalize(&path1) {
-            if let Ok(path2) = fs::canonicalize(&path2) {
+        if let Ok(path1) = fs::canonicalize(path1) {
+            if let Ok(path2) = fs::canonicalize(path2) {
```
rafaelbeckel:
```suggestion
        if let Ok(path1) = fs::canonicalize(&path1) {
            if let Ok(path2) = fs::canonicalize(&path2) {
```

src/manifest/mod.rs
```diff
@@ -683,7 +683,7 @@ impl CrateData {
             None
         };
 
-        let keywords = if pkg.keywords.len() > 0 {
+        let keywords = if !pkg.keywords.is_empty() {
```
rafaelbeckel:
```suggestion
        let keywords = if pkg.keywords.len() > 0 {
```

PR #1543, opened 2025-12-15
fix(npm): update npm package download url

qinyuhang:
The npm binary.js install still referred to old wasm repo. So I made the change to point to new url.

```bash
```


drager (approved):
Thanks!

PR #1583, opened 2026-05-16
feat: support wasm32-unknown-emscripten target

guybedford:
This adds support for the `wasm32-unknown-emscripten` target alongside the existing `wasm32-unknown-unknown` flow. The emscripten target makes `std`-library APIs that aren't available under `wasm32-unknown-unknown` usable from wasm-pack-built crates: `SystemTime`, `std::env`, `std::fs` (via MEMFS), `HashMap` default random state, `getentropy`-backed randomness, and POSIX-style I/O.

**The whole build is a single cargo invocation.** rustc drives `emcc` as the linker for the emscripten target, and emcc's `-sWASM_BINDGEN` runs `wasm-bindgen` itself as a post-link step, detected via the marker section wasm-bindgen embeds for emscripten builds (shipped in wasm-bindgen 0.2.122+; emcc integration upstream in emscripten-core/emscripten#23493 and emscripten-core/emscripten#27208). There is no wasm-pack-orchestrated link/bindgen/post-link pipeline and no wasm-opt step — emcc owns its own optimization pipeline, driven by the rustc opt-level.

wasm-pack's remaining role:

* installs the `wasm-bindgen` CLI matching `Cargo.lock` and prepends it to `PATH` (emcc resolves `wasm-bindgen` via PATH); requires >= 0.2.122. When `wasm-bindgen` is a git dependency the CLI is `cargo install`ed from the same revision (cached per rev) instead of a registry release, since a release would neither match the descriptor format nor carry the same fixes
* runs `cargo rustc` with the emcc output-shape settings injected as final-crate link args: `-sWASM_BINDGEN -sMODULARIZE=instance -sEXPORT_ES6 -sAUTO_INIT`, plus `-sENVIRONMENT=node` for `--target nodejs` and `-sSOURCE_PHASE_IMPORTS` for the new `--target module`; `--target no-modules` is rejected (emcc emits ES modules). Output uses standard (exnref) exception handling only: `-sWASM_LEGACY_EXCEPTIONS=0` plus `-sBINARYEN_EXTRA_PASSES=--translate-to-exnref` to translate the legacy instructions rustc's LLVM still emits
* copies the emitted self-initializing ES module + `.wasm` into `pkg/` and writes a tailored `package.json` (`type: "module"`, no `_bg.*` artifacts)

`cargo rustc` trailing args are used deliberately instead of `RUSTFLAGS`: they apply to the final crate only and *compose* with user rustflags from `.cargo/config.toml` (a `RUSTFLAGS` env var would replace them), so users keep their own emcc settings (`-sSTACK_SIZE`, etc).

Crate shape: emscripten packages are **bin crates** (an empty `fn main() {}` runs on module init, the API is the `#[wasm_bindgen]` exports). rustc links emscripten `bin` targets as main modules; `cdylib` would produce an emscripten *side module* (`-sSIDE_MODULE=2`), which is a different artifact entirely. Crate-config checks emit clear diagnostics for both directions.

Consuming the output is direct ESM — the module self-initializes on import:

```js
import { greet } from "./pkg/my_pkg.js";
greet("world");
```

Also included:

* `wasm-pack new --emscripten` template (bin crate + minimal `.cargo/config.toml`)
* `--target module` for the regular (non-emscripten) pipeline, mapping to wasm-bindgen's `module` output mode
* `wasm-opt` (non-emscripten flow) now passes the full Wasm 3.0 feature flag set so any standard-conforming module validates (supersedes the targeted `--enable-exception-handling`/`--enable-memory64`/`--enable-reference-types` pushes; will reconcile with #1600 on rebase)
* docs at `docs/src/emscripten-target.md` describing the flow and toolchain status
* `wasm-pack test` is rejected for emscripten with a clear message (`wasm-bindgen-test-runner` can't drive emcc artifacts yet)
* **Emscripten toolchain auto-install**: if no `emcc` is found, wasm-pack offers to install the toolchain into its cache (one-time ~1.3 GB; confirm-prompted on a tty, automatic otherwise) and applies it process-scoped to the `cargo` child — no shell activation needed. Resolution order: `PATH` emcc → `$EMSDK` → `~/emsdk` → wasm-pack cache → install. The managed install is a pinned emsdk tip-of-tree build (prebuilt LLVM/binaryen/emscripten from one emscripten-releases revision; becomes a plain `6.0.10` version bump once tagged). The resolved emcc entry point is passed to rustc explicitly via `-Clinker`, since rustc's default `emcc.bat` name never matches the pylauncher `emcc.exe` prebuilt SDKs ship on Windows.

Testing: 8 integration tests build the fixture end-to-end for each `--target` and drive the full `#[wasm_bindgen]` surface from Node in one invocation (strings, closures, typed arrays, `Result`→exception, classes/getters, JS imports, `module = "node:os"`/`node:path` ESM imports, `js_namespace` including nested class namespaces), plus rejection cases. The fixture pins `wasm-bindgen`/`js-sys` to wasm-bindgen `main` (carrying walrus 0.27.2 via wasm-bindgen#5328), exercising the git-sourced CLI path. The tests are opted into via `WASM_PACK_TEST_EMSCRIPTEN=1`; CI runs them on runners with no preinstalled emcc on both Linux and Windows, exercising the auto-install path itself.

Toolchain status (the PR tracks the final story; upstream items remain):

* **emscripten post-link `-sWASM_BINDGEN`** — landed on `main` in emscripten-core/emscripten#27208 for 6.0.10, not yet in a tagged release. The managed toolchain pins a tip-of-tree emsdk build until 6.0.10 ships.
* **DWARF through emcc's post-link pipeline** — wasm-bindgen's `--keep-debug` output made binaryen's `wasm-opt -g` fail: `.debug_loc` tombstones read as base-address-selection entries (wasm-bindgen/walrus#322) and `DW_AT_high_pc` re-encoded as variable-width `udata` (wasm-bindgen/walrus#323). Both are fixed in walrus 0.27.2, on wasm-bindgen `main` via wasm-bindgen#5328, which the fixture pins; debug builds keep DWARF through the exnref translation. The fixture pin moves back to a release once one includes #5328.
* **Optimized builds** — fixed by [wasm-bindgen#5270](https://github.com/wasm-bindgen/wasm-bindgen/pull/5270) (glue now references emcc's top-level export bindings so metadce/minification track them); the integration tests build `--dev` until a wasm-bindgen release containing it ships (TODO'd in the tests).
* No `.d.ts` for emscripten packages yet — pending emcc `--emit-tsd` support for wasm-bindgen's multi-value-return exports.
* `panic=unwind` on emscripten is tracked in wasm-bindgen#5165; the template ships `panic=abort`.


docs/src/emscripten-target.md
```diff
@@ -0,0 +1,178 @@
...
+
+Two things are notable compared to the default template:
+
+1. **`crate-type = ["staticlib"]`** — emcc consumes a static library and
+   links it into the final wasm itself. `cdylib` would skip emcc entirely.
+
+2. **`.cargo/config.toml`** — selects the emscripten target and sets the
+   rustflags emcc expects:
+   - `-Cpanic=abort` — `panic=unwind` isn't supported across the wasm-bindgen
+     boundary on emscripten yet.
```
walkingeyerobot: should we have a tracking issue for this and link it here? would be nice to eventually resolve this.

docs/src/emscripten-target.md
```diff
@@ -0,0 +1,178 @@
...
+  typed; if you need them you can extend the `.d.ts` in `pkg/` after the
+  build.
+- **Optimization is capped at `-O2`.** emcc's `-O3` enables wasm-opt's
+  `--minify-imports-and-exports` pass, which renames wasm exports to
+  single letters — but wasm-bindgen's generated JS glue references the
+  original names. `-O2` produces near-equivalent output without the
+  renaming.
+- **`--target module` builds run unoptimized today.** emcc's bundled JS
+  optimizer (`acorn-optimizer`) can't parse `import source` syntax. Pass
+  `--no-opt` for `--target module` builds, or wait for the upstream emcc
```
walkingeyerobot, line 152: a linked issue / PR here would be helpful. then once the upstream fix is in, readers could see the PR was merged and would know this is no longer the case.

walkingeyerobot (approved):
This looks awesome, thanks!

I think this is awesome overall, but I did want to ask: do we need to have emcc invoke wasm-ld? What are the pros and cons vs just invoking wasm-ld directly here? I think this is a pretty minor point and you could reasonably ignore this question.

docs/src/emscripten-target.md
guybedford: The upstream fix has now landed for this one, I will merge it in here and reenable after the next emscripten release.

docs/src/emscripten-target.md
guybedford: Added a tracking issue.

renovatus:
Looking forward to this, thank you @guybedford!

PR #1606, opened 2026-09-23
ci(release): add Windows ARM binaries

6x68:

```bash
```

Don't know why this wasn't already added, but it just builds wasm-pack for Windows ARM, so...


6x68:
realized the issue I linked initially isn't actually the issue this fixes.

